PR #3, opened 2023-09-04
Add rangeLookup implementation for Data.Map

phadej:
To be used in model implementation

### DIFF
--- a/lsm-tree.cabal
+++ b/lsm-tree.cabal
@@ -29,6 +29,7 @@ common warnings
 library
   import:           warnings
   exposed-modules:
+    Data.Map.Range
     Database.LSMTree
     Database.LSMTree.Common
     Database.LSMTree.Monoidal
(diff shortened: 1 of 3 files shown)


phadej:
I think build (9.4.6, 3.10.1.0, ...) jobs are not yet configured but already required, so this PR is in the weird state.

jorisdral:
After #2 is merged, the workflow file should be updated to include the `ghc-9.4.6`/`cabal-3.10.1.0` job

dcoutts (approved):
This looks fine, and the tests look reliable.

I do wonder however if the range lookup could be done more elegantly using existing functions from the public API, rather than going directly to the internals. There's `spanAntitone` and `drop/takeWhileAntitone` which look like they'd do what we want.

phadej:
*more elegantly* for sure, but it will do more work constructing intermediate maps.

PR #17, opened 2023-10-11
Run Windows and MacOS jobs only on select GHC versions

jorisdral:
I'll remove required GHA jobs that are blocking this PR once I get an approval


dcoutts (approved):
LGTM

PR #25, opened 2023-10-24
Remove session argument from `snapshot`

phadej:
The table handle should already know its session,
and not having to duplicate the information will only help the code to be correct, as you wouldn't be able to try to snapshot table handle in the wrong session.


dcoutts (approved):
I agree.

jorisdral (approved):

PR #26, opened 2023-10-25
Add blob-specific model tests

phadej:


test/Test/Database/LSMTree/ModelIO/Normal.hs
```diff
@@ -201,7 +203,50 @@ prop_insertLookupRange h ups k v r = ioProperty $ do
...
+     IsTableHandle h
+  => Proxy h -> [(Key, Update Value Blob)]
+  -> Key  -> Value -> Blob -> Property
+prop_lookupInsertBlob h ups k v blob = ioProperty $ do
+    -- create session, table handle, and populate it with some data.
+    (_, hdl) <- makeNewTable h ups
+
+    -- the main dish
+    inserts hdl [(k, v, Just blob)]
+    res <- lookupsWithBlobs hdl [k]
```
jorisdral, line 217: If we run this prop for the real implementation, it would only test blob retrieval for blobs that are in the write buffer. Maybe we want a pattern that also puts some data in the table after `inserts`

Note: technically, if the insert triggers a flush of the write buffer, then we are testing blob retrieval for a run in the first level, but we are still not testing for later levels

Also, idem for other properties in this module. I think we should interleave "operations that are interesting" with "manipulating the table" (for example, by filling the table with some insert/delete data)

test/Test/Database/LSMTree/ModelIO/Normal.hs
```diff
@@ -201,7 +203,50 @@ prop_insertLookupRange h ups k v r = ioProperty $ do
 -- implement classic QC tests for split-value BLOB retrieval
```
jorisdral, line 203: How about adding props that are analogous to `prop_lookupDelete` and co?

phadej: `prop_lookupDelete` is already blob-aware.

```haskell
-- | You cannot lookup what you have just deleted
prop_lookupDelete ::
     IsTableHandle h
  => Proxy h -> [(Key, Update Value Blob)]
  -> Key -> Property
prop_lookupDelete h ups k = ioProperty $ do
    (_, hdl) <- makeNewTable h ups
    deletes hdl [k] -- this key may have a blob attached, may not. Depends on the table generated above.
    res <- lookupsWithBlobs hdl [k]
    return $ res === [NotFound k]
```

test/Test/Database/LSMTree/ModelIO/Normal.hs
phadej: Sure, but this is not relevant to this PR either.

I think what you described is better (and naturally) done by lockstep tests, where these properties are making sure simple stuff works.

Or to put differently, if we implement interleaved "operations that are interesting" with "manipulating the table" properties, than we (almost? completely?) make lockstep tests unnecessary.

jorisdral: I'm not saying we should try to test as many interleavings as possible here, and I agree about the lockstep tests, but I also am not against adapting these properties a little bit such that we'd be testing (the "real" implementation) a little more thoroughly.

jorisdral (approved):

dcoutts (approved):

PR #35, opened 2023-11-13
Add bloom filters with false-negative tests and FPR measurements

jorisdral:


src/Database/LSMTree/Internal/Run/BloomFilter.hs
```diff
@@ -22,4 +24,66 @@
...
+    monkeyFPR
+  , monkeyBits
+  , monkeyHashFuncs
+    -- * Bloom
+  , Bloom
+  , Bloom.elem
+    -- ** Hashes
+  , Hash
+  , Hash.Hashable (..)
+  , Hash.cheapHashes
```
phadej, line 38: I don't think we can use these. As `bloomfilter` package *does not* provide any serialization guarantees, i.e. is a memory only package, it for sure doesn't provide strong guarantees that the `cheapHashes` will have the same implementation indefinitely.

jorisdral: Alright, maybe we can "fork" the implementation of `cheapHashes` so we don't have to come up with our own hashing functions. How does that sound?

jorisdral: Or, we can add a number of golden tests so we can see when the hash function changes

phadej: The file format will still depend on an implementation detail of a package, i.e. cannot be stated independently in a specification (of said file format).

If that is not an issue, it's ok.

I wouldn't use a persistence layer which has such 3rd party implementation dependency myself.

dcoutts (approved):
@phadej makes a good point about file format stability. We will need to think about that carefully and agree an adequate solution.

PR #39, opened 2023-11-22
Add simple k-merge implementations, tests and benchmarks

phadej:
The binary-tree recursive two-way merge is quite fast (but less so for 5 inputs vs. 8 inputs):

    eight
      sortConcat:    OK (0.25s)
        63.2 μs ± 3.8 μs
      twoWayMerge:   OK (0.41s)
        48.7 μs ± 1.6 μs
      twoWayMerge2:  OK (0.15s)
        40.0 μs ± 3.1 μs
      heapKWayMerge: OK (0.28s)
        133  μs ±  11 μs
    five
      sortConcat:    OK (0.28s)
        33.7 μs ± 1.4 μs
      twoWayMerge:   OK (0.16s)
        20.7 μs ± 1.7 μs
      twoWayMerge2:  OK (0.15s)
        20.1 μs ± 1.9 μs
      heapKWayMerge: OK (0.13s)
        65.9 μs ± 5.3 μs

That will be time to beat.

### DIFF
--- a/lsm-tree.cabal
+++ b/lsm-tree.cabal
@@ -121,6 +121,24 @@ benchmark lsm-tree-macro-bench
   main-is:          Main.hs
   build-depends:    base
 
+test-suite kmerge-test
+  import:           warnings
+  default-language: Haskell2010
+  type:             exitcode-stdio-1.0
+  hs-source-dirs:   test
+  main-is:          kmerge-test.hs
+  build-depends:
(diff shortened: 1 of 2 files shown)


dcoutts (approved):
Nice.

So then I'm guessing the next step is something like
```
heapKWayMergeST :: forall a. Ord a => [[a]] -> [a]
```
that internally uses an in-place N-size heap for the N input lists.

PR #40, opened 2023-11-24
Separate module for QuickCheck generators

jorisdral:
Some of the code for generators/shrinkers can be shared with the micro-benchmark executable, so I moved it to the utils library


jorisdral:
Superseded by #38

PR #47, opened 2023-12-06
Vendor bloomfilter

phadej:


jorisdral (approved):
Let's also include the `bloomfilter` tests in the cabal file, but that can also happen in a separate PR

dcoutts (approved):
Ok nice, just the vendoring, so we can adjust the serialisation support as a follow-up.

PR #52, opened 2023-12-13
The in-memory aspect of `lookups` and comparative benchmarks

jorisdral:


bench.log
jorisdral, line 1: This file shouldn't be merged: it's only for reviewers to inspect. Also, see `analysis` in the micro-benchmarking code

lsm-tree.cabal
```diff
@@ -57,6 +57,7 @@ library
     , io-classes            ^>=1.2
     , lsm-tree:bloomfilter
     , primitive             ^>=0.8
+    , unix                  ^>=2.8
```
jorisdral, line 60: Doesn't work on Windows and MacOS -- maybe we should disable windows/macos in CI from now on, until we get to using `fs-sim` and `fs-api`

bench.log
jorisdral, line 1: This file shouldn't be merged: it's only for reviewers to inspect. Also, see `analysis` in the micro-benchmarking code

PR #67, opened 2024-01-16
Add Word64 backed BitVec and use it to read blobref bitmap

phadej:
I'll do CrumbVec for reading operations, and then lookup will be trivial to implement.


src/Database/LSMTree/Internal/BitVec.hs
```diff
@@ -0,0 +1,443 @@
...
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase                 #-}
+{-# LANGUAGE MagicHash                  #-}
+{-# LANGUAGE MultiParamTypeClasses      #-}
+{-# LANGUAGE TypeFamilies               #-}
+{-# LANGUAGE UnboxedTuples              #-}
+{-# LANGUAGE ViewPatterns               #-}
+
+-- | This is Data.Bit.Internal module from @bitvec@ package,
+-- however explicitly using 'Word64', and not machine native word.
```
jorisdral, line 15: Let's also specify somewhere in this module that the `Vector Bit` doesn't guarantee thread safety

src/Database/LSMTree/Internal/BitVec.hs
jorisdral, line 1: What would be problematic about using a BitVec that stores machine words instead of defaulting to `Word64`s?

jorisdral, line 1: I wonder how the performance of this `BitVec` compares to a `BitVec` using machine words. Not that we should benchmark it now, but it would be interesting to know!

dcoutts (approved):
Endianness strikes again!

PR #69, opened 2024-01-23
Arbitrary FormatPage generates unique and sorted pages

phadej:


dcoutts (approved):
Quite right. Very good.

PR #76, opened 2024-01-30
Make `SerialisedKey` sliceable

jorisdral:
A nice result of this is that we can go from `SerialisedKey` to the `RawPage`-style keys and back without copying the raw bytes. However, copying is still necessary if the `RawPage`-style key comes from a reusable buffer

### DIFF
--- a/lsm-tree.cabal
+++ b/lsm-tree.cabal
@@ -51,9 +51,9 @@ library
     Database.LSMTree.Normal
 
   build-depends:
-    , base                  >=4.16  && <4.19
+    , base                  >=4.16      && <4.19
     , bitvec                ^>=1.1
-    , bytestring
+    , bytestring            ^>=0.11.4.0
     , containers
     , deepseq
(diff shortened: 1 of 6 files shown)


mheinzel (approved):
Looks reasonable, no blockers from my side.

`sliceBits32` makes me a bit uneasy, so it's good that you added tests.

Some conversions (that you only slightly touched) seem like the could be a bit more direct, e.g. something like `byteArrayFromListN` instead of going through a builder and intermediate bytestring in `Serialise Word64`, but that's probably something for later.

jorisdral:
> Some conversions (that you only slightly touched) seem like the could be a bit more direct, e.g. something like byteArrayFromListN instead of going through a builder and intermediate bytestring in Serialise Word64, but that's probably something for later.

We'll add more performant versions to the main lib at some point

dcoutts (commented):
LGTM

PR #79, opened 2024-02-01
Change bloomfilter to use bitvec as its backend

phadej:
... instead of `array`

This removes most of bitfiddling from `bloomfilter`, making further changes easier to get right.

As we already use `bitvec` elsewhere, I don't see a reason to use it more.


mheinzel (approved):
Nice, seems like it will simplify things a bit. I don't expect a major impact on the benchmarks, but did you check?

phadej:
> but did you check?

I didn't. Given the other major changes upcoming, I don't see a point of benchmarking intermediate steps.

dcoutts (commented):
Yeah, very nice.

PR #87, opened 2024-02-08
Relax bloomfilter to allow non-power-of-2 sizes

phadej:
... size is rounded up to next 64 bits though.


phadej:
Fixed the literal

PR #92, opened 2024-02-12
Write checksum file during run construction

mheinzel:
Depends on #73

### DIFF
--- a/src/Database/LSMTree/Internal/CRC32C.hs
+++ b/src/Database/LSMTree/Internal/CRC32C.hs
@@ -20,6 +20,7 @@ module Database.LSMTree.Internal.CRC32C (
   hGetSomeCRC32C,
   hPutSomeCRC32C,
   hPutAllCRC32C,
+  hPutAllChunksCRC32C,
   readFileCRC32C,
 
   -- * Checksum files
@@ -101,6 +102,21 @@ hPutAllCRC32C fs h = go 0
       if BS.null bs'
         then return (written', crc')
         else go written' bs' crc'
+
+-- | This function makes sure that the whole /lazy/ 'BSL.ByteString' is written.
+hPutAllChunksCRC32C :: forall m h
+                    .  Monad m
+                    => HasFS m h
+                    -> Handle h
+                    -> BSL.ByteString
+                    -> CRC32C -> m (Word64, CRC32C)
+hPutAllChunksCRC32C fs h = \lbs crc ->
+    foldM (uncurry putChunk) (0, crc) (BSL.toChunks lbs)
+  where
+    putChunk :: Word64 -> CRC32C -> BS.ByteString -> m (Word64, CRC32C)
+    putChunk !written !crc !bs = do
+      (n, crc') <- hPutAllCRC32C fs h bs crc
+      return (written + n, crc')
 
 readFileCRC32C :: forall m h. MonadThrow m => HasFS m h -> FsPath -> m CRC32C
 readFileCRC32C fs file =
--- a/src/Database/LSMTree/Internal/Run/Mutable.hs
+++ b/src/Database/LSMTree/Internal/Run/Mutable.hs
@@ -51,6 +51,8 @@ type RefCount = IORef Int
 -- representation of a run. Each file handle is opened write-only and should be
 -- written to using normal buffered I\/O.
 --
+-- __Not suitable for concurrent construction from multiple threads!__
+--
 data MRun fhandle = MRun {
       -- | The reference count for the LSM run. This counts the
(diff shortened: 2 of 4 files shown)


src/Database/LSMTree/Internal/Run/Mutable.hs
```diff
@@ -51,29 +54,34 @@ type RefCount = IORef Int
 -- representation of a run. Each file handle is opened write-only and should be
 -- written to using normal buffered I\/O.
 --
+-- __Not suitable for concurrent construction from multiple threads!__
+--
```
jorisdral, line 58: Makes sense -- we can wrap the `MRun` type in something to handle concurrent access

src/Database/LSMTree/Internal/Run/Mutable.hs
```diff
@@ -216,3 +219,34 @@ writeIndexFinalChunk _ _ _ = return ()
...
+    readIORef checksum
+
+writeToHandle ::
+     HasFS IO h
+  -> ChecksumHandle (FS.Handle h) -> BSB.Builder -> IO ()
+writeToHandle fs (ChecksumHandle h checksum) builder = do
+    let bs = LBS.toStrict (BSB.toLazyByteString builder)
+    crc <- readIORef checksum
+    (_, crc') <- CRC.hPutAllCRC32C fs h bs crc
+    writeIORef checksum crc'
```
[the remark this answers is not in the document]
mheinzel: At the end, we want to work with strict bytestrings when interfacing with the C code, but using `toChunks` we can write the strict chunks without having to copy them once more. We don't currently have any benchmarks we could look at to compare different implementations, but I see there's a task about adding one.

I could just optimistically add a chunking variant into the CRC module and use that. Later on, we can look at performance more in-depth.

src/Database/LSMTree/Internal/Run/Mutable.hs
```diff
@@ -216,3 +219,34 @@ writeIndexFinalChunk _ _ _ = return ()
...
+-- Does not write anything, serialisation for the filter is not yet implemented.
+writeFilter :: HasFS IO h -> MRun (FS.Handle h) -> Bloom SerialisedKey -> IO ()
+writeFilter _ _ _ = return ()
+
+{-------------------------------------------------------------------------------
+  ChecksumHandle
+-------------------------------------------------------------------------------}
+
+-- | Tracks the checksum of a (write mode) file handle.
+data ChecksumHandle fhandle = ChecksumHandle !fhandle !(IORef CRC32C)
```
jorisdral, line 232: Maybe note here too that this isn't meant to be concurrently accessed

src/Database/LSMTree/Internal/Run/Mutable.hs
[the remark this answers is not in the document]
jorisdral: True, I had been thinking about that too -- we could leave it as is for now, but then let's add a TODO

src/Database/LSMTree/Internal/Run/Mutable.hs
mheinzel: I think it's fine, since this is only used internally within this module and the `MRun` itself has a note that it can only be used in a single-threaded manner.

src/Database/LSMTree/Internal/CRC32C.hs
```diff
@@ -102,6 +103,25 @@ hPutAllCRC32C fs h = go 0
...
+    foldM (uncurry putChunk) (0, crc) (BSL.toChunks lbs)
+  where
+    putChunk :: Word64 -> CRC32C -> BS.ByteString -> m (Word64, CRC32C)
+    putChunk !written !crc !bs = do
+      (n, crc') <- hPutSomeCRC32C fs h bs crc
+      let bs'      = BS.drop (fromIntegral n) bs
+          written' = written + n
+      if BS.null bs'
+        then return (written', crc')
+        else putChunk written' crc' bs'
```
jorisdral, line 123: Similar to how `hPutAll` for lazy bytestrings does it, you could reuse `hPutAllCRC32C` here
```suggestion
    foldM (uncurry putChunk) (0, crc) (BSL.toChunks lbs)
  where
    putChunk :: Word64 -> CRC32C -> BS.ByteString -> m (Word64, CRC32C)
    putChunk !written !crc !bs = do
      (n, crc') <- hPutAllCRC32C fs h bs crc
      pure $ (written + n, crc')
```

mheinzel: Oh, absolutely. Thanks!

jorisdral (approved):

PR #94, opened 2024-02-16
Cache dependencies even if `cabal build` or `cabal test` fails

jorisdral:
Also: update index-states and update versions for GHC, Cabal, `stylish-haskell`, `cabal-fmt`

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -52,8 +52,9 @@ jobs:
       run: |
         cabal build all --dry-run
 
-    - name: Cache cabal store
-      uses: actions/cache@v3
+    - name: "Restore cache"
+      uses: actions/cache/restore@v4
+      id: restore-cabal-cache
       env:
         cache-name: cache-cabal-build
       with:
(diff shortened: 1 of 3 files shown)


.github/workflows/haskell.yml
```diff
@@ -65,8 +66,17 @@ jobs:
           ${{ runner.os }}-${{ matrix.ghc }}-${{ env.cache-name }}-
 
     - name: Install cabal dependencies
+      id: build-dependencies
       run: cabal build --only-dependencies --enable-tests --enable-benchmarks all
 
+    - name: "Save cache"
+      uses: actions/cache/save@v4
+      id: save-cabal-cache
+      if: steps.build-dependencies.outcome == 'success' && steps.restore-cabal-cache.outputs.cache-hit != 'true'
```
phadej, line 75: you want to save cache also when stuff succeeded, because when we bump `index-state` the cache will be outdated.

or is `cache-hit` `'true'` only when the primary key is hit? If so, add a comment.

jorisdral: > Note: cache-hit will be set to true only when cache hit occurs for the exact key match. For a partial key match via restore-keys or a cache miss, it will be set to false.

I'll add a comment

dcoutts (approved):

PR #95, opened 2024-02-16
Compatibility with ghc-9.8.1

jorisdral:


jorisdral:
> Looks good and the `9.8.1` job also ran successfully.
> 
> I'm just wondering why it is not marked as required, since I don't see anything in the YAML file that would indicate it being different to the other jobs.

I'll mark it as required once the PR is merged, otherwise it would block other PRs from being mergeable

mheinzel:
Let me rephrase my question: Where are checks marked as required or not? But it seems like it's something to click on the GItHub UI.

jorisdral:
> Let me rephrase my question: Where are checks marked as required or not? But it seems like it's something to click on the GItHub UI.

Sorry, I misread. It's in the repository settings: Settings -> Branches -> main

PR #97, opened 2024-02-20
Set `-fno-ignore-asserts` in GHA

jorisdral:


.github/workflows/haskell.yml
```diff
@@ -45,7 +45,7 @@ jobs:
 
     - name: Configure the build
       run: |
-        cabal configure --enable-test --enable-benchmark --ghc-options="-Werror"
+        cabal configure --enable-test --enable-benchmark --ghc-options="-Werror" --ghc-options="-fno-ignore-asserts"
```
phadej, line 48: I never remember whether `--ghc-options` on command line affects all packages (i.e. also dependencies).

I'd rather do

```sh
cp cabal.project.ci.local cabal.project.local
```

with

```
package lsm-tree
  ghc-options: -Werror -fno-ignore-asserts
```

phadej: TL;DR

I advice to not ever use `cabal configure`. Or if you do, `cat cabal.project.local` afterwards.

jorisdral: https://cabal.readthedocs.io/en/stable/cabal-package.html#pkg-field-ghc-options doesn't specify, but https://cabal.readthedocs.io/en/stable/cabal-project.html#package-configuration-options does seem to provide some insight into the flag

> On the command line, these options are applied to all local packages. There is no per-package command line interface.

So I'm expecting it to apply only to local packages. I'm going to use `cat cabal.project.local` to verify

dcoutts (approved):

.github/workflows/haskell.yml
jorisdral: Indeed, it only applies to local packages

PR #98, opened 2024-02-20
Serialization of bloomfilter

phadej:


src/Database/LSMTree/Internal/Bloomfilter.hs
```diff
@@ -0,0 +1,68 @@
...
+    when (P.sizeofByteArray ba < 16) $ Left "doesn't contain a header"
+    -- on big endian platforms we'd need to byte-swap
+    let ver = P.indexPrimArray word32pa 0
+        hsn = P.indexPrimArray word32pa 1
+        len = P.indexPrimArray word32pa 2 -- length in bits
+        end = P.indexPrimArray word32pa 3
+
+    when (ver /= bloomFilterVersion) $ Left "Unsupported version"
+    when (end /= bloomFilterEndianess) $ Left "Non-matching endianess"
+    when (len .&. 0x6 /= 0) $ Left "Length is not multiply of 64"
```
mheinzel, line 57:
```suggestion
    when (len .&. 0x6 /= 0) $ Left "Length is not multiple of 64"
```

src/Database/LSMTree/Internal/Bloomfilter.hs
mheinzel, line 1: Everywhere else we spell it `BloomFilter`.

Also, I think this would fit into the submodule under `Run` (so `Database.LSMTree.Internal.Run.Bloomfilter`), like the compact index.

test/Test/Database/LSMTree/Internal/BloomFilter.hs
```diff
@@ -0,0 +1,27 @@
...
+    [ testProperty "roundtrip" roundtrip_prop
+    ]
+
+roundtrip_prop :: [Word64] -> Property
+roundtrip_prop ws =
+    counterexample (show sbs) $
+    Right lhs === rhs
+  where
+    -- 300 bits is just under 5x 64 bit words
+    lhs = BF.fromList 3 300 ws
```
mheinzel, line 25: Looks reasonable, but why only for a fixed size and number of hash functions?

mheinzel (commented):
A few small points, but everything else looks great to me.

src/Database/LSMTree/Internal/Bloomfilter.hs
phadej: or `Database.LSMTree.Internal.Run.BloomFilter`? ;)

I don't agree, this functionality is not run specific, i.e. it's adjacent

Different perspective: almost everything is `Run` specific, which makes that namespace crammed up.

I favor flat namespaces until they actually become too large too manage, which shouldn't be a concern for a while. I.e let the patterns emerge first, and only then organize if we need to.

phadej: ... and in fact there is `Database.LSMTree.Internal.Run.BloomFilter` with actually run specific stuff (size calculations).

mheinzel: Hehe, right.

I get your point. IMHO you could also move the current content of `Database.LSMTree.Internal.Run.BloomFilter` into your new module outside of the run, since it's also not really run-specific. Just having it in two different places feels wrong.

test/Test/Database/LSMTree/Internal/BloomFilter.hs
```diff
@@ -0,0 +1,27 @@
...
+tests = testGroup "Database.LSMTree.Internal.BloomFilter"
+    [ testProperty "roundtrip" roundtrip_prop
+    ]
+
+roundtrip_prop :: [Word64] -> Positive (Small Int) -> Positive Int -> Property
+roundtrip_prop ws (Positive (Small hfN)) (Positive bits) =
+    counterexample (show sbs) $
+    Right lhs === rhs
+  where
+    -- 300 bits is just under 5x 64 bit words
```
mheinzel, line 24: Now this comment is not needed any more. :stuck_out_tongue:

mheinzel (approved):

src/Database/LSMTree/Internal/Bloomfilter.hs
phadej: I'd move the relevant function to `Database.LSMTree.Internal.Monkey`, everything else in that module seems useless.

It's up to @jorisdral and @dcoutts 

---

Anyway, if you say they should be in one module, I'll move either stuff.

mheinzel: I would also want to hear their opinion, since it's the module structure they sketched out.

If you don't want to wait for them, you can just follow the existing structure and module header Haddock (i.e. move your new stuff to `Database.LSMTree.Internal.Run.BloomFilter`) and I'd be happy to get it merged.

phadej: Another example of unnecessarily long names is `Database.LSMTree.Internal.Serialise.RawBytes`, there's nothing serization specific in `RawBytes`. It's quite general datatype.

src/Database/LSMTree/Internal/ByteString.hs
```diff
@@ -0,0 +1,49 @@
...
+-- https://github.com/haskell/bytestring/issues/664
+{-# INLINE shortByteStringFromTo #-}
+shortByteStringFromTo :: Int -> Int -> ShortByteString -> BB.Builder
+shortByteStringFromTo = \i j sbs -> BB.builder $ shortByteStringCopyStepFromTo i j sbs
+
+-- | Like 'shortByteStringFromTo' but for 'ByteArray'
+--
+-- https://github.com/haskell/bytestring/issues/664
+--
+-- TODO: will break with newer bytestring
```
jorisdral, line 27: How so? Isn't it still a `ByteArray` in `bytestring-0.12`?

src/Database/LSMTree/Internal/BloomFilter.hs
```diff
@@ -0,0 +1,68 @@
...
+-----------------------------------------------------------
+
+-- | Read 'BF.Bloom' from a 'ShortByteString'.
+--
+-- In successful case the data portion of bloom filter is /not/ copied (the short bytestring has only 16 bytes of extra data in the header).
+--
+bloomFilterFromSBS :: ShortByteString -> Either String (BF.Bloom a)
+bloomFilterFromSBS (SBS ba') = do
+    when (P.sizeofByteArray ba < 16) $ Left "doesn't contain a header"
+    -- on big endian platforms we'd need to byte-swap
```
jorisdral, line 49: https://github.com/input-output-hk/lsm-tree/pull/100

src/Database/LSMTree/Internal/BloomFilter.hs
```diff
@@ -0,0 +1,68 @@
...
+
+-- deserializing
+-----------------------------------------------------------
+
+-- | Read 'BF.Bloom' from a 'ShortByteString'.
+--
+-- In successful case the data portion of bloom filter is /not/ copied (the short bytestring has only 16 bytes of extra data in the header).
+--
+bloomFilterFromSBS :: ShortByteString -> Either String (BF.Bloom a)
+bloomFilterFromSBS (SBS ba') = do
```
jorisdral, line 47: Did you pick a non-incremental deserialisation function for a specific reason? An incremental version could, for example, use a `MutableByteArray` for the `vec64`, which we can copy chunks into from a lazy byte string

src/Database/LSMTree/Internal/Bloomfilter.hs
jorisdral: It was a sketch, so I don't see a problem with moving definitions eventually. I agree that we can remove the `Run` namespace

> I favor flat namespaces until they actually become too large too manage, which shouldn't be a concern for a while. I.e let the patterns emerge first, and only then organize if we need to.

Sounds reasonable. I suggest large-scale re-organising in WP7, so we can avoid the churn of moving definitions for now. It's fine if PRs move things around a bit, as long as they don't cause a myriad merge conflicts

src/Database/LSMTree/Internal/ByteString.hs
phadej: I haven't tried. There `SBS` compatibility pattern, but without trying, one cannot know

src/Database/LSMTree/Internal/BloomFilter.hs
phadej: You can read the whole bloom filter file into `MutableByteArray` directly (and freeze it). The file of the size is known.

src/Database/LSMTree/Internal/BloomFilter.hs
```diff
@@ -0,0 +1,68 @@
...
+    when (P.sizeofByteArray ba < 16) $ Left "doesn't contain a header"
+    -- on big endian platforms we'd need to byte-swap
+    let ver = P.indexPrimArray word32pa 0
+        hsn = P.indexPrimArray word32pa 1
+        len = P.indexPrimArray word32pa 2 -- length in bits
+        end = P.indexPrimArray word32pa 3
+
+    when (ver /= bloomFilterVersion) $ Left "Unsupported version"
+    when (end /= bloomFilterEndianess) $ Left "Non-matching endianess"
+    when (len .&. 0x6 /= 0) $ Left "Length is not multiple of 64"
```
mheinzel, line 57:
```suggestion
    when (len .&. 0x3F /= 0) $ Left "Length is not multiple of 64"
```

mheinzel: Or better yet `0b11_1111`

src/Database/LSMTree/Internal/Bloomfilter.hs
jorisdral: Created an issue so we don't forget: https://github.com/input-output-hk/lsm-tree/issues/106#issue-2153874709

jorisdral (approved):
LGTM

dcoutts (commented):
LGTM too.

One thing we will need I think, is deserialising from a `ByteString` (rather than `ShortByteString`) because the fs-api uses ByteString for all I/O. This shouldn't be a problem in principle because we can convert without copying, though we'll need to add checks for the bytstring being of the bytearray variety, and for alignment.

PR #102, opened 2024-02-21
Batched lookups in IO

jorisdral:


jorisdral:
GHA fails on ghc's before 9.6. It is related to https://github.com/well-typed/quickcheck-lockstep/pull/19

src/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -30,3 +60,171 @@ prepLookup :: Run fd -> SerialisedKey -> Maybe PageSpan
...
+                -- vector. TODO: verify if appending always computes a new
+                -- primary vector
+                Just $ first (\(SerialisedValue v) -> SerialisedValue (v <> vOverflow)) e
+    pure $ resolves resolve1 kess
+
+forceEntry :: Entry SerialisedValue blobref -> Entry SerialisedValue blobref
+forceEntry (Insert v)            = Insert (forceSerialisedValue v)
+forceEntry (InsertWithBlob v br) = InsertWithBlob (forceSerialisedValue v) br
+forceEntry (Mupdate v)           = Mupdate (forceSerialisedValue v)
+forceEntry e@Delete              = e
```
mheinzel, line 162: Could this just use `first`?

src/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -30,3 +60,171 @@ prepLookup :: Run fd -> SerialisedKey -> Maybe PageSpan
...
+            let vOverflow = RawBytes (P.Vector
+                                        (unBufferOffset (ioopBufferOffset ioop) + 4096)
+                                        (fromIntegral m - 4096)
+                                        buf)
+            in  -- Laziness ensures that we only compute the appending of the
+                -- prefix and suffix when the result is needed. We do not use
+                -- 'force' here, since appending already creates a new primary
+                -- vector. TODO: verify if appending always computes a new
+                -- primary vector
+                Just $ first (\(SerialisedValue v) -> SerialisedValue (v <> vOverflow)) e
```
mheinzel, line 155: The two parts are already adjacent in the buffer, so I wonder if `LookupEntryOverflow` should just return a value offset, so we can construct (and force) a single vector. This would also avoid the potential footgun of it currently returning a `SerialisedValue` that is in fact only a prefix (so it can't be used in the same way as others). :thinking:

src/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -30,3 +60,171 @@ prepLookup :: Run fd -> SerialisedKey -> Maybe PageSpan
...
+-- TODO: optimise by reducing allocations, possibly looking at core, or
+-- revisiting the batching approach.
+lookupsInBatches ::
+     forall m h. (MonadAsync m, PrimMonad m, MonadThrow m)
+  => HasBlockIO m h
+  -> Resolve1
+  -> BatchSize
+  -> [Run (Handle h)]
+  -> [SerialisedKey]
+  -> m [(SerialisedKey, Maybe (Entry SerialisedValue BlobSpan))]
```
mheinzel, line 126: Returning only the `BlobSpan` here is problematic, since reading the actual blob requires knowing which blob file (or run) to read from, but that information is lost at this point. So we might want to return a `BlobRef`, or some internal reference type that can be turned into a `BlobRef` later.

mheinzel (commented):
I added a few thoughts, since I encountered many similar situations with run merging.

src/Database/LSMTree/Internal/Entry.hs
```diff
@@ -120,23 +124,55 @@ combine f   (Mupdate u)       (Insert v)              = Insert (f u v)
...
 
-combinesNormal :: NonEmpty (Entry v blob) -> Entry v blob
-combinesNormal = NE.head
+{-# SPECIALIZE combinesMonoidal' :: (v -> v -> v) -> NonEmpty (Maybe (Entry v blob)) -> Maybe (Entry v blob) #-}
+combinesMonoidal' ::
+     Foldable1 f
+  => (v -> v -> v)
+  -> f (Maybe (Entry v blob))
+  -> Maybe (Entry v blob)
+combinesMonoidal' f = foldr1 (combineMaybe f) -- short-circuit fold
```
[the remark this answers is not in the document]
mheinzel: It's more flexible of course, e.g. when switching to vectors. So makes sense!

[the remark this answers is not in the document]
jorisdral: This function has disappeared in more recent versions

src/Database/LSMTree/Internal/Lookup.hs
mheinzel: I don't know what the external/public `BlobRef` type should look like, but for the internal one to be used here (and in run merging) I don't see the need to index it by `m` or its "type" `blob`. Here we are dealing with serialised plain bytes. Everything else comes in the layer above.

Maybe something like `data BlobRef fhandle = BlobRef !(Run.Run fhandle) !BlobSpan`, defined locally in this module? That's what I use for run merging, too, but we can still pull it into a shared definition once we're sure the same representation really makes sense in both places.

jorisdral: I agree, the `m` type is probably not necessary, and we should instead index by the file handle type

It would be an advantage if we can unify the external and internal `BlobRef` types, so we don't have to do additional traversals of lookup results to convert internal `BlobRef` types to external `BlobRef` types. It sounds like `data BlobRef fhandle !(Run.Run fhandle) !blob` would work for your case too

mheinzel: So the `fhandle` would be the same one that is inside the run? Feels a bit redundant.

jorisdral: That was a typo :P it doesn't need that one

mheinzel: > It would be an advantage if we can unify the external and internal BlobRef types, so we don't have to do additional traversals of lookup results to convert internal BlobRef types to external BlobRef types.

So what is going to get returned to the user? I think I remember the users themselves have to use the reference to fetch the `SerialisedBlob` (they might only want to pay that cost for a few select ones), but currently you return `BlobRef _ SerialisedBlob`. Also, is there anything else than a `BlobSpan` in the ref, e.g. to track whether the reference is still valid?

mheinzel (commented):
(not done yet, just want these comments to be visible already)

src/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -165,20 +185,233 @@ indexSearches !rs !ks !rkixs = do
...
+  -> V.Vector SerialisedKey
+  -> m (V.Vector (Maybe (Entry SerialisedValue (BlobRef m SerialisedBlob))))
+lookupsInBatches hbio n rs blooms indexes kopsFiles ks = assert precondition $ do
+    (rkixs0, ioops0) <- prepLookups blooms indexes kopsFiles ks
+    res <- VM.replicate (V.length ks) Nothing
+    let batches = batchesOfN (unBatchSize n) rkixs0 ioops0
+    -- TODO: @rkixs@ is currently unused, but woudl come in handy when/if we
+    -- perform intra-page lookups asynchronously
+    ioress <- forConcurrently batches $ \(_rkixs, ioops) -> submitIO hbio ioops
+    -- TODO: perform intra page lookups asynchronously?
```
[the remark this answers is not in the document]
mheinzel: I think the idea is that each thread would immediately do the page lookups for the IO ops that finished, along the lines of like `forConcurrently _ $ submitIO >=> intraPageLookups`.

src/Database/LSMTree/Internal/Lookup.hs
jorisdral: The current types are:

* Public: https://github.com/IntersectMBO/lsm-tree/blob/d6cb82d02718347f0de1066be6abf62a97853fa2/src/Database/LSMTree/Common.hs#L264

* Internal: https://github.com/IntersectMBO/lsm-tree/blob/d6cb82d02718347f0de1066be6abf62a97853fa2/src/Database/LSMTree/Internal/BlobRef.hs#L41-L46

Whether we're going to need the `m` and `blob` parameters, specially for the internal type, that isn't clear. It should at least be in the public type so we know what type to deserialise the `SerialisedBlob` to, but maybe we can do without it internally

src/Database/LSMTree/Internal/Entry.hs
```diff
@@ -120,6 +122,29 @@ combine f   (Mupdate u)       (Insert v)              = Insert (f u v)
...
+combineMaybe _ Nothing  b        = b
+combineMaybe _ a        Nothing  = a
+combineMaybe f (Just a) (Just b) = Just (combine f a b)
+
+combineMonoidal ::
+     (v -> v -> v)
+  -> Maybe (Entry v blobref)
+  -> Maybe (Entry v blobref)
+  -> Maybe (Entry v blobref)
+combineMonoidal f = combineMaybe f
```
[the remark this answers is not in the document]
jorisdral: I've added this alias just to document that both `combineNormal` and `combineMonoidal` are (special) cases of `combineMaybe`, which is very much like a `Semigroup (Maybe (Entry v blobref)` instance.

src/Database/LSMTree/Internal/Lookup.hs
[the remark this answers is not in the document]
jorisdral: ^ what @mheinzel said. 

However, it just occurred to me that we're actually planing on making `blockio-uring` responsible for divvying up a mega batch into smaller batches that can be submitted to `io-uring`, in which case doing the intra-page lookups concurrently makes less sense

Furthermore, doing the intra-page lookups concurrently adds complexity. If we use a single result vector, then we have to orchestrate concurrent access to the vector, and if we a result vector per thread then we pay the cost of combining those result vectors in a single result vector in the end.

So I think it's safe to say that I can remove this TODO

src/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -165,20 +185,233 @@ indexSearches !rs !ks !rkixs = do
...
+    intraPageLookups rs ks rkixs0 ioops0 (VU.concat ioress) res
+    V.unsafeFreeze res
+  where
+    precondition = and [
+          V.length rs == V.length blooms
+        , V.length rs == V.length indexes
+        , V.length rs == V.length kopsFiles
+        ] -- TODO: alternatively, check `Eq` pairwise on the elements
```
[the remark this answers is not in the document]
jorisdral: We need the `rs` to create `BlobRef`s

src/Database/LSMTree/Internal/Lookup.hs
mheinzel: I don't think the internal blob ref needs any type parameters except `fhandle` or probably `run` instead, to avoid cyclic dependencies and also in case we want to use it with something else than `Run`.

The public one could then be `data BlobRef m blob = forall h. Eq h => BlobRef (Internal.BlobRef (Internal.Run h))`.

I'm not sure how feasible it is to unify both representations, especially if we need existential type variables and constraints like `Eq h`.

mheinzel: (has been addressed in #177)

src/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -166,20 +186,231 @@ indexSearches !rs !ks !rkixs = do
...
+            LookupEntry e         -> do
+                let e' = Just (bimap copySerialisedValue (mkBlobRef r) e)
+                unsafeModifyMMaybeStrict res (`combineNormal` e') kix
+            -- Laziness ensures that we only compute the appending of the prefix
+            -- and suffix when the result is needed. We do not use 'force' here,
+            -- since appending already creates a new primary vector.
+            --
+            -- TODO: verify if appending always computes a new primary vector
+            LookupEntryOverflow e m -> do
+                let v' (SerialisedValue v) = SerialisedValue $ v <>
```
mheinzel, line 358: Ah, so since `combine` often doesn't need to force its second argument, we don't pay for the copy, but if the entry ends up in `res` (i.e. potentially lives long), it is forced, so we can't keep references? That's a bit subtle, but neat.

src/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -166,20 +186,231 @@ indexSearches !rs !ks !rkixs = do
...
+      | V.length ioops == 0 = []
+      | otherwise =
+          let (ioops1, ioops2) = V.splitAt n ioops -- O(1)
+          in  ioops1 : go ioops2
+
+mkBlobRef :: Run h -> BlobSpan -> BlobRef (Run h)
+mkBlobRef run bspan = BlobRef {
+      blobRefRun = run
+    , blobRefSpan = bspan
+    }
```
mheinzel, line 416: `mkBlobRef = BlobRef` now, so might not be needed.

src/Database/LSMTree/Internal/Run.hs
```diff
@@ -40,6 +40,7 @@ module Database.LSMTree.Internal.Run (
   , Run (..)
   , addReference
   , removeReference
+  , close
```
mheinzel, line 43: This function is potentially dangerous, since it could close a run that others still have references to. If we export it, maybe it should be called `unsafeClose`. On the other hand, I don't think you really need it. Usually  `removeReference` should be enough, e.g. to clean up  a run you just created in the tests.

test/Test/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -245,6 +270,46 @@ prop_inMemRunLookupAndConstruction dat =
...
+  where
+    mkRuns hasFS = first V.fromList . unzip <$> sequence [
+          (,wb) <$> Run.fromWriteBuffer hasFS (Run.RunFsPaths i) wb
+        | (i, dat) <- zip [0..] (getSmallList dats)
+        , let wb = WB.WB (runData dat)
+        ]
+    lookupss = concatMap lookups dats
+
+opaqueifyBlobs :: V.Vector (k, Maybe (Entry v b)) -> V.Vector (k, Maybe (Entry v Opaque))
+opaqueifyBlobs = fmap (fmap (fmap (fmap Opaque)))
```
mheinzel, line 311: peak Haskell :trophy:

test/Test/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -348,48 +446,62 @@ tabulateInMemLookupData dat run =
...
          | runData' <- liftShrink2Map shrinkKey shrinkEntry runData
          , Map.size runData' > 0 ]
       ++ [ InMemLookupData runData lookups'
          | lookups' <- liftShrink shrinkKey lookups ]
     where
-      shrinkEntry = liftShrink2 shrinkValue shrink
+      shrinkEntry = liftShrink2 shrinkValue shrinkBlob
 
 genSerialisedKey :: Gen SerialisedKey
 genSerialisedKey = arbitrary `suchThat` (\k -> sizeofKey k >= 6)
```
mheinzel, line 507: At some point we can think about unifying approaches for generation of these core types, but might make sense to wait until we see what @dcoutts has come up with his `LogicalPage` related generators.

mheinzel (commented):
Most of my comments are:
1. suggestions for the benchmark, which can still be done later, or you could even move the whole benchmark to a follow-up PR
2. notes and ideas on the main lookup code, which you might like or not, in which case you can safely discard them

So most things are not a blocker, but before merging it would be good to at least add some TODOs and ideally avoid the export of `Run.close`.

bench/micro/Bench/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -106,79 +124,166 @@ deriving anyclass instance NFData h => NFData (RunLookupView h)
...
+    | otherwise = pick (n-k) xs
+  pick _ _  = error "QuickCheck.pick used with empty list"
+
+randomEntry :: StdGen -> (Entry UTxOValue UTxOBlob, StdGen)
+randomEntry g = frequency [
+      (10, \g' -> let (v, g'') = uniform g' in (Insert v, g''))
+      -- TODO: implement blobs once blob retrieval is implemented.
+    , (0,  \g' -> let (v, g'') = uniform g'
+                      (b, g''') = error "random blob not implemened" $ g''
+                  in  (InsertWithBlob v b, g'''))
```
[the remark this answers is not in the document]
jorisdral: Good point. Before, when I still had an undefined for `mkBlobRef` in place, this would have made the benchmark fail, but that's no longer the case

bench/micro/Bench/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -106,79 +124,166 @@ deriving anyclass instance NFData h => NFData (RunLookupView h)
...
+    let nentriesReal = unNumEntries $ Run.runNumEntries r
+    assert (nentriesReal == nentries) $ pure ()
+    let npagesReal = Index.sizeInPages (Run.runIndex r)
+    assert (npagesReal * 42 <= nentriesReal) $ pure ()
+    assert (npagesReal * 43 >= nentriesReal) $ pure ()
+    pure ( benchTmpDir
+         , hasFS
+         , hasBlockIO
+         , BatchSize 64
+         , V.singleton r
```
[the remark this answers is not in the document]
jorisdral: The plan is to adapt the macro-benchmark for bloomfilters to do `indexSearches` and `prepLookups` too. I don't think we're currently planning (yet) on benchmarking the IO lookup part for multiple runs. We'll be doing that in a macro-benchmark too.

src/Database/LSMTree/Internal/Run.hs
jorisdral: Using `removeReference` now

test/Test/Database/LSMTree/Internal/Lookup.hs
jorisdral: Wait until you meet their cousin... 
```haskell
opaqueifyBlobs :: V.Vector (k, Maybe (Entry v b)) -> V.Vector (k, Maybe (Entry v Opaque))
opaqueifyBlobs = getCompose . getCompose . getCompose . fmap Opaque . Compose . Compose . Compose
```

test/Test/Database/LSMTree/Internal/Lookup.hs
jorisdral: Let's wait for @dcoutts

src/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -166,20 +186,231 @@ indexSearches !rs !ks !rkixs = do
...
+            -- TODO: verify if appending always computes a new primary vector
+            LookupEntryOverflow e m -> do
+                let v' (SerialisedValue v) = SerialisedValue $ v <>
+                      RawBytes (mkPrimVector
+                                  (unBufferOffset (ioopBufferOffset ioop) + 4096)
+                                  (fromIntegral m)
+                                  buf)
+                    e' = Just $ bimap v' (mkBlobRef r) e
+                unsafeModifyMMaybeStrict res (`combineNormal` e') kix
+          loop res (ioopix + 1)
```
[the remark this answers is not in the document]
mheinzel: Looks good! (the `M` is not really need in the name IMO)

bench/micro/Bench/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -10,93 +11,113 @@
...
+              whnfAppIO (\ks' -> prepLookups blooms indexes kopsFiles ks') ks
+          , bench "Batched lookups in IO" $
+              nfAppIO (\ks' -> lookupsInBatches hasBlockIO bsize resolveV rs blooms indexes kopsFiles ks') ks
           ]
+  where
+    withEnv = envWithCleanup
+                (lookupsInBatchesEnv conf)
+                lookupsInBatchesCleanup
+    -- TODO: pick a better value resolve function
+    resolveV = \v1 _v2 -> v1
```
mheinzel, line 97: I think it's fine to just use `const`, we're not too focused on `Mupdate` performance for now.

mheinzel (approved):
This looks ready to me!

PR #110, opened 2024-02-28
Make deserialisation of Bloom filter and compact index safe

mheinzel:


src/Database/LSMTree/Internal/BloomFilter.hs
```diff
@@ -61,8 +61,13 @@ bloomFilterFromSBS (SBS ba') = do
     -- limit to 2^48 bits
     when (len >= 0xffff_ffff_ffff) $ Left "Too large bloomfilter"
 
+    -- make sure byte array is long enough for vector
+    let len64 = fromIntegral (div64 len)
+    when (2 + len64 > div8 (P.sizeofByteArray ba)) $
```
jorisdral, line 66: Something I realised just now: if the byte array is much larger than the part of it that the bloom filter uses, then we are still keeping that unused part alive. Maybe we should check that the size of byte array is *exactly* `(2 + len64) * 8)`? Or that it is within a small bound? Or, possibly, we slice the byte array

src/Database/LSMTree/Internal/Run/Index/Compact.hs
```diff
@@ -697,28 +697,43 @@ putPaddingTo64 written
...
+    let len64 = div8 len8
+    let getPositive offset64 = do
+          let w = indexByteArray ba offset64 :: Word64
+          when (w > fromIntegral (maxBound :: Int)) $
+            Left "Size information is too large for Int"
+          return (fromIntegral w)
 
+    rfprec <- getPositive (len64 - 3)
+    numPages <- getPositive (len64 - 2)
+    numEntries <- getPositive (len64 - 1)
```
jorisdral, line 719: Similar idea to my previous comment: if there is a large slice of the bytearray that is unused between the data and the footer, then we're keeping those bytes live. Maybe we need an exact check of file length vs. these parameters?

src/Database/LSMTree/Internal/Run/Index/Compact.hs
```diff
@@ -697,28 +697,43 @@ putPaddingTo64 written
...
+    let (!offset2_32, ciPrimary) = getVec32 ba offset1_32 numPages
+    let (!offset3_32, ciRangeFinder) = getVec32 ba offset2_32 numRanges
     -- offsets in 64 bits
-    let offset3 = ceilDiv2 offset3_32
-    let (offset4, ciClashes) = getBitVec ba offset3 numPages
-    let (offset5, ciLargerThanPage) = getBitVec ba offset4 numPages
-    let ciTieBreaker = getTieBreaker ba offset5
+    let !offset3 = ceilDiv2 offset3_32
+    let (!offset4, ciClashes) = getBitVec ba offset3 numPages
+    let (!offset5, ciLargerThanPage) = getBitVec ba offset4 numPages
```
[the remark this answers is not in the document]
mheinzel: I'll add one for `P.Vector` and one for `BitVec64`.

src/Database/LSMTree/Internal/Run/Index/Compact.hs
```diff
@@ -737,26 +752,38 @@ getBitVec ba offset64 numEntries = (offset64', vec)
...
+        let offset8 = mul8 offset64
+        let offset64' = offset64 + ceilDiv8 len8
+        when (offset64' > to64) $ Left "Clash map key is out of bounds"
+
+        -- We avoid retaining references to the bytearray.
+        -- Probably not needed, since the bytearray will stay alive as long as
+        -- the compact index anyway, but we expect very few keys in the tie
+        -- breaker, so it is cheap and we don't have to worry about it any more.
+        let !key = SerialisedKey' (VP.force (VP.Vector offset8 len8 ba))
+        return (offset64', key)
```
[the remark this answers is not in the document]
mheinzel: No, since the length of the keys is not known.

test/Test/Database/LSMTree/Internal/BloomFilter.hs
```diff
@@ -31,3 +39,30 @@ roundtrip_prop (Positive (Small hfN)) (limitBits -> bits) ws =
...
 limitBits b = b .&. 0xffffff
+
+prop_total_deserialisation :: [Word32] -> Property
+prop_total_deserialisation word32s =
+    case bloomFilterFromSBS sbs of
+      Left err -> label err $ property True
+      Right bf -> label "parsed successfully" $ property $
+        -- Just forcing the filter is not enough. The underlying vector might
+        -- point to invalid memory, so we try accessing it.
+        bf `deepseq` bitVecToList (BF.bitArray bf) `deepseq` True
```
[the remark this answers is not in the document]
mheinzel: Which I wanted to find out in the tests. But you're right, I remember thinking that this is not the proper way of doing it and I should instead check the bounds against the underlying byte array.

src/Database/LSMTree/Internal/BloomFilter.hs
mheinzel: It's more restrictive than necessary, but the way we use it it should always be true, I think. So might make sense to add an assertion/check.

src/Database/LSMTree/Internal/Run/Index/Compact.hs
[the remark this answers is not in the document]
mheinzel: I decided not to test the smart constructors themselves, as the test would just check exactly the same invariants as the smart constructors do. However, they're still tested through `fromSBS`.

src/Database/LSMTree/Internal/BloomFilter.hs
mheinzel: I added a check.

src/Database/LSMTree/Internal/Run/Index/Compact.hs
mheinzel: (done)

mheinzel:
I finally got back to this PR. I'm still not really happy with it, but I think the most important feedback has been addressed and I should not spend too much time here.

src/Database/LSMTree/Internal/Run/Index/Compact.hs
```diff
@@ -691,72 +693,130 @@ putPaddingTo64 written
...
-            keyOffset8 = mul8 (offset64 + 1)
-            -- We avoid retaining references to the bytearray.
-            -- Probably not needed, since the bytearray will stay alive as long
-            -- as the compact index anyway, but we expect very few keys in the
-            -- tie breaker, so it is cheap to do and we don't have to worry
-            -- about the issue anymore.
-            !key = SerialisedKey' (VP.force (VP.Vector keyOffset8 keyLen8 ba))
-        in (key, PageNo pageNo)
-         : getEntries (n - 1) (offset64 + 1 + ceilDiv8 keyLen8)
+-- | Offset and length are in number of bits.
```
jorisdral, line 810: Technically, also "in number of elements" :P

src/Database/LSMTree/Internal/Run/Index/Compact.hs
```diff
@@ -691,72 +693,130 @@ putPaddingTo64 written
...
-    let (offset3_32, ciRangeFinder) = getVec32 ba offset2_32 (2 ^ rfprec + 1)
+    let off1_32 = 1  -- after version indicator
+    (!off2_32, ciPrimary) <- getVec32 "Primary array" ba off1_32 numPages
+    (!off3_32, ciRangeFinder) <- getVec32 "Range finder" ba off2_32 numRanges
     -- offsets in 64 bits
-    let offset3 = ceilDiv2 offset3_32
-    let (offset4, ciClashes) = getBitVec ba offset3 numPages
-    let (offset5, ciLargerThanPage) = getBitVec ba offset4 numPages
-    let ciTieBreaker = getTieBreaker ba offset5
+    let !off3 = ceilDiv2 off3_32
```
jorisdral, line 730:
```suggestion
    when (mod2 off3_32 /= 0) $ Left "Not enough padding after range finder component" 
    let !off3 = ceilDiv2 off3_32
```

src/Database/LSMTree/Internal/Run/Index/Compact.hs
```diff
@@ -691,72 +693,130 @@ putPaddingTo64 written
...
+      Nothing  -> Left (name <> " is out of bounds")
+      Just vec -> Right (off + ceilDiv64 numEntries, vec)
+
+-- | Checks bounds.
+--
+-- Inefficient, but okay for a small number of entries.
+getTieBreaker ::
+     ByteArray -> Offset64
+  -> Either String (Offset64, Map SerialisedKey PageNo)
+getTieBreaker ba = \off -> do
```
jorisdral, line 769: The offsets in this function aren't converted from `Word64` to `Word8` offsets

test/Test/Database/LSMTree/Internal/BloomFilter.hs
```diff
@@ -31,3 +40,28 @@ roundtrip_prop (Positive (Small hfN)) (limitBits -> bits) ws =
...
+prop_total_deserialisation_whitebox :: Word32 -> Small Word32 -> Property
+prop_total_deserialisation_whitebox hsn (Small len64) =
+      forAll (vector (fromIntegral len64 * 2)) $ \word32s ->
+        prop_total_deserialisation (prefix <> word32s)
+  where
+    prefix =
+      [ 1 {- version -}
+      , hsn
+      , unsafeShiftL len64 6         -- len64 * 64 (lower 32 bits)
+      , unsafeShiftR len64 (32 - 6)  -- len64 * 64 (upper 32 bits)
```
jorisdral, line 66:
```suggestion
      , unsafeShiftR len64 (32 - 6)  -- len64 / 64 (upper 32 bits)
```

test/Test/Database/LSMTree/Internal/Run/Index/Compact.hs
```diff
@@ -287,6 +297,36 @@ prop_roundtrip csize ps numEntries =
...
+prop_total_deserialisation_whitebox :: RFPrecision -> Small Word16 -> Small Word16 -> [Word32] -> Property
+prop_total_deserialisation_whitebox (RFPrecision rfprec) numEntries numPages word32s =
+    prop_total_deserialisation $
+         [1]  -- version
+      <> word32s  -- primary array, range finder, clash bits, LTP bits, clash map
+      <> [0 | even (length word32s)]  -- padding
+      <> [ fromIntegral rfprec, 0
+         , fromIntegral numPages, 0
+         , fromIntegral numEntries, 0
+         ]
```
jorisdral, line 328: This deserialises succesfully only with a very small probability. However, since we're doing roundtrip serialisation/deserialisation tests elsewhere, that's probably not a bad thing. Since the serialisation format for the compact index is relatively complex (compared to the bloom filter, for example), it would be quite hard anyway to generate inputs that make the parse succeed

jorisdral (approved):
Finished my second round of reviewing. I've approved the PR, provided that you address the small comments I left this round

Some follow-up ideas:
* Add golden tests for (de)serialisation
* Expand `bloomInvariant`: for example, with the upper bound on `numBits`
* Add more assertions in places where `bloomInvariant` is relevant

test/Test/Database/LSMTree/Internal/Run/Index/Compact.hs
mheinzel: Yes, I tried for a little bit, but also realised that the error cases are the more important ones here.

test/Test/Database/LSMTree/Internal/BloomFilter.hs
mheinzel: No, it's multiplication (left shift), but since we only take the upper 32 bits, it turns into an overall right shift.

src/Database/LSMTree/Internal/Run/Index/Compact.hs
mheinzel: Yes, it should be `mul 8 off >= sizeofByteArray ba`. Good catch!

src/Database/LSMTree/Internal/Run/Index/Compact.hs
mheinzel: If the 32 bit offset after the tie breaker is not 64 bit aligned, that just means there are 32 bits of padding before the bit vectors start. This is what `ceilDiv` handles and it's totally expected.

mheinzel:
test output:

```
  Database.LSMTree.Internal.BloomFilter
    total-deserialisation:                 OK (0.25s)
      +++ OK, passed 10000 tests:
      83.08% Unsupported version
      16.56% Doesn't contain a header
       0.34% Byte array is too small for components
       0.01% Byte array is too large for components
       0.01% parsed successfully
    total-deserialisation-whitebox:        OK (0.48s)
      +++ OK, passed 10000 tests:
      94.98% parsed successfully
       5.02% Length is zero
  Test.Database.LSMTree.Internal.Run.Index.Compact
    (De)serialisation
      prop_total_deserialisation:          OK (0.21s)
        +++ OK, passed 10000 tests:
        48.16% Length is not multiple of 64 bit
        35.04% Unsupported version
        16.78% Doesn't contain header and footer
         0.02% Invalid range finder precision
      prop_total_deserialisation_whitebox: OK (0.26s)
        +++ OK, passed 10000 tests:
        47.32% Range finder is out of bounds
        28.11% Primary array is out of bounds
         9.63% Doesn't contain header and footer
         6.66% Clash map key is out of bounds
         4.88% Clash map entry is out of bounds
         1.33% LTP bit vector is out of bounds
         1.18% Clash bit vector is out of bounds
         0.89% Byte array is too small for components
```

test/Test/Database/LSMTree/Internal/BloomFilter.hs
jorisdral: Hmm okay I think I misunderstood the calculation, but it's clear now

src/Database/LSMTree/Internal/Run/Index/Compact.hs
jorisdral: You're right

PR #115, opened 2024-03-05
Always import the `RawBytes` module qualified

jorisdral:


mheinzel (approved):
There's a little bit of tension between choosing identifiers that work well when imported qualified and on the other hand going for similarity to the primitive vector API, e.g. `sizeofRawBytes` could just be `size`.
Still makes sense overall I think.

phadej (changes requested):
> There's a little bit of tension between choosing identifiers that work well when imported qualified and on the other hand going for similarity to the primitive vector API, e.g. `sizeofRawBytes` could just be `size`. Still makes sense overall I think.

Indeed.

- `sizeofRawBytes` -> `size`
- `rawBytes` -> `builder` or `toBuilder`
-  `takeRawBytes` -> `take`
- `hashRawBytes` -> `hash`
- `mkRawBytes` -> `fromByteArray`

And I insist that the renaming is done simultaneously, as it will touch most of the same use-sites as importing qualified.

phadej:
```haskell
{- | This module is intended to be imported qualified, to avoid name clashes with "Prelude" functions, e.g.

@
 import           Database.LSMTree.Internal.Serialise.RawBytes (RawBytes)
 import qualified Database.LSMTree.Internal.Serialise.RawBytes as RB
@
-}
module Database.LSMTree.Internal.Serialise.RawBytes (
    ..
```

jorisdral:
@phadej @mheinzel valid feedback. I've updated the PR, let me know what you think

mheinzel:
Looks good to me.

PR #116, opened 2024-03-06
Change bloomfilter serialization format

phadej:
- Make length field 64bit (for the future)
- Encode endianess in the version tag


dcoutts (commented):
LGTM!

PR #124, opened 2024-03-13
Add singletonPage

phadej:


src/Database/LSMTree/Internal/PageAcc1.hs
```diff
@@ -0,0 +1,95 @@
...
+
+-- | Create a singleton page, also returning the overflow value bytes.
+singletonPage
+    :: SerialisedKey
+    -> Entry SerialisedValue BlobSpan
+    -> (RawPage, RawBytes)
+singletonPage k (Insert v) = runST $ do
+    -- allocate bytearray
+    ba <- P.newByteArray pageSize :: ST s (P.MutableByteArray s)
+    P.fillByteArray ba 0 pageSize 0
```
mheinzel, line 30: Makes the code more obviously correct, good. You could leave a TODO to try optimising it away later (during benchmarking & optimisation phase), if you think it makes sense.

mheinzel (approved):
There's the thing about handling `Mupdate`, which I agree would be good to address now that you're working on this code already. Apart from that, everything looks very sensible to me.

src/Database/LSMTree/Internal/PageAcc1.hs
phadej: I don't think it makes. `memset` is cheap.

phadej:
Ah, good point. `Mupdate` has value too, I forgot about that. I'll add that too.

dcoutts (approved):
LGTM

Issue #128, opened 2024-03-19
Fix sporadically failing tests in GHA on MacOS


jorisdral:
First noticed in #120 : https://github.com/input-output-hk/lsm-tree/actions/runs/8263348980/job/22604567176
Recent example in #103: https://github.com/input-output-hk/lsm-tree/actions/runs/8344452814/job/22836988078?pr=103

The logs don't show why the tests fail. It's probably hard to debug, since one would need access to a MacOS system, and even then the error seems not occur reliably.

PR #132, opened 2024-03-25
Dont mavx2 on OSX

phadej:
I'm curious whether CI will pass if we dont avx2


jorisdral:
This would resolve #128

jorisdral (approved):

PR #136, opened 2024-03-26
Move kmerge benchmarks into non-test-suite component

phadej:
benchmarks take non-deterministic time to run.

`tasty-bench` test methodology is crappy, as on noisy machine it easily uses time up to the timeout. Maybe we should only use `criterion` (which is crappy too, but at least it has strict time limits)


dcoutts (approved):
LGTM

PR #142, opened 2024-03-27
Set tasty timeout to 5 minutes, set GHA timeout to 60 minutes

jorisdral:


dcoutts (approved):

.github/workflows/haskell.yml
```diff
@@ -103,8 +105,19 @@ jobs:
...
 
+    - name: Set test timeout (Unix)
+      if: ${{ startsWith(matrix.os, 'ubuntu') || startsWith(matrix.os, 'macOS') }}
+      run: |
+        export TASTY_TIMEOUT="5m"
+
+    - name: Set test timeout (Windows)
+      if: ${{ startsWith(matrix.os, 'windows') }}
+      run: |
+        $env:TASTY_TIMEOUT="5m"
```
phadej, line 116: these are not preserved.

phadej: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

PR #149, opened 2024-04-04
Hardcode expected allocations per key in bloom macro bench

phadej:
Also move debug part of cabal.project into separate import, so it's easier to disable.

Add a CPP, so we know if debug stuff is on.

### DIFF
--- a/cabal.project.debug
+++ b/cabal.project.debug
@@ -1,0 +1,19 @@
+package lsm-tree
+  -- apply this to all components
+  -- relevant mostly only for development & testing
+  ghc-options: -fno-ignore-asserts
+
+  -- there are no cpp-options in cabal.project files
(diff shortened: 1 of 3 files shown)


cabal.project.debug
```diff
@@ -0,0 +1,19 @@
+package lsm-tree
+  -- apply this to all components
+  -- relevant mostly only for development & testing
+  ghc-options: -fno-ignore-asserts
+
+  -- there are no cpp-options in cabal.project files
+  ghc-options: -optP -DNO_IGNORE_ASSERTS
```
jorisdral, line 7: :+1:

jorisdral (approved):

dcoutts (approved):
LGTM

PR #150, opened 2024-04-04
Remove redundant HasCallStack constraint

jorisdral:
The constraint is redundant when assertions are optimised away.


phadej:
I wonder why `hlint` job doesn't fail, even it finds stuff.

dcoutts (approved):

PR #154, opened 2024-04-05
`submitIO` using vectors

jorisdral:
Update `blockio-uring` to the latest version, and adapt the `fs-api-blockio` code to use vectors instead of lists

### DIFF
--- a/fs-api-blockio/src-linux/System/FS/BlockIO/Async.hs
+++ b/fs-api-blockio/src-linux/System/FS/BlockIO/Async.hs
@@ -1,4 +1,6 @@
-{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE BangPatterns        #-}
+{-# LANGUAGE NamedFieldPuns      #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 
 module System.FS.BlockIO.Async (
     asyncHasBlockIO
@@ -6,7 +8,10 @@ module System.FS.BlockIO.Async (
 
 import           Control.Exception
 import qualified Control.Exception as E
-import           Control.Monad
+import           Control.Monad.Primitive
+import qualified Data.Vector as V
+import qualified Data.Vector.Unboxed as VU
+import qualified Data.Vector.Unboxed.Mutable as VUM
 import           Foreign.C.Error
 import           GHC.IO.Exception
 import           GHC.Stack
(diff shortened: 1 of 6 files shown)


fs-api-blockio/src-linux/System/FS/BlockIO/Async.hs
```diff
@@ -97,3 +101,29 @@ ioopConv (IOOpWrite h off buf bufOff c) = handleFd h >>= \fd ->
...
+-- | Heterogeneous blend of `V.zipWithM` and `VU.zipWithM`
+--
+-- The @vector@ package does not provide functions that take distinct vector
+-- containers as arguments, so we write it by hand to prevent having to convert
+-- one vector type to the other.
+hzipWithM ::
+     (IOOp IO HandleIO -> I.IOResult -> IO IOResult)
+  -> V.Vector (IOOp IO HandleIO)
+  -> VU.Vector I.IOResult
+  -> IO (VU.Vector IOResult)
```
mheinzel, line 114: The implementation is also not a based on bundles, which is probably sensible here.

I like that you split it out as a separate function that has a familiar interface. I think it would be even nicer if it was a little more polymorphic, i.e. took a function `a -> b -> m c`.

fs-api-blockio/src/System/FS/BlockIO/API.hs
```diff
@@ -60,5 +75,30 @@ ioopHandle :: IOOp m h -> Handle h
...
+ioopBuffer (IOOpRead _ _ buf _ _)  = buf
+ioopBuffer (IOOpWrite _ _ buf _ _) = buf
+
+ioopBufferOffset :: IOOp m h -> BufferOffset
+ioopBufferOffset (IOOpRead _ _ _ bufOff _)  = bufOff
+ioopBufferOffset (IOOpWrite _ _ _ bufOff _) = bufOff
+
+ioopByteCount :: IOOp m h -> ByteCount
+ioopByteCount (IOOpRead _ _ _ _ c)  = c
+ioopByteCount (IOOpWrite _ _ _ _ c) = c
```
mheinzel, line 93: Not directly related to this PR, but would it make sense to turn `IOOp` into a flat record with a field `IOOpType = IOOPRead | IOOPWrite`?
Then we could just have field accessors. It also enables the `Constructed Product Result` optimisation, but that's probably not too relevant here, since the `IOOps` need to allocated for the vector either way.

fs-api-blockio/src/System/FS/BlockIO/Serial.hs
```diff
@@ -37,15 +42,24 @@ close :: MonadMVar m => IOCtx m -> m ()
...
+    res <- VUM.unsafeNew n
+    loop res 0
+  where
+    !n = V.length ioops
+    loop !res !i
+      | i == n = VU.unsafeFreeze res
+      | otherwise = do
+          iores <- ioop hfs hbfs (ioops `V.unsafeIndex` i)
+          VUM.unsafeWrite res i iores
+          loop res (i+1)
```
mheinzel, line 62: That could be a local `vmapM`. IMO it helps understand the code.

mheinzel (approved):
Looks good to me. Small notes on vector operations again, but no strong opinions. Do what you think is sensible.

fs-api-blockio/src/System/FS/BlockIO/API.hs
mheinzel: Makes sense to keep it as separate constructors!

PR #155, opened 2024-04-06
Improve WriteBuffer generator improvements

### LINKED ISSUE #137: Why WriteBuffer generator has invariant?
Why the invariant is not in the library? Why it's restricting the keys? Why we are restricting keys only *there*.

This is probably related to compact index requiring 6+ byte keys, but that restriction seems to leak into wrong places.

mheinzel:
Fixes #137.

Also, empty bloom filters aren't useful, but make sense to have in tests.


bloomfilter/src/Data/BloomFilter/Easy.hs
```diff
@@ -88,7 +88,7 @@ safeSuggestSizing
     -> Double           -- ^ desired false positive rate (0 < /e/ < 1)
     -> Either String (Word64, Int)
 safeSuggestSizing (fromIntegral -> capacity) errRate
-    | capacity <= 0                = Left "invalid capacity"
+    | capacity < 0                 = Left "invalid capacity"
```
phadej, line 91: Note, the size will be always at least one though: https://github.com/IntersectMBO/lsm-tree/blob/fbd59d98477cbb684c581911096a60d38cca39e3/bloomfilter/src/Data/BloomFilter/Mutable.hs#L67

This is worth a comment why the negative capacity is allowed

mheinzel: > Note, the size will be always at least one

In fact, with this function, it's at least 63. For very small `n`, creating a bloom filter is not super useful. I could add a comment like this:

-- Note that creating bloom filter with extremely small capacity is
-- often not very useful. This function will always suggest at least
-- 63 bits, even if the capacity is 0.

> negative capacity

Do you mean zero?

phadej: > Do you mean zero?

Yes.

---

But I also wondere why not simply

```
| capacity <= 0 = Right (63, 1)
```

why allow zero, but not negative `Int`s?

phadej: Except that `fromIntegral` converts argument to `Word64`, so there is nasty overflow happening, maybe

mheinzel: Which `fromIntegral` do you mean? The ones here all convert to `Double` I think.

I'm leaning towards not allowing negative capacities. A bloom filter with 0 elements is not useful, but possible. A bloom filter with -1 elements is just not a thing.

phadej: `take negative elements` is also not a thing, yet

```haskell
Prelude> take (-3) "foobar"
""
```

works.

phadej (commented):
FWIW. Why we use `Easy` module anywhere?

mheinzel:
Good question.

The Monkey calculations say:

```
-- These functions are experimental, and will not yet guarantee correct false
-- positive rates. For now, use 'Easy.suggestSizing' and 'Easy.safeSuggestSizing' instead.
--
-- TODO: un-experimental these functions.
```

Switching over seems to be scheduled for WP9. :shrug:

bloomfilter/src/Data/BloomFilter/Easy.hs
mheinzel: Makes sense, no strong preference either way. I'll change it.

jorisdral (approved):

PR #157, opened 2024-04-09
Module reorganisation

### LINKED ISSUE #106: Review module organisation

jorisdral:
WIP

Resolves #106 

My current suggestion is to organise the `lsm-tree` package roughly as follows:
```
|-- src
|   |-- lib (library lsm-tree)
|   |-- core (library lsm-tree-core)
|   |-- utils (library lsm-tree-utils)
|   |-- prototypes (library prototypes)
|-- test (test-suite lsm-tree-test)
|-- bench
    |-- micro (benchmark lsm-tree-micro-bench)
    |-- macro (benchmark lsm-tree-macro-bench)
```
* `src`
  * `lsm-tree`: public API
  * `lsm-tree-core`: internal modules containing the core functionality
  * `lsm-tree-utils`: internal utility modules and orphan instances, such as `Arbitrary` and `NFData`
  * `prototypes`: internal modules for prototyping 
* `test`: all tests for both public and private (sub-)libraries
* `bench`: micro- and macro-benchmarks for both public and private (sub-)libraries

Other changes:
* Get rid of the `Internal` module namespace
* Move internal modules (the ones that were previously in the `Internal` namespace) from `lsm-tree` to `lsm-tree-core`
* Flatten module namespaces where possible, such as moving `*.Serialise.RawBytes` to `*.RawBytes`

We also have a number of other libraries, test-suites and benchmarks that are less `lsm-tree` specific, like:
* `xxhash*`
* `bloomfilter*`
* `kmerge*`
* `map-range-test`
* `lsm-prototypes-tests`
* `fs-api-blockio*`

We should decide on a case by case basis if we want to factor them out as separate packages or include them in the reorganised module hierarchy


jorisdral:
This PR is parked for now. We'll only flatten the module hierarchy, and decide on some of the other aspects of re-organisation later

PR #158, opened 2024-04-09
Move `BloomFilter` module out of `Run` namespace and rename it to `Monkey`

jorisdral:


phadej (changes requested):
I don't like this.

This module now mixes serialization code and some size calculations.

These are unrelated, and should stay in separate modules.

(

jorisdral:
> I don't like this.
> 
> This module now mixes serialization code and some size calculations.
> 
> These are unrelated, and should stay in separate modules.
> 
> (

I suppose I could add a `Monkey` module. Is that what you had in mind?

phadej:
> I suppose I could add a `Monkey` module. Is that what you had in mind?

Monkey sounds great!

dcoutts (approved):
LGTM!

PR #164, opened 2024-04-10
Rename `MCompactIndex `to `CompactIndexAcc`

jorisdral:


dcoutts (approved):
I'm happy to approve as-is, but I'd also be happy to flatten it further and keep the sort-order, i.e. call the module and the type: `IndexCompact` / `IndexCompactAcc`.

jorisdral:
> I'm happy to approve as-is, but I'd also be happy to flatten it further and keep the sort-order, i.e. call the module and the type: `IndexCompact` / `IndexCompactAcc`.

I've done the renaming -- I was planning on doing it, but hadn't gotten around to it yet

mheinzel (approved):
Looks good!

PR #165, opened 2024-04-10
Fix the index-state for formatting jobs in GHA

jorisdral:
Without a fixed index-state and with only one unique cache name before, GHA is likely to rebuild dependencies from scratch each time we run the formatting jobs. Fixing the index-state and using the value in the primary key of the GHA caches, we won't rebuild dependencies superfluously any longer, and we have control over when we want to build newer versions of our formatting tools, or update their dependencies.

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -132,6 +132,16 @@ jobs:
         cabal: ["3.10.2.1"]
         os: [ubuntu-latest]
 
+    # Fix the index-state so we can get proper caching effects. Change this to a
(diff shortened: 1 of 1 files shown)


dcoutts (approved):
Ok, but now we've got the index state specified in two places, the GHA config and the cabal.project file. Do we now have to remember to update both?

Would it be possible set the GHA one from the cabal.project file one?

phadej:
>  Do we now have to remember to update both?

No.

We need to update formatting `index-state` only when we bump the versions of formatting tools (which is quite rare)

PR #166, opened 2024-04-11
Move `Run.Construction` and `Run.Mutable`, rename `MRun` to `RunBuilder`

jorisdral:


src/Database/LSMTree/Internal/RunBuilder.hs
```diff
@@ -3,19 +3,19 @@
 {-# LANGUAGE RankNTypes        #-}
 {-# LANGUAGE RecordWildCards   #-}
 
--- | Mutable runs ('MRun') that are under construction.
+-- | Mutable runs ('RunBuilder') that are under construction.
```
mheinzel, line 6:
```suggestion
-- | A mutable run ('RunBuilder') that is under construction.
```

Since you're touching it already, I'd argue that's clearer.

src/Database/LSMTree/Internal/RunBuilder.hs
```diff
@@ -60,49 +60,49 @@ type RefCount = IORef Int
...
-    , lsmMRunAcc        :: !(RunAcc ST.RealWorld)
+    , lsmRunBuilderAcc        :: !(RunAcc ST.RealWorld)
 
       -- | The byte offset within the blob file for the next blob to be written.
-    , lsmMRunBlobOffset :: !(IORef Word64)
+    , lsmRunBuilderBlobOffset :: !(IORef Word64)
 
       -- | The (write mode) file handles.
-    , lsmMRunHandles    :: {-# UNPACK #-} !(ForRunFiles (ChecksumHandle fhandle))
+    , lsmRunBuilderHandles    :: {-# UNPACK #-} !(ForRunFiles (ChecksumHandle fhandle))
```
mheinzel, line 82: I think I'd like to remove the `lsm` field prefixes. They're only here and on `Run` itself and to me they just add noise without much benefit.
Do you think it makes sense to remove them? If so, you could do it for `RunBuilder` here (since you touch them already) and I could make a PR for `Run` a little later?

test/Test/Database/LSMTree/Internal/RunAcc.hs
```diff
@@ -21,15 +21,15 @@ import           Database.LSMTree.Internal.RawOverflowPage (RawOverflowPage)
 import qualified Database.LSMTree.Internal.RawOverflowPage as RawOverflowPage
 import           Database.LSMTree.Internal.RawPage (RawPage)
 import qualified Database.LSMTree.Internal.RawPage as RawPage
-import           Database.LSMTree.Internal.Run.Construction as Real
+import           Database.LSMTree.Internal.RunAcc as Real
```
mheinzel, line 24:
```suggestion
import           Database.LSMTree.Internal.RunAcc
```

The `Real` synonym is not used, not a qualified import.

mheinzel (approved):
Looks good, just had some further suggestions since you are touching the code anyway.

src/Database/LSMTree/Internal/RunBuilder.hs
jorisdral: Sounds good

PR #172, opened 2024-04-15
Remove lsm- prefix of Run record fields

mheinzel:
As mentioned in https://github.com/IntersectMBO/lsm-tree/pull/166#discussion_r1562569738, this is a similar change as for `RunBuilder`. No other types have these prefixes. Since we're refactoring a few things at the moment, it seems like a good opportunity to make this change.


jorisdral (approved):
LGTM

PR #173, opened 2024-04-16
Add instances for SerialiseKey and SerialiseValue ByteArray and Short…

phadej:
…ByteString

In benchmarks it's easier (and faster) to generate bytearrays directly

### DIFF
--- a/src/Database/LSMTree/Internal/RawBytes.hs
+++ b/src/Database/LSMTree/Internal/RawBytes.hs
@@ -35,6 +35,7 @@ module Database.LSMTree.Internal.RawBytes (
     -- | Use 'Semigroup' and 'Monoid' operations
     -- ** Restricting memory usage
   , copy
+  , force
     -- * Conversions
   , fromVector
   , fromByteArray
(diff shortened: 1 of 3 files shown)


src/Database/LSMTree/Internal/RawBytes.hs
```diff
@@ -218,6 +220,19 @@ instance Monoid RawBytes where
 copy :: RawBytes -> RawBytes
 copy (RawBytes pvec) = RawBytes (PV.force pvec)
 
+-- | Force 'RawBytes' to not retain any extra memory. This may copy the contents.
+force :: RawBytes -> ByteArray
+force (RawBytes (PV.Vector off len ba))
+    | off == 0
+    , BA.sizeofByteArray ba == len
+    = ba
```
mheinzel, line 228: Is there still a good reason to use `copy` over this? I don't see when we would require a byte array to be copied if it doesn't retain unnecessary memory, but I guess we could avoid the branch in situations where we already know we're dealing with a slice.

If there are reasons, it would be nice to have a comment explaining when to use which. If not, you could just replace `copy`.

mheinzel (approved):
It's still not quite clear where these (and similar) instances really belong, but I think adding them is uncontroversial, so IMO we can merge this and move things around afterwards. :+1:

src/Database/LSMTree/Internal/RawBytes.hs
mheinzel: Okay, let's keep them both.

PR #174, opened 2024-04-16
Initial lsm bench

phadej:


bench/macro/initial-lsm-bench.hs
```diff
@@ -0,0 +1,112 @@
...
+
+benchRun :: Opts -> FilePath -> IO ()
+benchRun opts sessionRoot = do
+    let fs = FsIO.ioHasFS (FS.MountPoint sessionRoot)
+    -- flush write buffer
+    let wb :: WB.WriteBuffer K V B
+        !wb = foldl'
+            (\wb' i -> WB.addEntryNormal (makeKey i) (N.Insert theValue Nothing) wb')
+            WB.empty
+            [ 1 .. fromIntegral (opts.size) :: Word64 ]
```
jorisdral, line 77: It might be interesting to investigate how we could generate an initial table with a mix of inserts of fresh values, and deletes of existing values

jorisdral: With only inserts, we end up with a "perfect" table

bench/macro/initial-lsm-bench.hs
```diff
@@ -0,0 +1,112 @@
...
+withSessionDir :: (FilePath -> IO a) -> IO a
+withSessionDir = withTempDirectory "" "session"
+
+-------------------------------------------------------------------------------
+-- Opts
+-------------------------------------------------------------------------------
+
+data Opts = Opts
+    { size :: Int
+    -- TODO: do we need a flag to not remove temporary directory?
```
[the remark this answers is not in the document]
phadej: >  so if we can skip generating it from scratch every time that would save some time

That benchmark would be setup differently. Probably best to have a generation completely separate (run separately before the benchmark).

If we can make the benchmark executable do only what we are interested in benchmarking, then using external tools for collecting data will be a lot easier.

We don't want to fiddle with subtracting costs of generating initial data etc.

PR #175, opened 2024-04-17
Move `Serialise*` instances to the core library

jorisdral:


phadej (dismissed):
There are dependencies / imports of `Orphans ()` just for these instances. They may be removed as well.

jorisdral:
> There are dependencies / imports of `Orphans ()` just for these instances. They may be removed as well.

Good point. I'm doing this in https://github.com/IntersectMBO/lsm-tree/pull/176/commits/252c6bc2c28ccb69da734a8750542ae4212b77e7

mheinzel (approved):
It's not visible on the commit linked above, but removing the orphan imports is part of https://github.com/IntersectMBO/lsm-tree/pull/176.

In that case, the two PRs together look fine to me.

PR #177, opened 2024-04-18
Remove unused type parameters for internal `BlobRef` type

jorisdral:


src/Database/LSMTree/Common.hs
```diff
@@ -261,4 +261,32 @@ mkSnapshotName s
...
+-- blob. Note that table operations such as snapshotting and duplication do
+-- /not/ invalidate blob references. These operations do not modify the logical
+-- contents or state of an existing table.
+--
+-- [Blob reference validity] as long as the table handle that the blob reference
+-- originated from is not updated or closed, the blob reference will be valid.
+--
+-- TODO: get rid of the @m@ parameter?
+type BlobRef :: (Type -> Type) -> Type -> Type
+data BlobRef m blob = forall h. Eq h => BlobRef (Internal.BlobRef (Internal.Run h))
```
mheinzel, line 292: We might want to give the phantom type parameters nominal role to prevent coercions breaking deserialisation.

src/Database/LSMTree/Internal/BlobRef.hs
```diff
@@ -1,45 +1,21 @@
...
---
--- [Blob reference validity] as long as the table handle that the blob reference
--- originated from is not updated or closed, the blob reference will be valid.
-type BlobRef :: (Type -> Type) -> Type -> Type -> Type
-data BlobRef m run blob = BlobRef {
+-- See 'Database.LSMTree.Common.BlobRef' for more info.
+data BlobRef run = BlobRef {
       blobRefRun    :: !run
     , blobRefOffset :: !Word64
     , blobRefSize   :: !Word32
```
mheinzel, line 21: Do you want to use `BlobSpan` here, since we often construct from a span or need to select the span? We could use an `UNPACK` pragma, but it might cause some re-packing and ultimately lead to more allocations.

mheinzel (commented):
Some quick suggestions.

mheinzel (approved):

PR #179, opened 2024-04-18
Rename MPageAcc to PageAcc

mheinzel:
For consistency. The Acc suffix already indicates mutability, so no need to have a prefix.


mheinzel:
Not sure spelling related renamings are worth it at this point (so feel free to tell me to drop that commit), but I added it as I was touching some places where it occurred.

jorisdral (approved):

PR #180, opened 2024-04-18
improve Arbitrary instances for serialised keys and values

mheinzel:
Not particularly elegant, but improves the chance of hitting interesting cases until we have a more principled approach.

Output:

```
  Test.Database.LSMTree.Generators
    key/op pairs
      prop_distributionKOps: OK (2.19s)
        +++ OK, passed 800 tests:
        57.2% any k/op is large
        32.5% has key with >2 collisions
        23.4% >10% of keys collide
        
        k/op is large (19686 in total):
        95.205% False
         4.795% True
        
        k/op sizes (19686 in total):
        38.535% 16 <= n < 64
        22.803% 64 <= n < 256
        13.629% 256 <= n < 1024
        10.556% 4 <= n < 16
         9.728% 1024 <= n < 4096
         4.750% 4096 <= n < 16384
        
        key counts (18044 in total):
        95.140% 1
         2.455% 2
         1.313% 3
         0.643% 4
         0.255% 5
         0.127% 6
         0.044% 7
         0.017% 8
         0.006% 10
        
        key sizes (19686 in total):
        52.845% 16 <= n < 64
        32.871% 4 <= n < 16
        14.284% 64 <= n < 256
        
        value sizes (18128 in total):
        25.359% 16 <= n < 64
        23.003% 4 <= n < 16
        14.447% 256 <= n < 1024
        11.557% 1 <= n < 4
        10.255% 1024 <= n < 4096
         5.787% n == 0
         5.136% 4096 <= n < 16384
         4.457% 64 <= n < 256
```

### DIFF
--- a/test/Test/Database/LSMTree/Generators.hs
+++ b/test/Test/Database/LSMTree/Generators.hs
@@ -9,14 +9,23 @@ module Test.Database.LSMTree.Generators (
   ) where
 
 import           Control.DeepSeq (NFData, deepseq)
+import           Data.Bifoldable (bifoldMap)
+import           Data.Coerce (coerce)
+import qualified Data.Map.Strict as Map
 import qualified Data.Vector.Primitive as PV
 import           Data.Word (Word64, Word8)
 
+import           Database.LSMTree.Extras (showPowersOf)
 import           Database.LSMTree.Extras.Generators
+import           Database.LSMTree.Internal.BlobRef (BlobSpan)
+import           Database.LSMTree.Internal.Entry
+import           Database.LSMTree.Internal.PageAcc (entryWouldFitInPage,
+                     sizeofEntry)
 import           Database.LSMTree.Internal.RawBytes (RawBytes (..))
+import           Database.LSMTree.Internal.Serialise
 
-import           Test.QuickCheck (Arbitrary (..), Gen, Testable (..),
-                     forAllShrink)
+import qualified Test.QuickCheck as QC
+import           Test.QuickCheck (Arbitrary (..), Gen, Property, Testable (..))
 import           Test.Tasty (TestTree, testGroup)
 import           Test.Tasty.QuickCheck (testProperty)
 
@@ -35,6 +44,10 @@ tests = testGroup "Test.Database.LSMTree.Generators" [
      ++ prop_arbitraryAndShrinkPreserveInvariant (deepseqInvariant @RawBytes)
     , testGroup "KeyForIndexCompact" $
         prop_arbitraryAndShrinkPreserveInvariant keyForIndexCompactInvariant
+    , testGroup "lists of key/op pairs" $
+        [ testProperty "prop_distributionKOps" $
+            prop_distributionKOps
+        ]
     ]
 
 prop_arbitraryAndShrinkPreserveInvariant ::
@@ -46,9 +59,9 @@ prop_forAllArbitraryAndShrinkPreserveInvariant ::
     forall a. Show a => Gen a -> (a -> [a]) -> (a -> Bool) -> [TestTree]
 prop_forAllArbitraryAndShrinkPreserveInvariant gen shr inv =
     [ testProperty "Arbitrary satisfies invariant" $
-            property $ forAllShrink gen shr inv
+            property $ QC.forAllShrink gen shr inv
     , testProperty "Shrinking satisfies invariant" $
-            property $ forAllShrink gen shr (all inv . shr)
+            property $ QC.forAllShrink gen shr (all inv . shr)
     ]
 
 -- | Trivial invariant, but checks that the value is finite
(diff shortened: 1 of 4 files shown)


src/utils/Database/LSMTree/Generators.hs
```diff
@@ -619,9 +631,14 @@ newtype KeyForIndexCompact =
...
-  arbitrary =
-      fmap KeyForIndexCompact . genRawBytesN
-          =<< QC.sized (\s -> QC.chooseInt (6, s + 6))
+  -- we try to make collisions and close keys more likely (very crudely)
+  arbitrary = KeyForIndexCompact <$> frequency
+      [ (5, genRawBytesN =<< QC.sized (\s -> QC.chooseInt (6, s + 6)))
+      , (1, do
+          lastByte <- QC.chooseBoundedIntegral (0, 3)
+          return (RB.pack ([1,3,3,7,0] <> [lastByte]))
+        )
```
jorisdral, line 640: This generates only 3 unique keys, which seems to be a little on the small side. How about something like this:

```suggestion
      , (1, do
          first4Bytes <- serialiseKey . getSmall <$> QC.arbitrary @Word64 
          then2Bytes <- QC.arbitrary @Word16
          otherBytes <- QC.arbitrary
          pure $ first4Bytes <> then2Bytes <> otherBytes
        )
```

jorisdral (approved):
Sounds like a step in the right direction. LGTM modulo some suggestions

src/utils/Database/LSMTree/Generators.hs
jorisdral: It's close to what I used before in the `CompactIndex` tests

test/Test/Database/LSMTree/Generators.hs
```diff
@@ -58,3 +71,28 @@ deepseqInvariant x = x `deepseq` True
 prop_packRawBytesPinnedOrUnpinned :: Bool -> [Word8] -> Bool
 prop_packRawBytesPinnedOrUnpinned pinned ws =
     packRawBytesPinnedOrUnpinned pinned ws == RawBytes (PV.fromList ws)
+
+type TestEntry = Entry SerialisedValue BlobSpan
+type TestKOp = (KeyForIndexCompact, TestEntry)
+
+prop_distributionKOps :: [TestKOp] -> Property
+prop_distributionKOps kops' =
+    QC.tabulate "key counts" (map (show . snd) (Map.assocs keyCounts)) $
```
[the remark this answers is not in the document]
mheinzel: A bit hard to phrase clearly, so I went a bit more verbose: "key occurrences (>1 is collision)"

src/utils/Database/LSMTree/Util.hs
```diff
@@ -1,13 +1,20 @@
...
 showPowersOf10 :: Int -> String
-showPowersOf10 n0
-  | n0 <= 0   = "n == 0"
-  | n0 == 1   = "n == 1"
-  | otherwise = go n0 1
+showPowersOf10 = showPowersOf 10
+
+showPowersOf :: Int -> Int -> String
+showPowersOf factor n
+  | n < 0     = "n < 0"
```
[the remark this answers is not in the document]
mheinzel: In fact, factor 1 is already problematic.

src/utils/Database/LSMTree/Generators.hs
mheinzel: I get zero collisions with that. In the compact index tests, the size parameter is increased a lot, which makes collisions more likely, but ideally we already see some of them at small sizes. We also want multiple collisions, e.g. to test resolving >2 entries in run merging or lookups.

Having a single fixed prefix under which collisions appear is a bit limited, I know, but maybe we can at least have it such that potentially more than 4 values are possible, but some of them are more likely than others (non-uniform distribution).

I just tried this:

```haskell
  arbitrary = KeyForIndexCompact <$> frequency
      [ (6, genRawBytesN =<< QC.sized (\s -> QC.chooseInt (6, s + 6)))
      , (1, do
          lastByte <- QC.sized $ skewedWithMax . fromIntegral
          return (RB.pack ([1,3,3,7,0] <> [lastByte]))
        )
      ]
      where
        -- generates a value in range from 0 to ub, but skewed
        skewedWithMax ub0 = do
          ub1 <- QC.chooseBoundedIntegral (0, ub0)
          ub2 <- QC.chooseBoundedIntegral (0, ub1)
          QC.chooseBoundedIntegral (0, ub2)
```

Allows for last bytes up to max size (usually 100 I think), but at small size limits that range and generally heavily skews towards smaller values. With that we get:

```
        10.1% has key with >2 collisions
         3.2% >10% of keys collide
```

Feels a bit hacky, but do you think that's better?

dcoutts (approved):
LGTM. No comments beyond what @jorisdral suggested.

test/Test/Database/LSMTree/Generators.hs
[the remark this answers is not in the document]
jorisdral: It hasn't changed

src/utils/Database/LSMTree/Generators.hs
jorisdral: Sounds good, we can use this for now

jorisdral: Different idea that I thought of just now, which we don't have to do in this PR: if it's interactions between two separately generated keys that we care about, then maybe we should generate a collection of keys for the compact index instead, so we can precisely decide when to generate two colliding keys or not. Something like:

```haskell
newtype KeysForIndexCompact = KeysForIndexCompact { getKeysForCompactIndex :: [SerialisedKey] }
instance Arbitrary KeysForIndexCompact where ...
```

jorisdral (approved):

src/utils/Database/LSMTree/Generators.hs
mheinzel: Added it.

PR #181, opened 2024-04-19
Add property tests for instances of SerialiseKey and SerialiseValue

mheinzel:
There was one bug in the instance for Word64. Now there is just one issue left: We require that instances should preserve the ordering of their un-serialised representation, but [`ByteArray` doesn't use lexicographic ordering](https://www.stackage.org/haddock/lts-22.7/base-4.18.2.0/src/Data.Array.Byte.html#line-212). So either:

1. the instance shouldn't exist (one could use a newtype wrapper or `RawBytes` directly, if we want to expose that)
2. ~~we do some hack with serialising the length as the first byte, but that's ugly and slow~~
3. we drop the requirement on instances

I feel like option 3 might be possible, in which case it seems preferable. I think the original reasoning was about the ordering in the write buffer, but we already use `SerialisedKey` there, so I think we're good?

### DIFF
--- a/test/Test/Database/LSMTree/Internal/Serialise/Class.hs
+++ b/test/Test/Database/LSMTree/Internal/Serialise/Class.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications    #-}
 
@@ -7,38 +8,40 @@ import           Data.ByteString (ByteString)
 import           Data.ByteString.Lazy (LazyByteString)
 import           Data.ByteString.Short (ShortByteString)
 import           Data.Primitive (ByteArray)
+import           Data.Proxy (Proxy (Proxy))
 import           Data.Word
 import           Database.LSMTree.Extras.Generators ()
+import qualified Database.LSMTree.Internal.RawBytes as RB
 import           Database.LSMTree.Internal.Serialise.Class
 import           Test.Tasty
 import           Test.Tasty.QuickCheck
 
 tests :: TestTree
-tests = testGroup "Test.Database.LSMTree.Internal.Serialise.Class" [
-      testProperty "prop_roundtripSerialiseKey @Word64" $
-        prop_roundtripSerialiseKey @Word64
-    , testProperty "prop_roundtripSerialiseValue @Word64" $
-        prop_roundtripSerialiseValue @Word64
+tests = testGroup "Test.Database.LSMTree.Internal.Serialise.Class"
+    [ testGroup "Word64"          (allProperties @Word64)
+    , testGroup "ByteString"      (allProperties @ByteString)
+    , testGroup "LazyByteString"  (allProperties @LazyByteString)
+    , testGroup "ShortByteString" (allProperties @ShortByteString)
+    , testGroup "ByteArray"       (allProperties @ByteArray)
+    ]
 
-    , testProperty "prop_roundtripSerialiseKey @ByteString" $
-        prop_roundtripSerialiseKey @ByteString
-    , testProperty "prop_roundtripSerialiseValue @ByteString" $
-        prop_roundtripSerialiseValue @ByteString
+allProperties :: forall a. (Ord a, Show a, Arbitrary a, SerialiseKey a, SerialiseValue a) => [TestTree]
+allProperties = keyProperties @a <> valueProperties @a
 
-    , testProperty "prop_roundtripSerialiseKey @LazyByteString" $
-        prop_roundtripSerialiseKey @LazyByteString
-    , testProperty "prop_roundtripSerialiseValue @LazyByteString" $
-        prop_roundtripSerialiseValue @LazyByteString
+keyProperties :: forall a. (Ord a, Show a, Arbitrary a, SerialiseKey a) => [TestTree]
+keyProperties =
+    [ testProperty "prop_roundtripSerialiseKey" $
+        prop_roundtripSerialiseKey @a
+    , testProperty "prop_orderPreservationSerialiseKey" $
+        prop_orderPreservationSerialiseKey @a
+    ]
 
-    , testProperty "prop_roundtripSerialiseKey @ShortByteString" $
-        prop_roundtripSerialiseKey @ShortByteString
-    , testProperty "prop_roundtripSerialiseValue @ShortByteString" $
-        prop_roundtripSerialiseValue @ShortByteString
-
-    , testProperty "prop_roundtripSerialiseKey @ByteArray" $
-        prop_roundtripSerialiseKey @ByteArray
-    , testProperty "prop_roundtripSerialiseValue @ByteArray" $
-        prop_roundtripSerialiseValue @ByteArray
+valueProperties :: forall a. (Ord a, Show a, Arbitrary a, SerialiseValue a) => [TestTree]
+valueProperties =
+    [ testProperty "prop_roundtripSerialiseValue" $
+        prop_roundtripSerialiseValue @a
+    , testProperty "prop_concatDistributesSerialiseValue" $
+        prop_concatDistributesSerialiseValue @a
     ]
 
 prop_roundtripSerialiseKey :: forall k. (Eq k, Show k, SerialiseKey k) => k -> Property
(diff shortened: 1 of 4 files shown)


phadej:
There is https://github.com/IntersectMBO/lsm-tree/blob/ce56074726084428bdcfb319c7931890718357bd/test/Test/Database/LSMTree/Common.hs and https://github.com/IntersectMBO/lsm-tree/blob/ce56074726084428bdcfb319c7931890718357bd/src/Database/LSMTree/Common.hs#L184

Which I find now very confusing.

test/Test/Database/LSMTree/Internal/Serialise/Class.hs
```diff
@@ -0,0 +1,53 @@
...
+          prop_roundtripSerialiseValue @a
+      , testProperty "prop_orderPreservationSerialiseKey" $
+          prop_orderPreservationSerialiseKey @a
+      , testProperty "prop_orderPreservationSerialiseValue" $
+          prop_orderPreservationSerialiseValue @a
+      ]
+
+prop_roundtripSerialiseKey :: (Eq k, Show k, SerialiseKey k) => k -> Property
+prop_roundtripSerialiseKey k =
+    deserialiseKey (serialiseKey k) === k
```
phadej, line 41: We have serialiseKeyIdentity serialiseKeyPreservesOrdering, sadly there are somewhat useless in tests, aren't they?

mheinzel: Ah, how did I miss them? I could use `serialiseValueIdentity`, it just doesn't print the mismatch as `===` does. And I should probably add a test for `serialiseValueConcatDistributes` (or something like it).

mheinzel:
> There is https://github.com/IntersectMBO/lsm-tree/blob/ce56074726084428bdcfb319c7931890718357bd/test/Test/Database/LSMTree/Common.hs and
> 
> https://github.com/IntersectMBO/lsm-tree/blob/ce56074726084428bdcfb319c7931890718357bd/src/Database/LSMTree/Common.hs#L184
> 
> Which I find now very confusing.

What do you find confusing? How we managed to do it right (tests + correct implementation) for `SomeSerialisationConstraint` (which is a placeholder, to be removed), but then didn't do it properly for the "real" code?

phadej:
I find it

> > There is https://github.com/IntersectMBO/lsm-tree/blob/ce56074726084428bdcfb319c7931890718357bd/test/Test/Database/LSMTree/Common.hs and
> > https://github.com/IntersectMBO/lsm-tree/blob/ce56074726084428bdcfb319c7931890718357bd/src/Database/LSMTree/Common.hs#L184
> > 
> > Which I find now very confusing.
> 
> What do you find confusing? How we managed to do it right (tests + correct implementation) for `SomeSerialisationConstraint` (which is a placeholder, to be removed), but then didn't do it properly for the "real" code?

I find it confusing that we have both `SomeSerialisationConstraint` and `SerialiseKey`.

test/Test/Database/LSMTree/Internal/Serialise/Class.hs
```diff
@@ -0,0 +1,53 @@
...
+
+prop_roundtripSerialiseValue :: (Eq k, Show k, SerialiseValue k) => k -> Property
+prop_roundtripSerialiseValue v =
+    deserialiseValue (serialiseValue v) === v
+
+prop_orderPreservationSerialiseKey :: (Ord k, SerialiseKey k) => k -> k -> Property
+prop_orderPreservationSerialiseKey x y =
+    compare x y === compare (serialiseKey x) (serialiseKey y)
+
+prop_orderPreservationSerialiseValue :: (Ord k, SerialiseValue k) => k -> k -> Property
```
phadej, line 51: there isn't requirement for this, is there?

mheinzel: The property is currently specified in the Haddock: https://github.com/IntersectMBO/lsm-tree/blob/9f7ac2896cba4563d4135307cb36ac7957daa58c/src/Database/LSMTree/Internal/Serialise/Class.hs#L37

But I also think there is no *real* requirement to have that property and and would be best to just get rid of it (option 3 I listed). Just not sure others agree, there might be some reason I'm missing.

phadej: these are `SerialiseKey` haddocks.

mheinzel:
> I find it confusing that we have both `SomeSerialisationConstraint` and `SerialiseKey`.

Yes, it was an initial draft, but it doesn't need to exist anymore. That's why I started removing it. Should be done with the PR in a moment!

test/Test/Database/LSMTree/Internal/Serialise/Class.hs
```diff
@@ -0,0 +1,53 @@
...
+allProperties name =
+    testGroup name
+      [ testProperty "prop_roundtripSerialiseKey" $
+          prop_roundtripSerialiseKey @a
+      , testProperty "prop_roundtripSerialiseValue" $
+          prop_roundtripSerialiseValue @a
+      , testProperty "prop_orderPreservationSerialiseKey" $
+          prop_orderPreservationSerialiseKey @a
+      , testProperty "prop_orderPreservationSerialiseValue" $
+          prop_orderPreservationSerialiseValue @a
```
jorisdral, line 36: `SerialiseValue` doesn't have to preserve ordering

test/Test/Database/LSMTree/Internal/Serialise/Class.hs
jorisdral: They are *yet* unused. My intent here was to expose these functions as part of the public API, so users can check whether their serialisation functions are good. This is similar to https://hackage.haskell.org/package/serialise-0.2.6.1/docs/Codec-Serialise-Properties.html

jorisdral: > Ah, how did I miss them? I could use `serialiseValueIdentity`, it just doesn't print the mismatch as `===` does. And I should probably add a test for `serialiseValueConcatDistributes` (or something like it).

If you want to redefine the properties in terms of `===` that should be fine. Maybe add an assertion that checks wether `serialiseKeyIdentity` and this property agree?

test/Test/Database/LSMTree/Internal/Serialise/Class.hs
jorisdral: Yes, this property can be removed

jorisdral (changes requested):

jorisdral:
Regarding the `ByteArray` question, that's a good question. The order-preservation property exists for the compact index so we can compare bit representations of keys instead of the actual keys. If we were to add a "simple" index later, then the property would be redundant when you use that simple index. Will we then have two serialisation classes? Should users provide instances for both even if they only use the simple index? I don't think we have to answer those questions now. It's okay to stick with the order-preservation property for now, since we don't need more generality right now

1. :+1:  Users can define their own `ByteArray` wrapper, or use `ShortByteString`
2. Prepending the length also means that the bit representation isn't going to preserve uniformity, which means we wouldn't want to use these keys in the compact index anyway
3. Not possible

test/Test/Database/LSMTree/Internal/Serialise/Class.hs
mheinzel: Yes, I shouldn't use it for values.

test/Test/Database/LSMTree/Internal/Serialise/Class.hs
mheinzel: I used the predefined properties and just added a few calls to `counterexample`.

The only property where I didn't use a simple wrapper is `serialiseValueConcatDistributes`. I don't test it on arbitrary `RawBytes`, but just on valid serialisation output.
One could argue I should use `[RawBytes]` as input and either require deserialisation to be total or catch `error` calls, but it's simpler and IMO okay to use valid input only.

mheinzel:
I think I addressed all feedback.

The conclusion was that order preservation is not necessary for the compact index, but useful for the range query interface. I kept the property, but removed `instance SerialiseKey ByteArray`.

test/Test/Database/LSMTree/Internal/Serialise/Class.hs
```diff
@@ -0,0 +1,88 @@
...
+  where
+    bytes = RB.unpack (serialiseValue v)
+
+-- | Randomly splits the input list into non-empty chunks.
+genChunks :: [a] -> Gen [[a]]
+genChunks [] = pure []
+genChunks xs = do
+    n <- chooseInt (1, length xs)
+    let (pre, post) = splitAt n xs
+    (pre :) <$> genChunks post
```
jorisdral, line 81: I believe this would generate chunks that are skewed: i.e., larger chunks at the front, smaller chunks at the back (something like halving the length of the input list each time).

One could generate a number of uniformly random numbers, sort that, and use those numbers as indexes where to split the list (maybe that's a fairer splitting approach?). You could change this if you want to, but otherwise I'm fine with it as is.

jorisdral (approved):

test/Test/Database/LSMTree/Internal/Serialise/Class.hs
mheinzel: It certainly could be improved. One IMO nice aspect of the current implementation is that the expected number of chunks only grows logarithmically with the length of the list, but one could of course still have that without skew.

I'll leave it, it should work alright here.

PR #187, opened 2024-04-24
apt-get update workaround

phadej:
Let's `apt-get update` fail. We really care that the `apt-get install` succeeds.


jorisdral:
Do we still need this? The GHA error seems to have subsided

phadej:
I don't think this would harm either, should similar failure happen in the future

PR #188, opened 2024-04-24
Use language: GHC2021

phadej:
I'm quite annoyed by having to add `{-# LANGUAGE BangPatterns, ScopedTypeVariables #-}` etc at every second file.

We could also take this a bit further by having

```
common language
  default-language: GHC2021
  default-extensions: GADTs RoleAnnotations ViewPatterns
```

as first two will be in GHC2024, (but `ViewPatterns` never be as it's blocked until someone figures something "better", but we still use it already)

EDIT: I didn't remove now unneeded LANGUAGE pragmas everywhere to keep the change diff small. We may cleanup as we go.


phadej:
> sure we might have to provide compatibility 

Then we can go for 

```
common language
  default-language: GHC2021
  default-extensions: GADTs RoleAnnotations ViewPatterns
```

approach, just with longer `default-extensions` list.

I think we are using some features present in GHC-9.2+ already, that adding support for GHC-8.10 won't be cheap&easy anyway.

jorisdral (approved):

PR #189, opened 2024-04-25
Add NFData (Normal.Update v blob) instance

phadej:
This is needed for benchmarks (measuring the overhead of operation generation).


mheinzel (approved):
:+1:

I need the same (+ `Bifunctor`) for the micro benchmark.

PR #190, opened 2024-04-25
Lsm full benchmark

phadej:


bench/macro/lsm-tree-bench-wp8.hs
```diff
@@ -0,0 +1,404 @@
...
+        someFs = FS.SomeHasFS (FsIO.ioHasFS mountPoint)
+
+    name <- maybe (fail "invalid snapshot name") return $
+        LSM.mkSnapshotName "bench"
+
+    let initGen = SM.mkSMGen opts.seed
+
+    withSession someFs (FS.mkFsPath []) $ \session -> do
+        -- open snapshot
+        tbl <- LSM.open @IO @K @V @B session name
```
mheinzel, line 339: It's very useful to be able to generate the initial LSM once and then reuse it, nice.

The snapshot API is only supposed to be implemented in WP15, but it makes sense to target it here (given that we only want to use the external API) and we should be able to come up with a hacky snapshot implementation as part of WP7 that isn't the real deal, but makes this benchmark work. :+1:

bench/macro/lsm-tree-bench-wp8.hs
```diff
@@ -0,0 +1,404 @@
...
+        -- we generate n random numbers in range of [ 1 .. d ]
+        -- what is the chance they are all distinct
+        let n = fromIntegral (opts.batchCount * opts.batchSize) :: Double
+        let d = fromIntegral gopts.initialSize :: Double
+        -- this is birthday problem.
+        let p = 1 - exp (negate $  (n * (n - 1)) / (2 * d))
+
+        -- number of people with a shared birthday
+        -- https://en.wikipedia.org/wiki/Birthday_problem#Number_of_people_with_a_shared_birthday
+        let q = n * (1 - ((d - 1) / d) ** (n - 1))
```
mheinzel, line 227: Very nice that you included the calculations and also the actual check!

One detail: The check below counts the collisions, but this formula calculates the number of people with shared birthday, i.e. 2 if one collision exists, which is about a factor 2 more. So both make sense, but since we look at them next to each other, it would be good for them to agree.

bench/macro/lsm-tree-bench-wp8.hs
```diff
@@ -0,0 +1,404 @@
...
+   benchmark, or to take any other measure to exclude the cost of
+   generating or reading the sequence of operations.
+H. The benchmark should use the external interface of the disk backend
+   to present batches of operations: a first batch consisting of 256
+   lookups, followed by a second batch consisting of 256 inserts plus
+   256 deletes. This corresponds to the UTxO workload using 64kb
+   blocks, with 512 byte txs with 2 inputs and 2 outputs.
+I. The benchmark should be able to run in two modes, using the
+   external interface of the disk backend in two ways: serially (in
+   batches), or fully pipelined (in batches).
```
mheinzel, line 35: Since we leave the pipelined implementation for later, please mention it in a TODO.

mheinzel (commented):
This looks great to me, just with a question about key generation that I'm not sure about.

bench/macro/lsm-tree-bench-wp8.hs
```diff
@@ -0,0 +1,404 @@
...
+
+{- | Implement generation of unbounded sequence of insert/delete operations
+
+matching UTxO style from spec: interleaved batches insert and lookup
+configurable batch sizes
+1 insert, 1 delete, 1 lookup per key.
+
+Current approach is probabilistic, but uses very little state.
+We could also make it exact, but then we'll need to carry some state around
+(at least the difference).
```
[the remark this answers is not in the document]
phadej: > rejection should allow us to generate sequences without duplicates.

I'm not sure. I have an better idea, but I don't want stall this PR indefinitely discussing options.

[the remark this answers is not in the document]
mheinzel: Sounds good. To clarify: I don't think a non-probabilistic implementation is necessary to merge this PR, but it would be good to find out soon whether we will need it or not.

Then let's leave the implementation as is and I will make sure the the questions gets answered. If necessary, we can then do it as a follow-up.

bench/macro/lsm-tree-bench-wp8.hs
mheinzel: Once this one is addressed (which could be telling me why I'm wrong) the PR seems good to merge.

phadej: > number of people with a shared birthday

is an overestimation, if there are three people with same birthdays it's only two collisions etc.

I can look out for formula for collisions, but not right now. We might as well scrap this way of key generation too.

I.e. if you don't like this formula, then I'd rather remove whole calculation as the quick solution.

mheinzel: Ah no, don't remove it. It's nice to have.

Okay, then there's a very quick solution: Just add a comment like "this overestimates the number of collisions by approximately factor of 2 (assuming it's extremely unlikely for a key to occur more than twice)". I think it's good to have that mentioned in case someone wonders about the the mismatch in the output.

I also think it would then be nicer to print `q / 2` instead of `q`, but you can leave it as is.

phadej: done

mheinzel (approved):

bench/macro/lsm-tree-bench-wp8.hs
```diff
@@ -0,0 +1,407 @@
...
+doDryRun' :: GlobalOpts -> RunOpts -> IO ()
+doDryRun' gopts opts = do
+    -- calculated some expected statistics for generated batches
+    id $ do
+        -- we generate n random numbers in range of [ 1 .. d ]
+        -- what is the chance they are all distinct
+        let n = fromIntegral (opts.batchCount * opts.batchSize) :: Double
+        let d = fromIntegral gopts.initialSize :: Double
+        -- this is birthday problem.
+        let p = 1 - exp (negate $  (n * (n - 1)) / (2 * d))
```
jorisdral, line 226: Does this account for the keys that are inserted/deleted during the run?

jorisdral (approved):
Neat that this benchmark is stateless. I think the PR can be merged as is, but I think it would be useful to investigate these two avenues in follow-ups:
* Mixed inserts/deletes during setup
* Exact, non-probablistic generation

bench/macro/lsm-tree-bench-wp8.hs
mheinzel: You mean in we don't just have `d = initialSize` "days", but the number of possible keys increases over the course of a run? It doesn't account for that (so it overestimates the expected collisions), but as long as we have relatively few batches, it doesn't make much of a difference.

dcoutts (approved):
LGTM!

bench/macro/lsm-tree-bench-wp8.hs
jorisdral: Yup, that's what I was referring to -- I also came to the conclusion you mentioned, but I wanted someone else to verify

PR #191, opened 2024-04-26
crc32c hotfix

phadej:


jorisdral (approved):
I've marked the Windows jobs as non-required for now. It's something with git submodules that's broken

PR #193, opened 2024-04-26
Orphan NFData instances to core lib

jorisdral:


src/Database/LSMTree/Internal/Run.hs
```diff
@@ -105,6 +108,8 @@ data Run fhandle = Run {
       -- I\/O, reading arbitrary file offset and length spans.
     , runBlobFile   :: !fhandle
     }
+    deriving stock Generic
+    deriving anyclass NFData
```
mheinzel, line 112: Maybe we shouldn't derive `Generic` for the core library types. I don't feel strongly about it, but other types have manually written instances even if slightly inconvenient (e.g. see `NFData IndexCompact`) and I can see a rationale behind it.

(same for `BlobRef` and `BlobSpan`)

mheinzel (commented):
Looks good, except for maybe the general question of using `Generic` or not.

src/Database/LSMTree/Internal/Run.hs
mheinzel: I generally agree with your reasoning, with the caveat that we now have to be careful about which types get exposed and which don't (which is a non-local property and can also change), and treat them differently. Forgetting to update an instance also is a risk, you're right, but it might be easier to avoid since the instance is directly next to the type. Also, `Generic` has a compile time cost and is quite a big hammer for just this one boilerplate instance per type.

But ultimately I don't see a strong reason for one or the other, both have their own small inconveniences and risks. Feel free to go ahead with your plan.

jorisdral: After giving it some thought, I've removed the `Generic` instances and added `NFData` instances by hand

mheinzel (approved):
Looks good, and the instances will now even stop compiling when forgetting to add a new field. :+1:

PR #196, opened 2024-04-29
WIP: MCG for lsm-tree-bench-wp8.hs

phadej:
MCG is as good as LFSR (mentioned by @mheinzel); the prime field is just different (prime vs. 2^n). Arbitrary MCG is easier to construct. Also GHC doesn't have [`CLMUL`](https://gitlab.haskell.org/ghc/ghc/-/issues/24075) which makes LFSR inconvenient to compute (you need to shift and xor), so to support a range of bitsizes (`n` in `2^n`) one would need to write a lot of repetitive code.

I don't think it's worth it, as the key generation will anyway be counted out of the benchmark. The MCG is not that much slower than LFSR than we would notice. (It's not key generation which should dominate the benchmark).


mheinzel (commented):
Neat! It's not clear yet whether we strictly need it, but now that it exists I don't see a reason against merging it regardless. Some extra code, but I think it's nice and relatively simple.

It's probably useful to keep the code that counts the collisions (to test that the key generation works as intended), but calculating the estimated collisions doesn't really make sense any more, so IMO should be removed.

mheinzel (approved):
I think we should merge this.

Do you want to remove the collision estimate calculations before we do that?

phadej:
> Do you want to remove the collision estimate calculations before we do that?

Not really.

phadej:
EDIT: to clarify the latter. I'd do cleanups only when the benchmark is actually "ready" and producing results; until then it's just an experiment.

mheinzel (approved):
Thanks for the clarification. Sounds okay to me.

PR #197, opened 2024-04-30
Monomorphise the write buffer

jorisdral:
All the internals deal with serialised representations

### DIFF
--- a/src-extras/Database/LSMTree/Extras/Generators.hs
+++ b/src-extras/Database/LSMTree/Extras/Generators.hs
@@ -15,8 +15,9 @@
 
 module Database.LSMTree.Extras.Generators (
     -- * WriteBuffer
-    genWriteBuffer
-  , shrinkWriteBuffer
+    TypedWriteBuffer (..)
+  , genTypedWriteBuffer
+  , shrinkTypedWriteBuffer
     -- * WithSerialised
   , WithSerialised (..)
     -- * UTxO keys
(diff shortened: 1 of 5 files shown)


src-extras/Database/LSMTree/Extras/Generators.hs
```diff
@@ -196,8 +219,8 @@ shrinkWriteBuffer ::
   => (k -> [k])
   -> (v -> [v])
   -> (blob -> [blob])
-  -> WriteBuffer k v blob
-  -> [WriteBuffer k v blob]
+  -> WriteBuffer
+  -> [WriteBuffer]
```
mheinzel, line 223: We could also monomorphise `gen/shrinkWriteBuffer`, so it would take a `SerialisedKey -> [SerialisedKey]` argument etc., remove `fromKOps` and `toKOps` and then in the instances do `shrinkWriteBuffer arbitrary _ _` and `shrinkWriteBuffer (fmap serialiseKey . shrink @k . deserialiseKey) _ _` respectively.

Does the same thing in the end, but fits the untyped/typed split and might be cleaner. Do think it would be worth it?

mheinzel (approved):
Looks good, the approach of `TypedWriteBuffer` for the tests is nice. Just one idea, up to you.

I think we could get this in quickly and then I'll rebase #184 on top of it?

src-extras/Database/LSMTree/Extras/Generators.hs
jorisdral: I've instead changed `genWriteBuffer`/`shrinkWriteBuffer` to `genTypedWriteBuffer`/`shrinkTypedWriteBuffer`. `Arbitrary WriteBuffer` is now a special case of `Arbitrary (TypedWriteBuffer k v blob)`. Basically I've flipped the relationship around

PR #198, opened 2024-04-30
Macro-benchmark for lookup functions

jorisdral:


bench/macro/Bench/Database/LSMTree/Internal/Lookup.hs
jorisdral, line 1: This commit is only here so I could see from the diff of later commits what I was changing with respect to the bloomfilter macro-benchmark

src-extras/Database/LSMTree/Extras/Orphans.hs
jorisdral, line 1: I don't remember why I added this -- I took a little break from this branch and I don't seem to be using the instances anywhere. So, I might remove them

jorisdral:
Current benchmark results are included below. We should probably first review this PR and make fixes before we decide these numbers will be accurate.

<details>
<summary>macro-bench-preplookups.log</summary>

```
Precomputed run stats:
(numEntries, sizeFactor)
(65536,1)
(65536,1)
(65536,1)
(65536,1)
(65536,1)
(65536,1)
(65536,1)
(65536,1)
(262144,4)
(262144,4)
(262144,4)
(262144,4)
(262144,4)
(262144,4)
(262144,4)
(262144,4)
(1048576,16)
(1048576,16)
(1048576,16)
(1048576,16)
(1048576,16)
(1048576,16)
(1048576,16)
(1048576,16)
(16777216,256)
total number of entries:	 27787264
total number of key lookups:	 25000000
Generating runs...
........................................................................................................................................................................................................................................................................................................................................................................................................................................ finished.
Actual run stats:
(numEntries)
NumEntries {unNumEntries = 65536}
NumEntries {unNumEntries = 65536}
NumEntries {unNumEntries = 65536}
NumEntries {unNumEntries = 65536}
NumEntries {unNumEntries = 65536}
NumEntries {unNumEntries = 65536}
NumEntries {unNumEntries = 65536}
NumEntries {unNumEntries = 65536}
NumEntries {unNumEntries = 262144}
NumEntries {unNumEntries = 262144}
NumEntries {unNumEntries = 262144}
NumEntries {unNumEntries = 262144}
NumEntries {unNumEntries = 262144}
NumEntries {unNumEntries = 262144}
NumEntries {unNumEntries = 262144}
NumEntries {unNumEntries = 262144}
NumEntries {unNumEntries = 1048576}
NumEntries {unNumEntries = 1048576}
NumEntries {unNumEntries = 1048576}
NumEntries {unNumEntries = 1048576}
NumEntries {unNumEntries = 1048576}
NumEntries {unNumEntries = 1048576}
NumEntries {unNumEntries = 1048576}
NumEntries {unNumEntries = 1048576}
NumEntries {unNumEntries = 16777216}

Benchmarking benchBloomQueriesBaseline ... 
TODO
Finished.
Time total:        0.22 seconds
Alloc total:       2.40e9 bytes
Time net:          0.22 seconds
Alloc net:         2.40e9 bytes
Time net per key:  8.78e-9 seconds
Alloc net per key: 96.19 bytes
WARNING: expecting 0, got 96 bytes allocated per key

Benchmarking benchBloomQueries ... 
TODO
Finished.
Time total:        31.82 seconds
Alloc total:       5.02e9 bytes
Time net:          31.60 seconds
Alloc net:         2.62e9 bytes
Time net per key:  1.26e-6 seconds
Alloc net per key: 104.75 bytes
WARNING: expecting 0, got 104 bytes allocated per key

Benchmarking benchIndexSearchesBaseline ... 
TODO
Finished.
Time total:        31.72 seconds
Alloc total:       5.02e9 bytes
Time net:          31.72 seconds
Alloc net:         5.02e9 bytes
Time net per key:  1.27e-6 seconds
Alloc net per key: 200.94 bytes
WARNING: expecting 0, got 200 bytes allocated per key

Benchmarking benchIndexSearches ... 
TODO
Finished.
Time total:        47.64 seconds
Alloc total:       3.14e11 bytes
Time net:          15.92 seconds
Alloc net:         3.09e11 bytes
Time net per key:  6.37e-7 seconds
Alloc net per key: 12347.61 bytes
WARNING: expecting 0, got 12347 bytes allocated per key

Benchmarking benchPrepLookupsBaseline ... 
TODO
Finished.
Time total:        0.22 seconds
Alloc total:       2.40e9 bytes
Time net:          0.22 seconds
Alloc net:         2.40e9 bytes
Time net per key:  8.71e-9 seconds
Alloc net per key: 96.19 bytes
WARNING: expecting 0, got 96 bytes allocated per key

Benchmarking benchPrepLookups ... 
TODO
Finished.
Time total:        46.47 seconds
Alloc total:       3.14e11 bytes
Time net:          46.25 seconds
Alloc net:         3.11e11 bytes
Time net per key:  1.85e-6 seconds
Alloc net per key: 12452.37 bytes
WARNING: expecting 0, got 12452 bytes allocated per key
```

</details>

src-extras/Database/LSMTree/Extras/Orphans.hs
jorisdral: Never mind, see #206. I actually need these instances!

bench/macro/Bench/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -0,0 +1,494 @@
...
+    unless (totalNumEntries runSizes >= benchmarkNumLookups) $
+      fail "number of key lookups is more than number of entries"
+
+    traceMarkerIO "Generating runs"
+    putStr "<Generating runs>"
+    -- this key is reused for both setting up the runs and generating lookup
+    -- batches. This ensures that we only look up keys that we have put in a
+    -- run. Also, prepLookupsEnv makes to shuffle the generated keys into the
+    -- different runs, such that the generated lookups access the runs in random
+    -- places.
```
jorisdral:
```suggestion
    -- this key is reused for both setting up the runs and generating lookup
    -- batches. This ensures that we only perform true positive lookups.
    -- Also, prepLookupsEnv makes to shuffle the generated keys into the
    -- different runs, such that the generated lookups access the runs in random
    -- places.
```

bench/macro/Bench/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -0,0 +1,494 @@
...
+-- true positives in only one run/filter. Thus most lookups will be true
+-- negatives.
+--
+-- The goal is to benchmark the benefits of optimisations for the LSM situation:
+--
+-- * where the same key is being looked up in many filters,
+-- * with a hit in only one filter expected, and
+-- * where the total size of the filters is too large to fully fit in cache
+--   (though the smaller ones may fit in the caches).
+--
```
jorisdral, line 259:
```suggestion
-- | Input environment for benchmarking lookup functions.
--
-- The idea here is to have a collection of runs corresponding to the sizes used
-- in a largeish LSM. In particular, the sizes are in increasing powers of 4.
-- The keys in the runs are non-overlapping, and lookups will be
-- true positives in only one run. Thus most lookups will be true
-- negatives.
--
-- The goal is to benchmark the critical path of performing asynchronous lookups serially.
--
-- * where the same key is being looked up in many runs,
-- * with a true positive lookup in only one run,
-- * with true negative lookups in the other runs
-- * with false positives lookups in a fraction of the runs according to the bloom filters' false positive rates
```

bench/macro/Bench/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -0,0 +1,494 @@
...
+
+-- | @'vectorOfUniforms' vec g0@
+--
+-- TODO: implement using generateM
+vectorOfUniforms ::
+     (PrimMonad m, VGM.MVector v a, RandomGen g, Uniform a)
+  => v (PrimState m) a
+  -> g
+  -> m g
+vectorOfUniforms !vec !g0 = do
```
jorisdral, line 460: This doesn't need a `v` argument. It can return a new `v` instead

bench/macro/Bench/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -0,0 +1,494 @@
...
+    let zero = serialiseValue zeroUTxOValue
+    foldM_
+      (\ !i (!rb, !n) -> do
+        let !mvecLocal = VUM.unsafeSlice i n mvec
+        Merge.sort mvecLocal
+        forM_ [0..n-1] $ \j -> do
+          -- progress
+          when (j .&. 0xFFFF == 0) (putStr ".")
+          !k <- mvecLocal `VUM.unsafeRead` j
+          void $ RunBuilder.addKeyOp hfs rb (serialiseKey k) (Insert zero)
```
jorisdral, line 301: Could be reworked using `V.imapM_`

bench/macro/Bench/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -0,0 +1,494 @@
...
+      -> ST s (V.Vector SerialisedKey, StdGen)
+    go !rng !i !mres
+      | n0 == i   = do
+          !res <- V.unsafeFreeze mres
+          pure (res, rng)
+      | otherwise = do
+          let (!k, !rng') = uniform @StdGen @UTxOKey rng
+              !sk = serialiseKey k
+          VM.write mres i $! sk
+          go rng' (i+1) mres
```
jorisdral, line 333: Move to utilities, possibly also rework using `V.generateM`

src/Database/LSMTree/Internal/RunAcc.hs
```diff
@@ -93,7 +93,7 @@ new :: NumEntries
                                   -- which is based on the 'NumPages'.
     -> ST s (RunAcc s)
 new (NumEntries nentries) npages rangeFinderPrecisionOverride = do
-    mbloom <- Bloom.Easy.easyNew 0.1 nentries -- TODO(optimise): tune bloom filter
+    mbloom <- Bloom.Easy.easyNew 0.02 nentries -- TODO(optimise): tune bloom filter
```
jorisdral, line 96: This is in line with what the bloom filter macro-benchmark was using. It leads to a little over 9 bits per entry in the filter

jorisdral (commented):
The PR should be ready for review. I've actually slightly expanded the scope of this PR to also macro-benchmark the `lookupsInBatches` function -- the original idea was to benchmark only the in-memory functions. As `lookupsInBatches` will do actual IO, the benchmark has a `getLine` pause built in that allows the user to drop the OS page caches before continuing. This should lead to relatively realistic IO costs.

One thing I'm planning to do later -- possibly in a follow-up PR -- is to also add factor the two main components of `lookupsInBaches` that aren't benchmarked separately yet:
* Submitting and waiting for the I/O 
* Intra-page lookups, and resolving values in case we have multiple lookup results for the same key
This factoring out can probably also be done in the micro-benchmarks, so I'll include that in the follow-up as well

mheinzel (commented):
The generation of inputs and calculation of relevant measurements looks reasonable to me. It might worth discussing the LSM shape we want to use.

bench/macro/Bench/Database/LSMTree/Internal/Lookup.hs
jorisdral: I recall now that I wrote it this way because I needed `StdGen` as part of the result of this function. `V.unfoldrExactN` doesn't return that

bench/macro/Bench/Database/LSMTree/Internal/Lookup.hs
jorisdral: Kept it the way it is

bench/macro/Bench/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -0,0 +1,494 @@
...
+    _bindexSearches <-
+      benchmark "benchIndexSearches"
+                "Calculate batches of keys, perform bloom queries for each batch, and perform index searches for positively queried keys in each batch. Net time/allocation is the result of subtracting the cost of benchBloomQueries."
+                (benchIndexSearches blooms indexes handles keyRng0) benchmarkNumLookups
+                bbloomQueries
+    _bprepLookups <-
+      benchmark "benchPrepLookups"
+                "Calculate batches of keys, and prepare lookups for each batch. This is roughly doing the same amount of work as benchIndexSearches. Net time/allocation is the result of subtracting the cost of benchGenKeyBatches."
+                (benchPrepLookups blooms indexes handles keyRng0) benchmarkNumLookups
+                bgenKeyBatches
```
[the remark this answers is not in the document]
jorisdral: `benchPrepLookups` is supposed to measure all the in-memory operations (bloom queries and index searches) but not the lookup batch generation. `benchIndexSearches` is supposed to only measure the index search, but I need to subtract the bloom query time from that.

So yeah, they are actually doing the same thing (`prepLookups` is just a thin wrapper around bloom queries and index searches). Running almost the same thing twice is maybe overkill but it's pretty fast so I don't mind the duplication here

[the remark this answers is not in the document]
jorisdral: This actually made me realise that in the top bit I'm only subtracting the net cost of doing the bloom queries, and not generating batches of lookups, so I changed that

bench/macro/Bench/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -0,0 +1,494 @@
...
+import qualified System.FS.BlockIO.IO as FS
+import qualified System.FS.IO as FS
+import           System.IO.Temp
+import           System.IO.Unsafe (unsafePerformIO)
+import           System.Mem (performMajorGC)
+import           System.Random
+
+-- | The number of entries in the smallest LSM runs is @2^benchmarkSizeBase@
+benchmarkSizeBase :: SizeBase
+benchmarkSizeBase = 18 -- should give us ~111 million entries in the database
```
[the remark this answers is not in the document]
jorisdral: Good point. I'm going to fiddle around a bit with this

bench/macro/Bench/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -0,0 +1,494 @@
...
+--
+-- Due to the incremental merging, each level actually has (in the worst case)
+-- 2x the number of runs, hence 8 per level for tiering levels.
+--
+lsmStyleRuns :: SizeBase -> [RunSizeInfo]
+lsmStyleRuns l1 =
+       replicate 8 (2^(l1+0), 1)   -- 8 runs at level 1 (tiering)
+    ++ replicate 8 (2^(l1+2), 4)   -- 8 runs at level 2 (tiering)
+    ++ replicate 8 (2^(l1+4),16)   -- 8 runs at level 3 (tiering)
+    ++            [(2^(l1+8),256)] -- 1 run  at level 4 (leveling)
```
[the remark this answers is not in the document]
jorisdral: I've mainly copied this from the bloomfilter macro-benchmark. I'll put some more thought into it

Regarding how many runs per level are used for lookup at most: that's a good point. I'm also going to think about this a little bit more

bench/macro/Bench/Database/LSMTree/Internal/Lookup.hs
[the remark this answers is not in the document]
jorisdral: Turns out `14` is a good number. `2^14` is roughly how many UTXO-like entries fit into a 2MB buffer

bench/macro/Bench/Database/LSMTree/Internal/Lookup.hs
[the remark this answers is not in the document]
jorisdral: I've decided to keep it as is for now, we can change it later if we want to -- I think that you're right that the levels setup could be more realistic, but for lookups having one fewer run per level is actually better for lookup performance

mheinzel (approved):

PR #201, opened 2024-05-03
`fs-sim` with simulated file I\/O using user-supplied buffers

jorisdral:


test/Test/Database/LSMTree/Internal/Run.hs
```diff
@@ -83,9 +83,6 @@ tests = testGroup "Database.LSMTree.Internal.Run"
   where
     withSessionDir = Temp.withSystemTempDirectory "session-run"
 
-    -- Currently doesn't support HasBufFS, so we still need the other version.
-    -- TODO: add support once simulation is merged:
-    -- https://github.com/input-output-hk/fs-sim/pull/48
     -- TODO: Also test file system errors.
```
jorisdral, line 86: @mheinzel This is now possible!

mheinzel: Nice, I'll do it once this PR is merged.

mheinzel (approved):
Nice, great to have!

PR #202, opened 2024-05-03
Check preconditions in compact index construction

jorisdral:


src/Database/LSMTree/Internal/IndexCompactAcc.hs
```diff
@@ -152,6 +152,15 @@ append (AppendMultiPage k n) ica        = appendMulti (k, n) ica
 -- INVARIANTS: see [construction invariants](#construction-invariants).
 appendSingle :: forall s. (SerialisedKey, SerialisedKey) -> IndexCompactAcc s -> ST s (Maybe Chunk)
 appendSingle (minKey, maxKey) ica@IndexCompactAcc{..} = do
+#ifdef NO_IGNORE_ASSERTS
+    lastMinKey <- readSTRef icaLastMinKey
```
mheinzel, line 156: I'm not sure adding CPP is worth it just to save a single `readSTRef`. I'm leaning towards "no", but I'm okay with it if you think it is.

src/Database/LSMTree/Internal/IndexCompactAcc.hs
```diff
@@ -152,6 +152,15 @@ append (AppendMultiPage k n) ica        = appendMulti (k, n) ica
...
+    let preconditionSorted :: Bool
+        preconditionSorted = minKey <= maxKey && smaybe True (<= minKey) lastMinKey
+    assert preconditionSorted $ pure ()
+    let preconditionPartitioned :: Bool
+        preconditionPartitioned = minRfbits == keyTopBits16 icaRangeFinderPrecision maxKey
+    assert preconditionPartitioned $ pure ()
```
mheinzel, line 162: I don't think the bindings and type signatures really make it clearer, just feels more noisy. :shrug: 

```suggestion
    assert (minKey <= maxKey && smaybe True (<= minKey) lastMinKey) $ pure ()    -- sorted
    assert (minRfbits == keyTopBits16 icaRangeFinderPrecision maxKey) $ pure ()  -- partitioned
```

src/Database/LSMTree/Internal/IndexCompactAcc.hs
jorisdral: What feels wrong about using CPP here in your opinion?

Conceptually, I think it makes sense. Assertion-related code should ideally be erased when compiling with assertions disabled, though it's slightly unfortunate that the value we want to write an assertion on is in an `STRef` and so we use CPP

mheinzel: Mainly a general dislike of CPP I think. I don't like the potential of introducing code that works with assertions disabled, but doesn't compile when they're enabled. When just using `assert`, the check will be erased, but still be type-checked.

It's not too bad here, though, and we usually compile with assertions enabled (plus we have it in CI). So I don't mind too much.

mheinzel (approved):

src/Database/LSMTree/Internal/IndexCompactAcc.hs
jorisdral: Having the `readSTRef` not be erased could lead to problems of its own, like `unused variable` warnings for the result of the read

mheinzel: I think the assert will only be erased after the unused variable check, so that shouldn't be an issue.

jorisdral: Ah yes, you're quite right

PR #210, opened 2024-05-13
Representation of sessions, table handles, and configuration options

jorisdral:


src/Database/LSMTree/Internal/TableHandle.hs
```diff
@@ -0,0 +1,183 @@
...
+  Table handle
+-------------------------------------------------------------------------------}
+
+-- | A handle to an on-disk key\/value table.
+--
+-- For more information, see 'Database.LSMTree.Normal.TableHandle'.
+data TableHandle m h = TableHandle {
+    tableSession     :: !(Session m h)
+  , tableConfig      :: !TableConfig
+  , tableWriteBuffer :: !(WriteBuffer)
```
mheinzel, line 73:
```suggestion
  , tableWriteBuffer :: !WriteBuffer
```

src/Database/LSMTree/Internal/TableHandle.hs
```diff
@@ -0,0 +1,183 @@
...
+    -- normal tables.
+  , confResolveMupsert   :: !(Maybe ResolveMupsert)
+  }
+  deriving Show
+
+data MergePolicy =
+    MergePolicyTiering
+{- TODO: disabled for now
+  | MergePolicyLevelling
+-}
```
mheinzel, line 149: We are not using pure tiering. I think the name for having levelling on the last level was "lazy levelling", so should we use that here?

src/Database/LSMTree/Internal/TableHandle.hs
```diff
@@ -0,0 +1,183 @@
...
+    -- false-positive rate. Do this for each bloom filter.
+    AllocRequestFPR
+      Double -- ^ Requested FPR.
+{- TODO: disabled for now
+  | -- | Allocate bits amongst all bloom filters according to the Monkey algorithm.
+    --
+    -- TODO: add more configuration paramters that the Monkey algorithm might
+    -- need once we implement it.
+    AllocMonkey
+      Word32 -- ^ Total number of bytes that bloom filters can use collectively.
```
mheinzel, line 175: This is commented out for now, but already thinking ahead a bit: without knowing how many entries there will be, it's difficult to target a total filter size. Initially you can spend a lot, but at some point you are approaching the limit and need to make sure new (merged) runs will use fewer bits per entry. And any runs you create will stay around for a while, so you have to start decreasing the bits early enough. Also might need some care when cloning tables.

src/Database/LSMTree/Internal/TableHandle.hs
```diff
@@ -0,0 +1,183 @@
...
+    tableSession     :: !(Session m h)
+  , tableConfig      :: !TableConfig
+  , tableWriteBuffer :: !(WriteBuffer)
+    -- | A hierarchy of levels. The vector indexes double as level numbers.
+  , tableLevels      :: !(V.Vector (Level h))
+    -- | Cache of flattened 'levels'.
+    --
+    -- INVARIANT: when 'level's is modified, this cache should be updated as
+    -- well, for example using 'mkLevelsCache'.
+  , tableLevelsCache :: !(LevelsCache h)
```
mheinzel, line 80: Once we have updates, these fields will need to allow for mutation. Also, we will need a synchronisation mechanism. We can leave that for later, but at least a comment might be a good idea.

bench/macro/lsm-tree-bench-wp8.hs
```diff
@@ -382,14 +388,21 @@ main = do
...
+withSession hfs hbio path = bracket (LSM.openSession hfs hbio path) LSM.closeSession
 
 defaultTableConfig :: LSM.TableConfig
 defaultTableConfig =  LSM.TableConfig
-    { LSM.tcMaxBufferMemory      = 1000
-    , LSM.tcMaxBloomFilterMemory = 1000
-    , LSM.tcBitPrecision         = ()
+    { LSM.confMergePolicy      = LSM.MergePolicyTiering
+    , LSM.confSizeRatio        = LSM.Four
+    , LSM.confWriteBufferAlloc = 1000
```
mheinzel, line 403: Picking a default is pretty arbitrary, but that's not even a page. Maybe at least a couple dozen kB?

src/Database/LSMTree/Internal/TableHandle.hs
```diff
@@ -0,0 +1,183 @@
...
+-- * Size ratio: 4
+data TableConfig = TableConfig {
+    confMergePolicy      :: !MergePolicy
+    -- Size ratio between the capacities of adjacent levels.
+  , confSizeRatio        :: !SizeRatio
+    -- | Total number of bytes that the write buffer can use.
+    --
+    -- The maximum is 4GiB, which should be more than enough for realistic
+    -- applications.
+  , confWriteBufferAlloc :: !Word32
```
mheinzel, line 137: Do we already have a clear idea of what the buffer flush policy is going to be more precisely?
The buffer gets flushed once it reaches the specified size in memory (i.e. serialised form of keys/values, including the `Entry` and `Data.Map` overhead)? Does it strictly need to be flushed before getting any larger or can a batch of updates that crosses the size still be completely inserted before flushing?

mheinzel (commented):
Looks good, but I have a few questions for clarification.

src/Database/LSMTree/Internal/TableHandle.hs
jorisdral: I've added a TODO about a synchronisation mechanism

I think that for these fields, mutation isn't worth it. The levels/cache will rarely change (only when merges are finished) and at that point it's relatively cheap to allocate a new vector where necessary. And then we get sharing for free

jorisdral: Or maybe I'm forgetting about a case where we would need these vectors to be mutable?

bench/macro/lsm-tree-bench-wp8.hs
jorisdral: I'll go for 2 MB, which would fill at least 512 pages

src/Database/LSMTree/Internal/TableHandle.hs
jorisdral: Good points. I think we can divide this problem in two: 
* **How much memory should I allocate in total?**: as much as you can spare, or however much you want to use. It's probably more subtle than using as much as possible; throwing more memory at bloom filters will not necessarily lead to better performance. Maybe there is a good balance here, but I think the next problem is more important
* **How to best divide this memory across runs?**: that's harder to answer, but it's something the `lsm-tree` package will take care of. So, we should put thought into that

Regarding cloning: I think we should keep it simple, give each clone a separately managed budget, but don't try anything fancy with sharing a common budget amongst clones. If a table is created with a 2MB budget, then the clone will have its own 2MB budget, and we might end up using 4MB in the end (but only if the clones are long-lived).

Also, wall of text below. I started writing about monkey bit allocation and ended up just putting a lot of my thoughts down. Let me know what you think

### Thoughts on monkey bit allocation

In the UTxO case, we know precisely how many *logical* entries we have in the database. We could maybe precompute the monkey bits in that case, though it would be hard because we won't precisely know how many *physical* entries we have in the database. I also don't think a purely static approach would work very well because we'd be using sub-optimal bit allocation until the database reaches our estimated logical size. I suppose that even if we estimate the logical size of the database, it's going to be relatively hard to estimate what the levels are going to look like. To me that sounds like we should allocate dynamically. 

Something to note: we can't change the number of bits a bloom filter uses once it has been created. This means that when merging, we have to decide up front how many bits to use for a filter.

Some ideas are listed below. For now, assume entries are of fixed size and runs are always full.

First consider the case where merges aren't scheduled but are fully computed immediately. When a last-level merge occurs, then a run is produced that should be moved to the next level (increasing the maximum level by 1, which means the optimal allocation changes). At this point, before starting the merge, we recompute the bit allocation per bloom filter per level. Also, yes, runs will stay around for a while, but a last-level merge like that will only happen when the previous levels have been emptied by merges. When thinking about it in a 4-ary number system, which corresponds to an LSM structure with Tiering 4: `3333 + 1 = 10000`.

Now the case where merges are scheduled and computed incrementally. Here  I think we should recompute the bit allocation per bloom filter per level whenever a last-level merge *starts*. From that point on, any merge that is started in a previous level will also use the updated allocation numbers. Using the 4-ary number example: `3777 + 1 = 44444`. Ongoing merges on the first three levels have finished, and each adds a new run to the next level, starting new merges of their own. Most importantly, a last-level merge is started.

What if runs are not necessarily full and/or entries have varying sizes? 
* A last-level merge might not produce a run that is moved to the next level. I'm not sure what to do then: either just keep using the updated numbers, or revert to the old numbers for merges that are created after we've seen the result of the last-level merge.
* Merges on any level may finish early, and merges on intermediate levels might produces runs that aren't moved to the next level. I'm not sure what the worst-case behaviour is here: is it a bad thing if some merges are started early (i.e., before a last-level merge) and use old allocation numbers?
* The "Optimal Bloom Filters and Adaptive Merging for LSM-Trees" papers provides an algorithm in appendix `B` that can redistribute bits based on how many entries are in each run when entries have varying sizes (or runs are underfull AFAICS). It is, however, not at all clear to me what to do after the new allocation numbers are computed. We can't change the number of bits that a filter uses once the filter is created, so would they then recreate all the bloom filters in one go? That doesn't seem doable for us.

src/Database/LSMTree/Internal/TableHandle.hs
jorisdral: My first idea would be to base it purely on the serialised size of keys and values, but that's not set in stone. We can try out different approaches when start implementing this

src/Database/LSMTree/Internal/TableHandle.hs
mheinzel: Sounds sensible, although we might even be able to more gradually adjust the bit allocation, not just when adding another level. Crucially, we know how many physical entries there are in total and we know how many runs at least will have been replaced once a given number of extra entries have been added.

Definitely very interesting, I would love to sketch out some ideas and discuss it soon!

Something I realised also becomes relevant here: In-progress merges can temporarily inflate the number of physical entries by quite a lot. Most data can be on the last level, in one big merge, so in the worst case that could almost double the number of entries. We can calculate an upper bound to guarantee that can't exceed our budget, but reduces our budget quite a bit. So it would really help (and this has been brought up a few times before) to have a mechanism for splitting runs into sections, each with their own filter, index etc. Traditionally, these are SSTables, but there are other ways of implementing them. Then lookups for low keys can already look at the merged prefix and we can drop the now unused prefix of input runs as we go. This gives us a constant overhead per level.

src/Database/LSMTree/Internal/TableHandle.hs
mheinzel: It depends a bit on whether the config values should be seen as strict upper bounds, but in most cases that shouldn't be necessary.

src/Database/LSMTree/Internal/TableHandle.hs
jorisdral: Yes, let's discuss it together :)

src/Database/LSMTree/Internal/TableHandle.hs
jorisdral: IIRC we decided that table handles would not have a locking mechanism to make using them thread-safe, so I've made the mutable fields into MVars, and added a note about thread safety

src/Database/LSMTree/Internal/TableHandle.hs
```diff
@@ -0,0 +1,191 @@
...
+
+{-------------------------------------------------------------------------------
+  Table handle
+-------------------------------------------------------------------------------}
+
+-- | A handle to an on-disk key\/value table.
+--
+-- For more information, see 'Database.LSMTree.Normal.TableHandle'.
+--
+-- NOTE: a table handle is __not__ thread-safe.
```
mheinzel, line 71: We might want to be clearer about the guarantees we want to have in the future. If this was not considered thread-safe at all, we could just go with `IORef`s, but it seems like we do in fact ensure some synchronisation.

src/Database/LSMTree/Internal/TableHandle.hs
```diff
@@ -0,0 +1,191 @@
...
+    tableSession     :: !(Session m h)
+  , tableConfig      :: !TableConfig
+  , tableWriteBuffer :: !(StrictMVar m (WriteBuffer))
+    -- | A hierarchy of levels. The vector indexes double as level numbers.
+  , tableLevels      :: !(StrictMVar m (V.Vector (Level h)))
+    -- | Cache of flattened 'levels'.
+    --
+    -- INVARIANT: when 'level's is modified, this cache should be updated as
+    -- well, for example using 'mkLevelsCache'.
+  , tableLevelsCache :: !(StrictMVar m (LevelsCache h))
```
mheinzel, line 82: We might want to iterate on the exact synchronisation mechanism (e.g. separate `MVar`s for the levels and levels cache seems like a risk), but I think it's okay for now and should not hold up this PR.

mheinzel (approved):

src/Database/LSMTree/Internal/TableHandle.hs
jorisdral: Let's do that

src/Database/LSMTree/Internal/TableHandle.hs
jorisdral: I'm using `MVar`s here because we don't have `IORef`s (yet) in `io-sim`.

Let's iterate on the thread safety guarantees in later PRs

src/Database/LSMTree/Internal/TableHandle.hs
phadej: >  although we might even be able to more gradually adjust the bit allocation, 

IMO that is stupid idea. State of the art does 10 bits per entry and performs quite well. That has simplicity benefits: when you create run, you know "locally" what kind of bloom filter to create, it depends only on the entries count in the run.

With monkey setup you cannot "locally" create runs anymore, you need to know a bit more of context. You need to know the level, so you can lookup the FPR (or equivalently bits/entry ratio) from the configuration. That's not much additional complexity, though.

However, if you want to optimise even further, i.e. also consider the sizes of other runs (and their existing filters, assuming you probably dont' want to recreate those), it becomes a lot more complex. I'm not convinced you gain much.

---

As far as I understand now, the state of the art 10 bits per entry is a very simple strategy. It has benefits that you don't really need to budget the total memory available for filters in advance. The memory used will be total elements * 10 bits; so it will grow with database.

In practice, you'd probably still budget somewhat, i.e .estimate the maximum size of your database and get a machine with enough memory. The same, I guess, applies to monkey bit allocation. You do budgeting and run with it. It doesn't make sense to use much more memory than needed, because of diminished benefits. The monkey already uses more memory in first levels, so if your database is smaller than estimated, it will perform even better than when full.

And in use cases like UTxO you don't need commit to the total size for years, as that database is "cache", and can be always rebuilt.

---

Insert/Delete problem: That is an interesting point. The monkey paper doesn't discuss what to do when there are a lot of successfully merged updates/deletes with inserts, i.e. deeper runs are not "full". *I guess* that you just use the bits/entries ratio you had, so the total costs stay the same, you just end up using less memory than originally budgeted (so you may need to over-allocate?)

Issue #212, opened 2024-05-14
Make the macro-benchmark for lookup functions more granular


jorisdral:
See https://github.com/IntersectMBO/lsm-tree/pull/198#pullrequestreview-2045129549

This involves factoring out the cost of submitting batches of lookups to `HasBlockIO`, and the cost of intra-page lookups. This is a similar approach to how we benchmark `bloomQueries` and `indexSearches` separately from `prepLookups`

PR #214, opened 2024-05-15
More granular micro-benchmark for the IO part of lookups

jorisdral:
This tackles the same thing as #212, but only for the lookup micro-benchmarks.

### DIFF
--- a/bench/micro/Bench/Database/LSMTree/Internal/Lookup.hs
+++ b/bench/micro/Bench/Database/LSMTree/Internal/Lookup.hs
@@ -17,7 +17,7 @@ module Bench.Database.LSMTree.Internal.Lookup (benchmarks) where
 import           Control.Exception (assert)
 import           Control.Monad
 import           Criterion.Main (Benchmark, bench, bgroup, env, envWithCleanup,
-                     nfAppIO, whnf, whnfAppIO)
+                     perRunEnv, whnf, whnfAppIO)
 import           Data.Bifunctor (Bifunctor (..))
 import qualified Data.List as List
 import           Data.Map.Strict (Map)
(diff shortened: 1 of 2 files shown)


bench/micro/Bench/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -63,22 +76,47 @@ benchLookups conf@Config{name} =
...
+                        ) $ \ ~(rkixs, ioops, ioress) -> do
+                !_ <- intraPageLookups resolveV rs ks rkixs ioops ioress
+                pure ()
+            -- The whole shebang: lookup preparation, doing the IO in batches,
+            -- and then performing intra-page-lookups. Again, we evaluate the
+            -- result to WHNF because it is the same result that
+            -- intraPageLookups produces (see above).
           , bench "Batched lookups in IO" $
-              nfAppIO (\ks' -> lookupsInBatches hasBlockIO bsize resolveV rs blooms indexes kopsFiles ks') ks
+              whnfAppIO (\ks' -> lookupsInBatches hasBlockIO bsize resolveV rs blooms indexes kopsFiles ks') ks
```
mheinzel, line 119: Did this make a noticeable difference?

mheinzel (approved):

bench/micro/Bench/Database/LSMTree/Internal/Lookup.hs
jorisdral: The difference is not very noticeable, or at least small enough that I can't say for certain whether there is a significant difference. We're talking about a ~10 micro-seconds difference in the positive lookup scenario, and no real difference in the negative lookup scenario

jorisdral: Conceptually `whnfAppIO` is the right thing to do: the data should already be fully evaluated

PR #219, opened 2024-05-17
Conditional `blockio-uring` dependency

jorisdral:


cabal.project
```diff
@@ -32,12 +32,13 @@ constraints: bloomfilter <0
...
-    location: https://github.com/well-typed/blockio-uring
-    -- submitIO using vectors
-    tag: fdb1a51e94ad731b1bf114ee1f8143720481a49a
+-- comment me if you don't have liburing installed
+--
+-- TODO: it is slightly unfortunate that one has to manually remove this file
+-- import in case they don't have liburing installed... Currently, it is not
+-- possible to have conditionals on package flags in a project file. Otherwise,
+-- we could add a conditional on (+serialblockio) to remove this import automatically.
+import: cabal.project.blockio-uring
```
mheinzel, line 41: Indeed unfortunate :(

mheinzel (approved):
A bit unfortunate, but seems like the best we can do at the moment.

CONTRIBUTING.md
```diff
@@ -1,9 +1,53 @@
...
+component from the build targets passed to cabal, i.e., do not use `cabal build
+all`.
+
+### MacOS
+
+There are no installation requirements.
+
+### Windows
+
+There are no installation requirements.
```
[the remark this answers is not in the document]
jorisdral: RocksDB is by default not buildable on systems other than Linux

cabal.project
jeltsch: Just a question of curiosity from someone who might not fully understand the issue: Wouldn’t it be possible to move the conditional treatment inside `cabal.project.blockio-uring`, so that this file could always be imported but simply wouldn’t add anything in case of `serialblockio` being enabled?

PR #220, opened 2024-05-20
Improve `openSession`, implement `closeSession`, provide a new `withSession`

jorisdral:


src/Database/LSMTree/Common.hs
```diff
@@ -172,7 +242,13 @@ openSession hfs hbio dir = do
 -- lock will be released).
 --
 closeSession :: IOLike m => Session m -> m ()
-closeSession = undefined
+closeSession (Session (Internal.Session{Internal.sessionState})) =
+    RW.modify_ sessionState $ \case
+      Internal.SessionClosed -> pure Internal.SessionClosed
+      Internal.SessionOpen (Internal.SessionEnv {Internal.sessionFileLock}) -> do
+        -- TODO: close known table handles
```
jorisdral, line 248: I'm now reconsidering whether we should guarantee that table handles are closed when their session is closed. Maybe it's just simpler to tell the user: close all your handles before you close your session, otherwise you get exceptions

lsm-tree.cabal
```diff
@@ -83,9 +83,11 @@ library
     , base                     >=4.16      && <4.20
     , bitvec                   ^>=1.1
     , bytestring               ^>=0.11.4.0 || ^>=0.12.1.0
+    , concurrent-extra         ^>=0.7
     , containers
     , crc32c                   ^>=0.2.1
     , deepseq                  ^>=1.4      || ^>=1.5
+    , filelock                 ^>=0.1
```
jorisdral, line 90: Use lukko instead

jorisdral: I used the file locking from `base` instead

test/Test/Database/LSMTree/Internal.hs
```diff
@@ -0,0 +1,53 @@
...
+    try @LSMTreeError (openSession hfs hbio (FS.mkFsPath [])) >>= \case
+      Left (SessionDirLocked _) -> pure ()
+      _ -> assertFailure "Opening a session twice in the same directory \
+                         \should fail with an SessionDirLocked error."
+
+sessionDirLayoutMismatch :: Assertion
+sessionDirLayoutMismatch = withTempIOHasBlockIO "sessionDirLayoutMismatch" $ \hfs hbio -> do
+    FS.createDirectory hfs (FS.mkFsPath ["unexpected-directory"])
+    try @LSMTreeError (openSession hfs hbio (FS.mkFsPath [])) >>= \case
+      Left _ -> pure ()
```
mheinzel, line 44: This one could theoretically be either `SessionDirMissingActiveDir` or `SessionDirMissingSnapshotsDir`, right? The inconvenience of potentially having to match on both of these here made me realise that usually the difference doesn't matter to a user, either. So it might be a nicer API to have a single constructor `SessionDirMissingDir` or even `SessionDirMalformed` with further information inside?

mheinzel (approved):
Just minor remarks, use at your own discretion!

src/Database/LSMTree/Internal.hs
```diff
@@ -0,0 +1,359 @@
...
+    --
+    -- The lock is only released when an exception is raised, otherwise the lock
+    -- is included in the returned Session.
+    bracketOnError
+      (try
+        @m
+        @SomeException
+        (do lockFile <- System.openFile lf System.WriteMode
+            success <- GHC.hTryLock lockFile GHC.ExclusiveLock
+            pure $ if success then Just lockFile else Nothing
```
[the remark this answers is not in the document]
jorisdral: Good point -- I've added an `hClose`

test/Test/Database/LSMTree/Internal.hs
jorisdral: Yeah, the user probably doesn't really care which exception is thrown in which order. I've changed to a single constructor `SessionDirMalformed` carrying an `FsErrorPath`. The path will give some indication of where the maformation is located. The user might benefit from slightly more detailed information, but I think it's also fine as is

PR #221, opened 2024-05-21
Implement table `new` and `close`

jorisdral:


src/Database/LSMTree/Normal.hs
```diff
@@ -167,8 +167,9 @@ import           Database.LSMTree.Internal.Normal
 -- such instances in use at once.
 type TableHandle :: (Type -> Type) -> Type -> Type -> Type -> Type
 data TableHandle m k v blob = forall h. Typeable h =>
-    TableHandle (Internal.TableHandle m h)
+    TableHandle {-# UNPACK #-} !(Internal.TableHandle m h)
```
mheinzel, line 170: The `UNPACK` pragma goes through newtypes here, so you are unpacking the `RWVar` fields into here?

I'm always a bit cautious about `UNPACK`. Without looking into the Core I wouldn't be sure it's an improvement, i.e. there is no function being called down the line that takes an internal handle or `RWVar` for which it needs to be reboxed.

src/Database/LSMTree/Internal.hs
```diff
@@ -235,18 +255,36 @@ closeSession Session{sessionState} =
 --
 -- For more information, see 'Database.LSMTree.Normal.TableHandle'.
 --
--- NOTE: a table handle is __not__ thread-safe.
-data TableHandle m h = TableHandle {
-    tableSession     :: !(Session m h)
+-- Concurrent access to a table handle is mediated by a multiple-read,
+-- single-write lock. This allows concurrent read access, but only sequential
+-- write access. TODO: more fine-grained concurrent access than a top-level
+-- read-write lock.
```
mheinzel, line 261: Do I understand this correctly: This comment is not talking about read and write access in the sense of lookups and inserts, but reading and writing the handle open/close state? Every operation except `close` would be considered read access here (but can of course still do modifications to the internal `Mvar`s)?

mheinzel (approved):

src/Database/LSMTree/Internal.hs
jorisdral: The intended use would be to acquire a write lock for anything that updates any state in the handle, so not just `TableHandleState`. However, I kept the `MVar`s in there so we can implement more fine-grained concurrent access later. I can remove the MVars for now if that makes it clearer

src/Database/LSMTree/Normal.hs
jorisdral: I've removed the `UNPACK` for now. We can add them back later if we think it's worth it

src/Database/LSMTree/Internal.hs
mheinzel: That's very coarse-grained, but okay for a first version.

PR #224, opened 2024-05-22
Use vectors in the public API functions for lookups

jorisdral:
Most of the changes are mechnical. One TODO I left in the `lsm-tree-bench-wp8` module is that we'd preferably use vectors directly in the generation of batches. I'm first going to ensure that the public API also uses vectors for updates.

### DIFF
--- a/src/Database/LSMTree/Internal/Normal.hs
+++ b/src/Database/LSMTree/Internal/Normal.hs
@@ -7,10 +7,10 @@ module Database.LSMTree.Internal.Normal (
 import           Control.DeepSeq (NFData (..))
 
 -- | Result of a single point lookup.
-data LookupResult k v blobref =
-    NotFound      !k
-  | Found         !k !v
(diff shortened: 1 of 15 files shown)


src/Database/LSMTree/Internal/Normal.hs
```diff
@@ -7,10 +7,10 @@ module Database.LSMTree.Internal.Normal (
...
 
 -- | Result of a single point lookup.
-data LookupResult k v blobref =
-    NotFound      !k
-  | Found         !k !v
-  | FoundWithBlob !k !v !blobref
+data LookupResult v blobref =
+    NotFound
+  | Found         !v
+  | FoundWithBlob !v !blobref
```
mheinzel, line 13: I think that makes things a bit nicer and in many places we can get away without having to allocate tuples `(k, res)` instead.

dcoutts (approved):
I'm not sure sure why all the models and tests had to switch to vectors too, but ok.

PR #231, opened 2024-05-23
Use crc32c using system-cxx-std-lib

phadej:


phadej:
https://github.com/leptonyu/crc32c/pull/9

dcoutts (approved):
LGTM. `system-cxx-std-lib` FTW!

And thanks for teaching me about the existence of the package.
https://downloads.haskell.org/~ghc/9.4.2/docs/users_guide/packages.html#system-cxx-std-lib

PR #232, opened 2024-05-24
Add fcntl-nocache

phadej:


dcoutts (approved):
This also looks good, and is indeed a bit shorter than the file open style.

PR #234, opened 2024-05-30
Show unexpected outputs in session tests

jorisdral:
If https://github.com/IntersectMBO/lsm-tree/actions/runs/9225657061/job/25383623470#step:15:512 shows up again, then this will show what the unexpected output of the test was.


dcoutts (approved):
LGTM

PR #237, opened 2024-06-04
Refactorings related to `RunFsPaths` (and other paths)

jorisdral:
These changes are a preamble for #242. This PR is intended to be read commit by commit

### DIFF
--- a/src/Database/LSMTree/Internal/Paths.hs
+++ b/src/Database/LSMTree/Internal/Paths.hs
@@ -1,5 +1,14 @@
 module Database.LSMTree.Internal.Paths (
-    RunFsPaths (..)
+    SessionRoot (..)
+  , lockFile
+  , activeDir
+  , snapshotsDir
+  , snapshot
+    -- * Snapshot name
+  , SnapshotName
+  , mkSnapshotName
+    -- * Run paths
+  , RunFsPaths (..)
   , pathsForRunFiles
   , runKOpsPath
   , runBlobPath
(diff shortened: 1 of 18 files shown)


jorisdral:
I'm coverting this PR back to a draft because I may have a better solution that I am working on in #242

src/Database/LSMTree/Internal/Paths.hs
```diff
@@ -16,33 +24,100 @@ module Database.LSMTree.Internal.RunFsPaths (
...
 --    files
 --
 -- The representation doesn't store the full, name, just the number @n@. Use
 -- the accessor functions to get the actual names.
 --
-newtype RunFsPaths = RunFsPaths { runNumber :: Int }
-  deriving (Show, NFData)
+data RunFsPaths = RunFsPaths {
+    -- | The directory that run files live in.
+    runDir    :: !FsPath
```
mheinzel, line 115: Alternatively, we could store the `SessionRoot` here. Otherwise, everyone constructing a `RunFsPaths` has to take care of adding the `active` subdirectory, which for example is not done in the tests and benchmarks.
Seems nicer to handle that concern (runs are always located in `${root}/active/${n}.keyops" etc.) in one central place, i.e. here.

mheinzel (approved):
5e3fc6bed6a2e1ab85cc5e2a0fbbfc6a1244881e is only very loosely related to the rest of the PR, but it's small, so I'm okay with including it.

I like the refactoring with the new `Paths` module. And good catch on `RunFsPaths` only working in the `HasFS` root dir! I just added one remark on the fix. Did this come up through some test and if not, can we add one?

jorisdral:
> I like the refactoring with the new `Paths` module. And good catch on `RunFsPaths` only working in the `HasFS` root dir! I just added one remark on the fix. Did this come up through some test and if not, can we add one?

It came up in #242, but it was also something I had spotted a a while ago and planned on changing . I used to have some tests on this branch -- I'll see if I can add them again

src/Database/LSMTree/Internal/Paths.hs
jorisdral: As a first version, I had the `SessionRoot` in there, but I think it's nicer if the `RunFsPaths` knows only about the directory it lives in. For most of our tests of the internals, it's not important that a run has a parent directory that is called `active` (which itself has a parent directory that is `${root}`). It inverts the responsibility: runs can now live anywhere by themselves, but it's the table/session that picks the `${root}/active` directory for the runs to live in, and that happens in a central place. There should be a single definition for how to create a run's paths in a session/table however, so what if I add this? Does that look reasonable?
 ```haskell
runPath :: SessionRoot -> Int -> RunFsPaths
runPath root = RunFsPaths (activeDir root)
```

> Otherwise, everyone constructing a RunFsPaths has to take care of adding the active subdirectory, which for example is not done in the tests and benchmarks.

So with this approach, none of our tests/benchmarks have to create additional directories. If the tests/benchmarks do pass in a `runDir` that does not exist, then our code will now fail early with an error telling us the directory does not exist (https://github.com/IntersectMBO/lsm-tree/pull/237/), which I think is also nicer because there is no silent directory creation that can cause errors down the line (which I experienced in #242)

File: test/Test/Database/LSMTree/Internal/RunBuilder.hs [hunk 2]
```diff
@@ -0,0 +1,58 @@
...
+
+-- | 'new' in an existing directory should be succesfull.
+prop_newInExistingDir :: HasFS IO HandleIO -> IO Property
+prop_newInExistingDir hfs = do
+    let runDir = FS.mkFsPath ["a", "b", "c"]
+    FS.createDirectoryIfMissing hfs True runDir
+    try (RunBuilder.new hfs (RunFsPaths runDir 17) (NumEntries 0) 0) <&> \case
+      Left e@FS.FsError{} ->
+        counterexample ("expected a success, but got: " <> show e) $ property False
+      Right _ -> property True
```
mheinzel, line 36: This seems like an individual test case, why are we running it as a property?

mheinzel (approved):
I like the adjustments, looks good to me!

test/Test/Database/LSMTree/Internal/RunBuilder.hs
jorisdral: The property is run only once because there are no arguments we have to generate, so it's isomorphic to `HUnit` test case

dcoutts (commented):
LGTM. I like the patch-by-patch PR. Easy to follow.

PR #239, opened 2024-06-05
Use `generateM` in `indexSearches`

jorisdral:


dcoutts (approved):
Ok, and then potentially this allows it to fuse with a consumer like `V.zipWith`.

PR #242, opened 2024-06-07
Interim implementation of `snapshot` and `open`

jorisdral:


src/Database/LSMTree/Normal.hs
```diff
@@ -275,6 +293,36 @@ retrieveBlobs = undefined
...
+  => Session m
+  -> Internal.TableConfig
+     -- ^ The table config is not persisted to disk (yet) until we have
+     -- snapshots. The user has to make sure to only load runs that were created
+     -- with the same configuration, or the behaviour might be undefined.
+  -> V.Vector (V.Vector Internal.RunFsPaths)
+     -- ^ Runs to open, identified by their 'RunFsPaths's. Acquire this
+     -- structure using 'snapshot''.
+  -> m (TableHandle m k v blob)
+open' (Session sesh) conf paths = TableHandle <$> Internal.open' sesh conf paths
```
jorisdral, line 324: An alternative I've been thinking of: write the `RunFsPaths`s to disk instead of passing them in memory, and use the `SnapshotName` to identify the file that contains the `RunFsPaths`s

src/Database/LSMTree/Internal.hs
```diff
@@ -403,39 +417,63 @@ mkLevelsCache lvls = LevelsCache_ {
...
+newWithLevels ::
+     m ~ IO -- TODO: replace by @io-classes@ constraints for IO simulation.
+  => Session m h
+  -> TableConfig
+  -> (forall a. HasFS m h -> Cont (m a) (V.Vector (Level (Handle h))))
+     -- ^ Resources have to be cleaned up correctly if an (asynchronous)
+     -- exception occurs. We achieve this by requiring that the resource acquire
+     -- action is a 'Cont'inuation. See 'openRun' for more information.
+  -> m (TableHandle m h)
+newWithLevels sesh conf acqLevels = do
```
mheinzel, line 449: I think we can re-use the code without having to parametrise by a function. Currently, the structure is roughly:

```haskell
new = newWithLevels (const empty)
open = newWithLevels openLevels
newWithLevels acq = do
  withOpenSession $ do
    runCont acq $ \levels ->
      _  -- create table

```

But this seems a little simpler:

```haskell
new =
  withCont $ do
    createTables empty
open = do
  withOpenSession $
    runCont acq $ \levels ->
      createTable levels
createTable levels = _  -- create table
```

src/Database/LSMTree/Internal.hs
```diff
@@ -492,6 +530,100 @@ toNormalLookupResult = \case
...
+{-# SPECIALISE openRun :: HasFS IO h -> RunFsPaths -> Cont (IO a) (Run (FS.Handle h)) #-}
+-- | Open a run.
+--
+-- If an error occurs when opening multiple runs in sequence, then we have to
+-- make sure that all runs that have been succesfully opened already are closed
+-- again. The 'Cont'inuation allows us to do so by folding 'bracketOnError' over
+-- the action @m a@.
+openRun ::
+     m ~ IO -- TODO: replace by @io-classes@ constraints for IO simulation.
+  => HasFS m h -> RunFsPaths -> Cont (m a) (Run (FS.Handle h))
```
mheinzel, line 602: As discussed, there might be a simpler solution. Even just using some function could make it easier to understand what's happening without knowing `Cont` well, also in terms of strictness of `>>=`, which is relevant for `mapMStrict`.

src/Database/LSMTree/Internal/Entry.hs
```diff
@@ -37,6 +38,12 @@ data Entry v blobref
     | Delete
   deriving (Eq, Show, Functor, Foldable, Traversable)
 
+hasBlob :: Entry v blobref -> Bool
+hasBlob Insert{}         = False
+hasBlob InsertWithBlob{} = True
+hasBlob Mupdate{}        = False
+hasBlob Delete{}         = False
```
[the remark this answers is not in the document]
jorisdral: Yes, quite right

src/Database/LSMTree/Internal/Paths.hs
```diff
@@ -16,33 +24,100 @@ module Database.LSMTree.Internal.RunFsPaths (
...
 --
 -- The representation doesn't store the full, name, just the number @n@. Use
 -- the accessor functions to get the actual names.
 --
-newtype RunFsPaths = RunFsPaths { runNumber :: Int }
-  deriving (Show, NFData)
+data RunFsPaths = RunFsPaths {
+    -- | The directory that run files live in.
+    runDir    :: !FsPath
+  , runNumber :: !Integer }
```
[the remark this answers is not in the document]
jorisdral: Yes, you're right, it's probably overkill. I am going to change it to `Word64`, however, since we're never using negative numbers anyway

src/Database/LSMTree/Internal.hs
jorisdral: That indeed simplified the code a bit. Thanks!

src/Database/LSMTree/Internal.hs
jorisdral: I'm pretty happy with the current implementation. I've gone through a few implementations, and this is the most concise and doesn't require us to fiddle with the vectors -- we get the structure for free using the `Cont` monad. I've updated the documentation a bit, and I've added an example in there too

RE the strictness concerns, you're right that the strictness depends on the implementation of bind. To verify whether the resulting levels don't contain thunks, I've added a `NoThunks` assertion that is checked before opening a table:

```haskell
assertNoThunks :: NoThunks a => a -> b -> b
assertNoThunks x = assert p
  where p = case unsafeNoThunks x of
              Nothing -> True
              Just thunkInfo -> error $ "Assertion failed: found thunk" <> show thunkInfo
```
It's slightly hacky because I'm hijacking the `assert` functionality to show a custom `assert` error message. The `Bool` I'm passing to the `assert` contains an `error` itself. :stuck_out_tongue: The nice thing is that the `error` is erased when compiled with optimisations

jorisdral: Example: when I remove the `$!` from `mapMStrict`
```diff
-mapMStrict f v = V.mapM (f >=> (pure $!)) v
+mapMStrict f v = V.mapM f v
```
Then this test shows a `NoThunks` assertion error
```
>  cabal run lsm-tree-test -- -p "interim"
lsm-tree
  Test.Database.LSMTree.Internal
    prop_interimOpenTable: OK (0.01s)
      +++ OK, failed as expected. (after 1 test and 8 shrinks):
      Exception:
        Assertion failed: found thunkThunkInfo {thunkInfo = Left ["Boxed.Vector","Boxed.Vector"]}
          error, called at src/Database/LSMTree/Internal/Assertions.hs:43:33 in lsm-tree-0.1.0.0-inplace:Database.LSMTree.Internal.Assertions
      InMemLookupData {runData = fromList [(SerialisedKey [0,0,0,0,0,0],Delete)], lookups = []}
```

mheinzel: Nice trick with the `assert`!

Question on exception handling: Do we need to include constructing the table handle (once we have the levels of loaded runs) into the `bracket` or can we do that outside? Similarly, if removing one reference (potentially closing a run) throws, should we continue trying to close the others or can we just give up?

I have one alternative implementation I have in mind that especially works well when answering "no" above, but can be adjusted. I think the complications here stem from having two stacks we want to keep simultaneously: the one tracking where we are in constructing the vectors, and the one tracking which runs need to be cleaned up. We could resort to managing the second stack ourselves with something like this:

```haskell
openLevels ::
     m ~ IO
  => HasFS m h
  -> V.Vector (V.Vector RunFsPaths)
  -> m (V.Vector (Level (Handle h)))
openLevels hfs levels = do
    openedRuns <- newIORef []
    -- if something fails, clean up all runs opened so far
    flip onException (traverse_ (Run.removeReference hfs) =<< readIORef openedRuns) $
      flip V.mapMStrict levels $ \l ->
        fmap Level $ flip V.mapMStrict l $ \p -> do
          r <- Run.openFromDisk hfs p
          modifyIORef' openedRuns (r :)
          return r
```

Maybe a bit more straightforward to see what's going on?

jorisdral: > Question on exception handling: ...

I think "yes" to both questions. Before proper snapshots are implemented, so in our current situation, runs are created, but never hard-linked or deleted. So, we'd run the risk of leaking file handles if we don't close runs properly, and with proper snapshots we might link leak (hard-links to) files. Hopefully GHC cleans the file descriptors up if we don't close the runs manually, but there is no guarantee even if the file handles have a finaliser attached. And when we start caring about linking/deleting files too, then those are going to leak regardless. The best thing is to just be extra sure that we clean up resources

mheinzel (approved):
Nice!

PR #244, opened 2024-06-07
Add the core of the ordinary index support

jeltsch:
This adds support for the ordinary index apart from serialization.

### DIFF
--- a/src/Database/LSMTree/Internal/IndexOrdinary.hs
+++ b/src/Database/LSMTree/Internal/IndexOrdinary.hs
@@ -1,0 +1,181 @@
+{- HLINT ignore "Avoid restricted alias" -}
+
+-- | A general-purpose fence pointer index.
+module Database.LSMTree.Internal.IndexOrdinary
+(
+    IndexOrdinary (IndexOrdinary),
+
+    -- * Search
+    search,
+
+    -- * Deserialisation
+    fromSBS
+)
+where
+
+import           Prelude hiding (drop, length, takeWhile)
+
+import           Control.Exception (assert)
+import           Control.Monad (when)
+import           Data.ByteString.Short (ShortByteString (SBS))
+import qualified Data.ByteString.Short as ShortByteString (length)
+import           Data.Primitive.ByteArray (ByteArray (ByteArray),
+                     indexByteArray)
+import           Data.Vector (Vector, drop, findIndexR, fromList, length,
+                     takeWhile, (!))
+import qualified Data.Vector.Primitive as Primitive (Vector (Vector), drop,
+                     force, length, null, splitAt, take)
+import           Data.Word (Word16, Word32, Word64, Word8, byteSwap32)
+import           Database.LSMTree.Internal.Entry (NumEntries (NumEntries))
+import           Database.LSMTree.Internal.IndexCompact (PageNo (PageNo),
+                     PageSpan (PageSpan))
+import           Database.LSMTree.Internal.Serialise
+                     (SerialisedKey (SerialisedKey'))
+import           Database.LSMTree.Internal.Vector (binarySearchL, mkPrimVector)
+
+{-|
+    The version of the index and its serialisation format that is supported by
+    this module.
+-}
+supportedVersion :: Word32
+supportedVersion = 1
+
+{-|
+    A general-purpose fence pointer index.
+
+    An index is represented by a vector that maps the number of each page to the
+    key stored last in this page or, if the page is an overflow page, to the key
+    of the corresponding key–value pair. The vector must have the following
+    properties:
+
+      * It is non-empty.
+
+      * Its elements are non-decreasing.
+
+    This restriction follows from the fact that a run must contain keys in
+    ascending order and must comprise at least one page for 'search' to be able
+    to return a valid page span.
+-}
+newtype IndexOrdinary = IndexOrdinary (Vector SerialisedKey)
+
+-- * Search
+
+{-|
+    Searches for a page span that contains a key–value pair with the given key.
+    If there is indeed such a pair, the result is the corresponding page span;
+    if there is no such pair, the result is an arbitrary but valid page span.
+-}
+search :: SerialisedKey -> IndexOrdinary -> PageSpan
+search key (IndexOrdinary lastKeys)
+    = assert (length lastKeys > 0) $ PageSpan (PageNo start) (PageNo end)
+    where
+
+    start :: Int
+    start | protoStart < pageCount = protoStart
+          | otherwise              = maybe 0 succ $
+                                     findIndexR (/= lastKeys ! pred pageCount) $
+                                     lastKeys
+
+    end :: Int
+    end | protoStart < pageCount
+            = start + length (takeWhile (== key) (drop (succ start) lastKeys))
+        | otherwise
+            = pred pageCount
+
+    protoStart :: Int
+    protoStart = binarySearchL lastKeys key
+
+    pageCount :: Int
+    pageCount = length lastKeys
+
+-- * Deserialisation
+
+{-|
+    Reads an index along with the number of entries of the respective run from a
+    byte string.
+
+    The byte string must contain the serialised index exactly, with no leading
+    or trailing space.
+-}
+fromSBS :: ShortByteString -> Either String (NumEntries, IndexOrdinary)
+fromSBS shortByteString@(SBS unliftedByteArray)
+    | fullSize < 12
+        = Left "Doesn't contain header and footer"
+    | version == byteSwap32 supportedVersion
+        = Left "Non-matching endianness"
+    | version /= supportedVersion
+        = Left "Unsupported version"
+    | otherwise
+        = (,) <$> entriesCount <*> index
+    where
+
+    fullSize :: Int
+    fullSize = ShortByteString.length shortByteString
+
+    byteArray :: ByteArray
+    byteArray = ByteArray unliftedByteArray
+
+    fullBytes :: Primitive.Vector Word8
+    fullBytes = mkPrimVector 0 fullSize byteArray
+
+    version :: Word32
+    version = indexByteArray byteArray 0
+
+    postVersionBytes :: Primitive.Vector Word8
+    postVersionBytes = Primitive.drop 4 fullBytes
+
+    lastKeysBytes, entriesCountBytes :: Primitive.Vector Word8
+    (lastKeysBytes, entriesCountBytes)
+        = Primitive.splitAt (fullSize - 12) postVersionBytes
+
+    entriesCount :: Either String NumEntries
+    entriesCount
+        | (fromIntegral asWord64 :: Integer) > fromIntegral (maxBound :: Int)
+            = Left "Number of entries not representable as Int"
+        | otherwise
+            = Right (NumEntries (fromIntegral asWord64))
+        where
+
+        asWord64 :: Word64
+        asWord64 = indexByteArray entriesCountRep 0
+
+        entriesCountRep :: ByteArray
+        Primitive.Vector _ _ entriesCountRep = Primitive.force entriesCountBytes
+
+    index :: Either String IndexOrdinary
+    index = IndexOrdinary <$> fromList <$> lastKeys lastKeysBytes
+
+    lastKeys :: Primitive.Vector Word8 -> Either String [SerialisedKey]
+    lastKeys bytes
+        | Primitive.null bytes
+            = Right []
+        | otherwise
+            = do
+                  when (Primitive.length bytes < 2)
+                       (Left "Too few bytes for key size")
+                  let
+
+                      firstSizeRep :: ByteArray
+                      Primitive.Vector _ _ firstSizeRep
+                          = Primitive.force (Primitive.take 2 bytes)
+
+                      firstSize :: Int
+                      firstSize = fromIntegral $
+                                  (indexByteArray firstSizeRep 0 :: Word16)
+
+                      postFirstSizeBytes :: Primitive.Vector Word8
+                      postFirstSizeBytes = Primitive.drop 2 bytes
+
+                  when (Primitive.length postFirstSizeBytes < firstSize)
+                       (Left "Too few bytes for key")
+                  let
+
+                      firstBytes, othersBytes :: Primitive.Vector Word8
+                      (firstBytes, othersBytes)
+                          = Primitive.splitAt firstSize postFirstSizeBytes
+
+                      first :: SerialisedKey
+                      first = SerialisedKey' (Primitive.force firstBytes)
+
+                  others <- lastKeys othersBytes
+                  return (first : others)
--- a/lsm-tree.cabal
+++ b/lsm-tree.cabal
@@ -55,6 +55,7 @@ library
     Database.LSMTree.Internal.Entry
     Database.LSMTree.Internal.IndexCompact
     Database.LSMTree.Internal.IndexCompactAcc
+    Database.LSMTree.Internal.IndexOrdinary
     Database.LSMTree.Internal.Lookup
     Database.LSMTree.Internal.Managed
     Database.LSMTree.Internal.Merge


src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,222 @@
...
+    start | protoStart < pageCount = protoStart
+          | otherwise              = maybe 0 succ $
+                                     findIndexR (/= lastKeys ! pred pageCount) $
+                                     lastKeys
+
+    end :: Int
+    end | protoStart < pageCount
+            = start + length (takeWhile (== key) (drop (succ start) lastKeys))
+        | otherwise
+            = pred pageCount
```
jorisdral, line 115: It looks to me like this could return `-1`

src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,222 @@
...
+    Returns the span of pages that contain the key–value pair of a certain key
+    if there is such a pair, or an arbitrary, valid page span if there is not.
+-}
+search :: SerialisedKey -> IndexOrdinary -> PageSpan
+search key (IndexOrdinary lastKeys) = PageSpan (PageNo start) (PageNo end) where
+
+    start :: Int
+    start | protoStart < pageCount = protoStart
+          | otherwise              = maybe 0 succ $
+                                     findIndexR (/= lastKeys ! pred pageCount) $
```
jorisdral, line 108: Also: possibly using `-1` here as the index

src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,222 @@
...
+
+    version :: Word32
+    version = indexByteArray byteArray 0
+
+    postVersionBytes :: Primitive.Vector Word8
+    postVersionBytes = Primitive.drop 4 fullBytes
+
+    lastKeysBytes, entriesCountBytes :: Primitive.Vector Word8
+    (lastKeysBytes, entriesCountBytes)
+        = Primitive.splitAt (fullSize - 12) postVersionBytes
```
jorisdral, line 170:
```suggestion
    lastKeysBytes, entriesCountBytes :: Primitive.Vector Word8
    (lastKeysBytes, entriesCountBytes)
        = Primitive.splitAt (fullSize - 12) fullBytes
```

src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,222 @@
...
+            = Left "Number of entries not representable as Int"
+        | otherwise
+            = Right (NumEntries (fromIntegral asWord64))
+        where
+
+        asWord64 :: Word64
+        asWord64 = indexByteArray entriesCountRep 0
+
+        entriesCountRep :: ByteArray
+        Primitive.Vector _ _ entriesCountRep = Primitive.force entriesCountBytes
```
jorisdral, line 184: I don't think this `force` is required

src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,222 @@
...
+            = Right []
+        | otherwise
+            = do
+                  when (Primitive.length bytes < 2)
+                       (Left "Too few bytes for key size")
+                  let
+
+                      firstSizeRep :: ByteArray
+                      Primitive.Vector _ _ firstSizeRep
+                          = Primitive.force (Primitive.take 2 bytes)
```
jorisdral, line 201: This `force` is not necessary

src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,222 @@
...
+        | otherwise
+            = pred pageCount
+
+    protoStart :: Int
+    protoStart = searchForProtoStart 0 (length lastKeys)
+
+    searchForProtoStart :: Int -> Int -> Int
+    searchForProtoStart first count
+        | count == 0               = first
+        | key <= lastKeys ! center = searchForProtoStart first frontCount
```
jorisdral, line 123: You can use unsafe indexing here, as long as you are sure the number you're indexing at is not out of bounds

src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,222 @@
...
+newtype IndexOrdinary = IndexOrdinary (Vector SerialisedKey)
+
+-- * Search
+
+{-|
+    Returns the span of pages that contain the key–value pair of a certain key
+    if there is such a pair, or an arbitrary, valid page span if there is not.
+-}
+search :: SerialisedKey -> IndexOrdinary -> PageSpan
+search key (IndexOrdinary lastKeys) = PageSpan (PageNo start) (PageNo end) where
```
jorisdral, line 103: Some bang patterns in this function could remove some allocations. Maybe we should also add a micro-benchmark in the style of the compact index micro-benchmark, so we can see whether there are optimisation opportunities

jorisdral (commented):
> This adds support for the ordinary index apart from serialization.

Another missing feature is incremental construction, or is that included in the serialisation?

Are you planning on adding tests in a follow-up PR?

jeltsch:
> Another missing feature is incremental construction, or is that included in the serialisation?

I understood that serialization is supposed to be incremental and thus incremental construction is entangled with serialization. Consequently, I wanted to include incremental construction in the serialization part. Did I get something wrong here?

> Are you planning on adding tests in a follow-up PR?

The testing issue is an interesting one. My understanding is that there should ultimately be a common interface for indexes, with instances for the compact and the ordinary index. Tests could and perhaps should then be written for indexes as such, using said interface, and run for the different instances. My impression is that adding tests specifically for the ordinary index would merely boil down to copying compact-index tests, which wouldn’t seem like a good idea to me. What are your thoughts about this?

src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,222 @@
...
+           PageNo (PageNo),
+           PageSpan (PageSpan)
+       )
+
+{-|
+    The version of the index and its serialisation format that is supported by
+    this module.
+-}
+supportedVersion :: Word32
+supportedVersion = 1
```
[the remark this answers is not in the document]
jeltsch: We had a similar discussion outside GitHub already. The question there was whether to have separate version spaces for the compact and the ordinary index, so that version numbers can coincide, or whether to have version numbers that encode also the kind of index used. @dcoutts said that there hasn’t been made a decision about this yet.

For now, I propose to just leave things as they are, as to not introduce additional complication that might not be needed in the end, but file an issue that calls for deciding this matter and then potentially change code accordingly. Does this sound good?

src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,222 @@
...
+           byteSwap32
+       )
+import Data.Primitive.ByteArray
+       (
+           ByteArray,
+           indexByteArray
+       )
+import Data.ByteString.Short
+       (
+           ShortByteString (ShortByteString)
```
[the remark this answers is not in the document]
jeltsch: Fixed in 97adfa23a4cc1d64ab0f0efe098645fc5838b3b0.

src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,222 @@
...
+    fromSBS
+)
+where
+
+import Prelude hiding (length, drop, takeWhile)
+
+import Control.Monad
+       (
+           when
+       )
```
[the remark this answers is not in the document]
jeltsch: Done. The result is in 983168f9c8f1dc9c580a7395d4a29e7da613e270.

src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,222 @@
...
+{-|
+    Returns the span of pages that contain the key–value pair of a certain key
+    if there is such a pair, or an arbitrary, valid page span if there is not.
+-}
+search :: SerialisedKey -> IndexOrdinary -> PageSpan
+search key (IndexOrdinary lastKeys) = PageSpan (PageNo start) (PageNo end) where
+
+    start :: Int
+    start | protoStart < pageCount = protoStart
+          | otherwise              = maybe 0 succ $
```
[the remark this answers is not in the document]
jeltsch: Hmm, what I see in some other modules looks like a mix of 2-space indentation and 4-space indentation to me. To some extend, this is because of an indentation style that partly uses two indentations for marking one logical level, but in other places I even cannot see why 4 spaces are used.

My opinion is that 4-space indentation makes the indentation structure more easily recognizable and is not really a problem if you avoid those two indentations for one logical level. Therefore, I’d like to keep it, but if there is a strong urge to have this module use 2-space indentation I may change the indentation depth.

src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,222 @@
...
+        = Left "Unsupported version"
+    | otherwise
+        = (,) <$> entriesCount <*> index
+    where
+
+    fullSize :: Int
+    fullSize = ShortByteString.length shortByteString
+
+    fullBytes :: Primitive.Vector Word8
+    fullBytes = Primitive.Vector 0 fullSize byteArray
```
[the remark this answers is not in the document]
jeltsch: Changed in 17907a90f773b44814220182c1be3520b6191f19.

src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,222 @@
...
+    A general-purpose fence pointer index.
+
+    An index is represented by a vector that maps the number of each page to the
+    key stored last in this page or, if the page is an overflow page, to the key
+    of the corresponding key–value pair. From this, the following constraints on
+    the vector follow:
+
+      * The vector must be non-empty.
+
+      * The elements of the vector must be non-decreasing.
```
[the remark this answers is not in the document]
jeltsch: That the elements of the vector must be non-decreasing follows from the constraints on runs in conjunction with the way indexes are represented, but that the vector must be non-empty indeed doesn’t so clearly follow from anything: you could only argue that runs are supposed to be non-empty and the non-emptiness of the vector follows from this. You’re absolutely right, though, that the non-emptiness is required for `search` working properly. I’ll try to make the above documentation clearer.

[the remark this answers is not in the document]
jeltsch: To be precise, it’s not just my particular implementation of `search` that doesn’t work with an empty vector, but no correct implementation of `search` can work with an empty vector, since an empty vector would represent a run without pages and there are no valid page spans to return for such a run.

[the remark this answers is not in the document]
jeltsch: The improved documentation is in 4948b05b99ef7eac5f0a5260ab4aa4acdd2a39a5.

src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,222 @@
...
+-}
+newtype IndexOrdinary = IndexOrdinary (Vector SerialisedKey)
+
+-- * Search
+
+{-|
+    Returns the span of pages that contain the key–value pair of a certain key
+    if there is such a pair, or an arbitrary, valid page span if there is not.
```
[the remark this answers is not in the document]
jeltsch: Nitpicking is welcome: I like precision. 🙂

That said, I think that my specification is correct, since it only says that a page span containing the search key is returned *if there is such a pair*. It doesn’t promised that the returned page span contains the key, since it could also have been returned because of the second part of the result specification: “an arbitrary, valid page span if there is not [an entry with the search key]”.

src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,222 @@
...
+    searchForProtoStart first count
+        | count == 0               = first
+        | key <= lastKeys ! center = searchForProtoStart first frontCount
+        | otherwise                = searchForProtoStart (succ center) rearCount
+        where
+
+        frontCount, rearCount, center :: Int
+        frontCount = count `div` 2
+        rearCount  = pred (count - frontCount)
+        center     = first + frontCount
```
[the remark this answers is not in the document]
jeltsch: I crafted the local search function `searchForProtoStart` specifically for finding the start of a page span and was under the impression that because of some specifics it isn’t appropriate as a general-purpose function. However, I realize now that it really just performs an ordinary binary search that finds the first occurrence of a value in case there are multiple occurrences. Therefore, it is sensible to turn it into sort of a library function. I’ve created #252, which calls for this.

src/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: Yes, it could, but only if the run contains no pages, which should be disallowed anyhow, because no sensible page span can be returned in this case (as also argued in [my comment on the documentation of `search`](https://github.com/IntersectMBO/lsm-tree/pull/244#discussion_r1636656680)).

That said, it might not be a good idea to silently return a page span from 0 to −1 in case of no pages. Maybe we should add an assertion that the length of the vector representing the index is not zero somewhere, perhaps at the outside of the right-hand side of the `search` definition. What is your opinion on this?

src/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: I think that my version is the correct one. We only want the payload (`lastKeysBytes`) and the footer (`entriesCountBytes`) and not the header; so we should start with `postVersionBytes`, which we got from `fullBytes` by removing the header. Of course, it seems strange that in the size calculation we then use `fullSize`. However, this is just about numbers, and we do get the size of the payload indeed by subtracting 12 from the full size, because 12 is the total size of the header and footer. Of course, it would be a bit more intuitive to use some `postVersionSize` here, but introducing this seemed overkill to me.

src/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: I think it is. 😉 The thing is that the byte array underlying `entriesCountBytes` can comprise more than just the bytes that represent the number of entries. In fact, this is always the case here, because `entriesCountBytes` is obtained from `fullBytes` by slicing and thus the byte array is the one that contains the whole byte sequence that is deserialized.

src/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: I think it is, for the same reasons as in the case discussed [above](https://github.com/IntersectMBO/lsm-tree/pull/244#discussion_r1638534358).

src/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: It seems that it is guaranteed that the index is not out of bounds. Still, I’m reluctant to use unsafe indexing here. If I was wrong and the index *can* be out of bounds, the behavior of the software becomes unpredictable, to the point that we cannot rely on property-based tests anymore. The index checking is probably very cheap, and for each search there is only a logarithmic number of checks to be made (for a run comprising 4 GB it would be about 20 checks). I would favor safety over performance here. What is your opinion on this?

src/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: This is a topic where I’m far from being an expert. Can you perhaps give suggestions as to where putting bang patterns would be appropriate.

src/Database/LSMTree/Internal/IndexOrdinary.hs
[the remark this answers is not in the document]
jorisdral: Sure, let's make an issue. My intuition says we should just use the version number to encode both the index type and version: there are enough bits in a `Word32` to represent both

src/Database/LSMTree/Internal/IndexOrdinary.hs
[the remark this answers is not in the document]
jorisdral: You're right, we use a mix: I was under the impression most of us had used 2-space indentation as the baseline. I'm not really sure what you mean by "... is not really a problem if you avoid those two indentations for one logical level". Could you elaborate?

IMO 2-space indentation saves a lot of horizontal real estate in source code, and that's why I prefer it. For example, [these lines](https://github.com/IntersectMBO/lsm-tree/pull/244/files#diff-d9780d09d4424733782603d355272b53fe91fcc4f67d60ed2edaa180f266be5aR181-R182) have an indentation of 24 spaces. However, we have no strict rules about style that so it's fine to keep the 4-space indentation here. Our only guideline is to try to keep formatting within a module consistent, but even that is not a strict rule

src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,222 @@
...
+import Database.LSMTree.Internal.IndexCompact
+       (
+           PageNo (PageNo),
+           PageSpan (PageSpan)
+       )
+
+{-|
+    The version of the index and its serialisation format that is supported by
+    this module.
+-}
```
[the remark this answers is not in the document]
jorisdral: > As with the previous point, I’d like to stick to the present comment style but would consider changing it if there is a strong urge to use single-line comments for documentation comments.

Again, that is fine -- same reasoning as https://github.com/IntersectMBO/lsm-tree/pull/244#discussion_r1639696227

src/Database/LSMTree/Internal/IndexOrdinary.hs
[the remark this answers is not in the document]
jorisdral: > The reason for me not using `vector-algorithms` is precisely that it only works with mutable vectors. I guess that it would indeed be possible to use it by employing `unsafeThaw`. However, I find it questionable to use an unsafe operation where there is a safe alternative, just because of limitations of current libraries.

In this case, it is a safe use of `unsafeThaw`: it is only thawed within a local scope where no alterations to the vector are made. Maybe it's a shortcoming of the library that they don't have a version on immutable vectors, but I think it's sensible to use here. IMO we care more about performance than strict safety in the `lsm-tree` library, and it's still safe to use `unsafeThaw`, because of reasoning stated above. Moreover, `vector-algorithms` is to some degree tested/benchmarked/battle-tested, so that's also an advantage

> I crafted the local search function searchForProtoStart specifically for finding the start of a page span and was under the impression that because of some specifics it isn’t appropriate as a general-purpose function. However, I realize now that it really just performs an ordinary binary search that finds the first occurrence of a value in case there are multiple occurrences. Therefore, it is sensible to turn it into sort of a library function. I’ve created https://github.com/IntersectMBO/lsm-tree/issues/252, which calls for this.

:+1: I think this may also be necessary to some degree even if we use `vector-algorithms`, depending on which element is returned by those search functions

src/Database/LSMTree/Internal/IndexOrdinary.hs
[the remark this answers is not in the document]
jorisdral: Okay, then I think the wording confuses me. It would probably be useful to have a unified specification of the `search` result, and I think we can find a middle ground between the description of this `search` function and the one for the compact index. In particular, I think it's clearer if we mention something along the lines of these two statements:

* `search` always returns a valid page span.
* `search` may return false positives, but not false negatives.

src/Database/LSMTree/Internal/IndexOrdinary.hs
jorisdral: It doesn't have to happen in this PR, it can be a follow-up. A micro-benchmark can tell us the allocation rate, and the allocation rate can be reduced by ensuring that GHC unboxes values (using bang patterns to give GHC hints about what to unbox). In any case, a micro-benchmark would be necessary to really evaluate performance

src/Database/LSMTree/Internal/IndexOrdinary.hs
jorisdral: An `assert` could be useful, yes

src/Database/LSMTree/Internal/IndexOrdinary.hs
jorisdral: `entiresCountRep` is only used in `asWord64`, and the byte array is not kept alive after that, that's why I think it's not necessary

src/Database/LSMTree/Internal/IndexOrdinary.hs
jorisdral: `firstSizeRep` is here only used in `firstSize`, and it is not kept alive after that

src/Database/LSMTree/Internal/IndexOrdinary.hs
jorisdral: Same reasoning as here https://github.com/IntersectMBO/lsm-tree/pull/244#discussion_r1639708395, if we locally use unsafe functions in a safe manner, then I think that's preferable

src/Database/LSMTree/Internal/IndexOrdinary.hs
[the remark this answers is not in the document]
jeltsch: > I'm not really sure what you mean by "... is not really a problem if you avoid those two indentations for one logical level". Could you elaborate?

Look at this code, for example:

https://github.com/IntersectMBO/lsm-tree/blob/2f5e9827fe2b8758fff32935628b2f3da7239a47/src/Database/LSMTree/Internal/IndexCompact.hs#L882-L887

The `where` is indented relative to the definition of `bitIndexFromTo`, and the local definitions are indented relative to the `where`, although they are just subordinate to `bitIndexFromTo` and thus logically one level below it. If you would use this indentation style with 4 spaces, you would indent the local definitions 8 spaces relative to the definition of `bitIndexFromTo`, which would be problematic.

> IMO 2-space indentation saves a lot of horizontal real estate in source code, and that's why I prefer it.

Yes, the use of horizontal space is an issue, particularly with Haskell, where code tends to be wider than with typical imperative languages. It’s a tradeoff between readability and economy. Actually, a while ago I came to the conclusion that 3-space indentation might be a good compromise but decided against using it, also in my personal projects, because it seemed too exotic (although it’s part of the standard style for Ada, the programming language).

I think I’ll stick to 4-space indentation for now.

src/Database/LSMTree/Internal/IndexOrdinary.hs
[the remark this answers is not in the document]
jeltsch: Okay, I’ll keep it then.

src/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: It’s not about resource usage but about correctness: The array that `entriesCountRep` contains is the original array, which contains the whole serialized index, and the fact that `entriesCountRep` spans only a small part of it comes from the other two fields of the primitive vector saying so. These other two fields we lose when converting a primitive vector into an array via pattern matching, and for the result to be correct we have to prune the array before the conversion, which we do using `force`.

src/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: And also my [above follow-up comment](https://github.com/IntersectMBO/lsm-tree/pull/244#discussion_r1644652360) applies here, I think. 🙂

src/Database/LSMTree/Internal/IndexOrdinary.hs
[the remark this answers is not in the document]
jeltsch: Let’s continue this discussion over at #256, where I’ve already created [a comment](https://github.com/IntersectMBO/lsm-tree/pull/256#issuecomment-2178870318) summarizing the state of this discussion. We should leave this thread open however, to remind ourselves that we actually have to use the search function of #256 in the code of this pull request.

src/Database/LSMTree/Internal/IndexOrdinary.hs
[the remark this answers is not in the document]
jeltsch: I’ve created the corresponding issue #263. I suggest that we leave the above code as it is for this pull request and have it fixed by a pull request associated with #263.

src/Database/LSMTree/Internal/IndexOrdinary.hs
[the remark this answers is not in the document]
jeltsch: An improved version of the documentation is in 7bb67533b2b4c2cf853aee387f20e930e4ef7b7b. Can you please check and tell me what you think about it?

src/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: Added in da5e3a80a45d590999e9e11c6f8c5de215b00319.

src/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: In any case, this is now a topic for #256.

src/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: I’ve filed issue #264 to deal with this.

jeltsch:
> Having at least one simple property test to give us some assurance that `search` is working would be nice to have […]. A little bit of duplication for now is not a bad thing

I’ve filed issue #265 to deal with this.

src/Database/LSMTree/Internal/IndexOrdinary.hs
jorisdral: I see, I was thinking about it wrong

jorisdral (approved):

src/Database/LSMTree/Internal/IndexOrdinary.hs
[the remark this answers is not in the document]
jeltsch: The corresponding change has been introduced in 378a844b737a6416ea5c789f646e939710007f04.

PR #248, opened 2024-06-11
Use `withTempHasFS` and `withTempHasBlockIO` in a few more places

jorisdral:


test/Test/Database/LSMTree/Internal/Merge.hs
```diff
@@ -37,15 +36,10 @@ tests = testGroup "Test.Database.LSMTree.Internal.Merge"
...
-    withSessionDir = Temp.withSystemTempDirectory "session-merge"
-
     -- TODO: run using mock file system once simulation is merged:
     -- https://github.com/input-output-hk/fs-sim/pull/48
     -- (also check all handles closed, see Test.Database.LSMTree.Internal.Run)
-    ioPropertyWithRealFS prop =
-        ioProperty $ withSessionDir $ \sessionRoot -> do
-          let mountPoint = FS.MountPoint sessionRoot
-          prop (FsIO.ioHasFS mountPoint)
+    ioPropertyWithRealFS prop = ioProperty $ withTempIOHasFS "session-merge" prop
```
mheinzel, line 42: We should switch this one to the mock `HasFS` soon anyway. So no need to touch it, but I'm okay with leaving the change in, now that it's there.

fs-api-blockio/src/System/FS/BlockIO/API.hs
```diff
@@ -44,7 +44,7 @@ data HasBlockIO m h = HasBlockIO {
     --
     -- Using 'submitIO' after 'close' should thrown an 'FsError' exception. See
     -- 'mkClosedError'.
-    close    :: HasCallStack => m ()
+    close    :: m ()
```
mheinzel, line 47: Out of interest, why did you remove the constraint?

mheinzel (approved):

fs-api-blockio/src/System/FS/BlockIO/API.hs
jorisdral: Because of `bracket`:

```
• Couldn't match type: HasCallStack => IO ()
                 with: IO c0
  Expected: HasBlockIO IO HandleIO -> IO c0
    Actual: HasBlockIO IO HandleIO -> HasCallStack => IO ()
• In the second argument of ‘bracket’, namely ‘close’
  In the expression: bracket (ioHasBlockIO hfs params) close action
  In an equation for ‘withIOHasBlockIO’:
      withIOHasBlockIO hfs params action
        = bracket (ioHasBlockIO hfs params) close actiontypecheck(-Wdeferred-type-errors)
```

But, we can keep the `HasCallStack` constraint because I've found a different implementation that works:

```diff
-withIOHasBlockIO hfs params action = bracket (ioHasBlockIO hfs params) close action
+withIOHasBlockIO hfs params action =
+    bracket (ioHasBlockIO hfs params) (\HasBlockIO{close} -> close) action
```

jorisdral: `(\hbio -> close hbio)` also works

mheinzel: Ah, interesting.

PR #251, opened 2024-06-13
Implement table updates by inserting into write buffer

mheinzel:
It might make sense to pull more parts of the table handle outside of the `RWVar` and sum type.

### DIFF
--- a/src/Database/LSMTree/Internal.hs
+++ b/src/Database/LSMTree/Internal.hs
@@ -320,7 +320,8 @@ closeSession Session{sessionState} =
 -- | A handle to an on-disk key\/value table.
 --
 -- For more information, see 'Database.LSMTree.Normal.TableHandle'.
-newtype TableHandle m h = TableHandle {
+data TableHandle m h = TableHandle {
+      tableConfig      :: !TableConfig
       -- | The primary purpose of this 'RWVar' is to ensure consistent views of
       -- the open-/closedness of a table when multiple threads require access to
       -- the table's fields (see 'withOpenTable'). We use more fine-grained
@@ -329,7 +330,7 @@ newtype TableHandle m h = TableHandle {
       -- TODO: RWVars only work in IO, not in IOSim.
       --
       -- TODO: how fair is an RWVar?
-      tableHandleState :: RWVar (TableHandleState m h)
+    , tableHandleState :: RWVar (TableHandleState m h)
     }
 
 -- | A table handle may assume that its corresponding session is still open as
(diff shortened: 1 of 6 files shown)


src/Database/LSMTree/Normal.hs
```diff
@@ -251,7 +254,15 @@ updates ::
   => V.Vector (k, Update v blob)
   -> TableHandle m k v blob
   -> m ()
-updates = undefined
+updates es (TableHandle th) = do
+    -- make sure not to insert any blobs into monoidal tables
+    when (isJust (Internal.confResolveMupsert (Internal.tableConfig th))) $
+      throwIO Internal.ErrExpectedNormalTable
+    Internal.updates Internal.constResolveMupsert (V.mapStrict serialiseEntry es) th
```
jorisdral, line 261: I think passing `Internal.constResolveMupsert` here is not necessary: we already checked that `Internal.confResolveMupsert` is `Nothing,` and `Internal.updates` can read the `confResolveMupsert` from the table handle using `resolveMupsert`

jorisdral (commented):
Looks good! There's only a minor thing I think we should change, but we can merge after that

> It might make sense to pull more parts of the table handle outside of the `RWVar` and sum type.

Let's do that on demand

src/Database/LSMTree/Normal.hs
mheinzel: Ah, yes. It's redundant.

src/Database/LSMTree/Internal.hs
```diff
@@ -530,13 +531,25 @@ toNormalLookupResult = \case
...
+updates es th = withOpenTable th $ \thEnv -> do
+    -- A placeholder implementation that is sufficient to pass the tests, but
+    -- keeps all entries in memory.
+    -- TODO: flush write buffer when full
+    -- TODO: merge runs when level becomes full
+    modifyMVar_ (tableWriteBuffer thEnv) $ \wb -> do
+      return $ foldl' (flip (uncurry (WB.addEntry resolve))) wb es
+  where
+    resolve =
+      maybe const unResolveMupsert (confResolveMupsert (tableConfig th))
```
jorisdral, line 552: You can use https://github.com/IntersectMBO/lsm-tree/pull/251/files#diff-b75c6514218a4064abdf1339bbf6a6e163f0be909f3d5cb7602dbfe6312f6023R660-R661 here

jorisdral (approved):

PR #254, opened 2024-06-17
Merge runs when levels become full

mheinzel:
Depends on #253.

The existing tests pass, but it would be good to add some new tests as well.


jorisdral:
We are going for a write buffer alloc method that is based on number of entries, not entry byte size. Superseded by #260

PR #260, opened 2024-06-20
Merge runs when levels become full

jorisdral:
Depends on #259 and #261

### DIFF
--- a/src/Database/LSMTree/Internal.hs
+++ b/src/Database/LSMTree/Internal.hs
@@ -60,6 +60,7 @@ import           Control.Concurrent.ReadWriteVar (RWVar)
 import qualified Control.Concurrent.ReadWriteVar as RW
 import           Control.Monad (unless, void, when)
 import           Control.Monad.Class.MonadThrow
+import           Control.Monad.ST.Strict (ST, runST)
 import           Data.Arena (newArenaManager)
 import           Data.Bifunctor (Bifunctor (..))
 import           Data.BloomFilter (Bloom)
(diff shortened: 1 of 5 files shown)


src/Database/LSMTree/Internal.hs
```diff
@@ -424,6 +424,23 @@ newtype Level h = Level {
...
 
+{-# SPECIALISE closeLevels :: HasFS IO h -> Levels (Handle h) -> IO () #-}
+closeLevels ::
+     m ~ IO
+  => HasFS m h
+  -> Levels (Handle h)
+  -> m ()
+closeLevels hfs levels = V.mapM_ (closeLevel hfs) levels
+
+-- | Runs in order from newer to older
```
jorisdral:
```suggestion
```

src/Database/LSMTree/Internal.hs
```diff
@@ -700,17 +706,163 @@ flushWriteBuffer hfs root uniqC reg tc
...
+
+    -- TODO: Until we implement proper scheduling, this does not only start a
+    -- merge, but it also steps it to completion.
+    newMerge conf' hfs' root' uniqC' mergepolicy mergelast rs
+      | Just (r, rest) <- V.uncons rs
+      , V.null rest = do
+          pure (SingleRun r)
+      | otherwise = do
+        assert (let l = V.length rs in l >= 2 && l <= 5) $ pure ()
+        r <- allocateTemp reg (mergeRuns conf' hfs' root' uniqC' mergepolicy mergelast rs) (Run.removeReference hfs')
```
jorisdral, line 786: This is very coarse grained masking: the whole merge is masked! If we want more coarse grained masking, then we'll have to thread the registry through merges and flushes

src/Database/LSMTree/Internal.hs
```diff
@@ -723,29 +705,261 @@ updatesWithInterleavedFlushes conf hfs root uniqC es reg tc = do
...
+    -- TODO: Until we implement proper scheduling, this does not only start a
+    -- merge, but it also steps it to completion.
+    newMerge :: MergePolicyForLevel -> Merge.Level -> V.Vector (Run (Handle h)) -> m (MergingRun (Handle h))
+    newMerge mergepolicy mergelast rs
+      | Just (r, rest) <- V.uncons rs
+      , V.null rest = do
+          pure (SingleRun r)
+      | otherwise = do
+        assert (let l = V.length rs in l >= 2 && l <= 5) $ pure ()
+        r <- allocateTemp reg (mergeRuns conf hfs root uniqC mergepolicy mergelast rs) (Run.removeReference hfs)
```
jorisdral, line 883: This is very coarse grained masking: the whole merge is masked! If we want more coarse grained masking, then we'll have to thread the registry through merges and flushes. However, it might not be worth it since we are going to schedule merges and then this code has to change anyway

src/Database/LSMTree/Internal/TempRegistry.hs
```diff
@@ -0,0 +1,202 @@
...
+-- | Like 'allocateTemp', but for resources that might fail to be acquired.
+allocateEitherTemp ::
+     (MonadMask m, MonadMVar m)
+  => TempRegistry m
+  -> m (Either e a)
+  -> (a -> m ())
+  -> m (Either e a)
+allocateEitherTemp reg acquire free =
+    mask_ $ do
+      eith <- acquire
```
mheinzel, line 129: Do we want the whole `acquire` to be masked? Is there a problem with using [`mask`](https://hackage.haskell.org/package/base-4.19.0.0/docs/Control-Exception.html#v:mask) and restoring around `acquire`?

src/Database/LSMTree/Internal.hs
```diff
@@ -723,29 +705,261 @@ updatesWithInterleavedFlushes conf hfs root uniqC es reg tc = do
...
+  -> HasFS m h
+  -> SessionRoot
+  -> UniqCounter m
+  -> Run (Handle h)
+  -> TempRegistry m
+  -> Levels (Handle h)
+  -> m (Levels (Handle h))
+addRunToLevels conf@TableConfig{..} hfs root uniqC r0 reg levels = do
+    ls' <- go 1 (V.singleton r0) levels
+    () <- stToIO $ levelsInvariant conf ls'
```
mheinzel, line 819: Oh, and checking the invariant needs to traverse the levels, even with asserts disabled. Since this is only done once per write buffer flush, it might be acceptable, but we could also make sure the whole traversal only happens on debug builds.

src/Database/LSMTree/Internal.hs
```diff
@@ -723,29 +705,261 @@ updatesWithInterleavedFlushes conf hfs root uniqC es reg tc = do
...
+-- the last level.
+--
+levelsInvariant :: forall s h. TableConfig -> Levels h -> ST s ()
+levelsInvariant conf = go 1
+  where
+    sr = confSizeRatio conf
+    wba = confWriteBufferAlloc conf
+
+    go :: Int -> Levels h -> ST s ()
+    go !_ (V.uncons -> Nothing)      = pure ()
```
mheinzel, line 742:
```suggestion
    go !_ (V.uncons -> Nothing) = pure ()
```

src/Database/LSMTree/Internal.hs
```diff
@@ -723,29 +705,261 @@ updatesWithInterleavedFlushes conf hfs root uniqC es reg tc = do
...
+
+    go !ln (V.uncons -> Just (Level mr rs, ls)) = do
+      mrs <- case mr of
+               SingleRun r                   -> pure $ CompletedMerge r
+               MergingRun (CompletedMerge r) -> pure $ CompletedMerge r
+      assert (length rs <= 3) $ pure ()
```
mheinzel, line 748: This also assumes a specific `SizeRatio`.

src/Database/LSMTree/Internal.hs
```diff
@@ -723,29 +705,261 @@ updatesWithInterleavedFlushes conf hfs root uniqC es reg tc = do
...
+    -- Check that a run is too small for next levels
+    fitsUB policy r ln = Run.runNumEntries r <= maxRunSize sr wba policy ln
+
+{-# SPECIALISE addRunToLevels :: TableConfig -> HasFS IO h -> SessionRoot -> UniqCounter IO -> Run (Handle h) -> TempRegistry IO -> Levels (Handle h) -> IO (Levels (Handle h)) #-}
+addRunToLevels ::
+     forall m h. m ~ IO -- TODO: replace by @io-classes@ constraints for IO simulation.
+  => TableConfig
+  -> HasFS m h
+  -> SessionRoot
+  -> UniqCounter m
```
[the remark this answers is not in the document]
jorisdral: I don't have a strong preference

src/Database/LSMTree/Internal.hs
```diff
@@ -415,13 +425,44 @@ withOpenTable th action = RW.with (tableHandleState th) $ \case
...
   }
-  deriving newtype NoThunks
+
+-- TODO: proper instance
+deriving via OnlyCheckWhnfNamed "Level" (Level h) instance NoThunks (Level h)
+
+-- | A merging run is either a single run, or some ongoing merge.
+data MergingRun h =
+    MergingRun !(MergingRunState h)
+  | SingleRun !(Run h)
```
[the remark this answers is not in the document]
jorisdral: I'll remove it for now

src/Database/LSMTree/Internal/TempRegistry.hs
mheinzel: Okay, resolved.

src/Database/LSMTree/Internal.hs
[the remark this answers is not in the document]
jorisdral: On second thought, I think I might keep it. The representation isn't strictly necessary, but it means `levelsInvariant` is a slightly stronger invariant. In particular, `SingleRun`s must be of *exactly* the right size, while a `MergingRun` only has an upper bound

src/Database/LSMTree/Internal.hs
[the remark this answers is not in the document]
jorisdral: I think an advantage of this approach is that the function is only passed what it will use, and not other stateful parts that it's not going to touch. In any case, I'll leave it like this for now, and we can change it in a follow-up if we want to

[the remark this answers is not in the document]
mheinzel: That would also work. Either way, IMO it's okay for now.

src/Database/LSMTree/Internal.hs
```diff
@@ -449,7 +490,12 @@ mkLevelsCache lvls = LevelsCache_ {
     , cachedIndexes   = V.map Run.runIndex rs
     , cachedKOpsFiles = V.map Run.runKOpsFile rs
     }
-  where rs = V.concatMap residentRuns lvls
+  where
+    rs = flip V.concatMap lvls $ \(Level mr rs') ->
+      V.cons (case mr of
+                SingleRun r                   -> r
+                MergingRun (CompletedMerge r) -> r)
+             rs'
```
mheinzel, line 498: Hmm, maybe instead of providing `closeLevels` above, it would have been nicer to provide `runsInLevels` and the we could use it here and also do `for_ (runsInLevels lvls) $ Run.removeReference (tableHasFS thEnv)` to close them.

But not too important, just a little bit of code duplication.

mheinzel (approved):

src/Database/LSMTree/Internal.hs
```diff
@@ -445,25 +445,20 @@ data MergingRun h =
 -- TODO: this should also represent ongoing merges once we implement scheduling.
 newtype MergingRunState h = CompletedMerge (Run h)
 
+-- | Return all runs in the levels, ordered from newest to oldest
+runsInLevels :: Levels h -> V.Vector (Run h)
+runsInLevels levels = flip V.concatMap levels $ \(Level mr rs) ->
+    case mr of
+      SingleRun r                   -> r `V.cons` rs
+      MergingRun (CompletedMerge r) -> r `V.cons` rs
```
dcoutts, line 453: This could probably be more efficient. Especially when used for `closeLevels`. It might be better to use a traversal function rather than allocating a new vector.

dcoutts (commented):
LGTM. This is looking good.

I like the partial adoption of the structure of the prototype.

src/Database/LSMTree/Internal.hs
mheinzel: Ah! When suggesting to add this function I had in mind that it would return a list, which I would expect to be optimised away when used with either `V.fromList` (for the cache) or `for_` (to close them).

PR #269, opened 2024-07-02
Put ArenaManager into TableHandle

phadej:


dcoutts (approved):
LGTM!

src/Database/LSMTree/Internal.hs
```diff
@@ -417,11 +419,11 @@ data TableContent h = TableContent {
 withOpenTable ::
      m ~ IO -- TODO: replace by @io-classes@ constraints for IO simulation.
   => TableHandle m h
-  -> (TableHandleEnv m h -> m a)
+  -> (ArenaManager (PrimState m) -> TableHandleEnv m h -> m a)
```
mheinzel, line 422: I would argue that `withOpenTable` should only provide the `TableHandleEnv` here, as the arena manager can simply be grabbed from the table handle (like we already do for the table config). That also avoids having to introduce unused `_arenaManager` arguments.

Does that sound reasonable?

mheinzel: Related, but totally optional for now: Do we ultimately want to share one arena manager per session? If so, I'd even lean towards putting it inside the `TableHandleEnv`, next to the other inherited fields, for consistency. But no strong opinion and we can still do that once we start sharing.

mheinzel (changes requested):
Looks good, just one change I'd like to see.

src/Database/LSMTree/Internal.hs
phadej: @mheinzel see https://github.com/IntersectMBO/lsm-tree/pull/258#discussion_r1661285454 comment

mheinzel: Ah, yes, just saw that. Then ignore my second comment.

phadej: changed `withOpenTable`

mheinzel (approved):

PR #275, opened 2024-07-03
Fix level calculations for tiny runs in the ScheduledMerges prototype

### LINKED ISSUE #236: Assertion failure in `ScheduledMerges` prototype
https://github.com/IntersectMBO/lsm-tree/actions/runs/9351899757/job/25738763256?pr=235

mheinzel:
Fixes #236 

I'm just pushing the regression test first so you can see the test failure in CI.

### DIFF
--- a/prototypes/ScheduledMerges.hs
+++ b/prototypes/ScheduledMerges.hs
@@ -137,13 +137,16 @@ levellingRunSize :: Int -> Int
 levellingRunSize n = 4^(n+1)
 
 tieringRunSizeToLevel :: Run -> Int
-tieringRunSizeToLevel r =
+tieringRunSizeToLevel r
+  | s <= bufferSize = 1  -- level numbers start at 1
+  | otherwise =
     1 + (finiteBitSize s - countLeadingZeros (s-1) - 1) `div` 2
   where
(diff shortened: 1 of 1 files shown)


prototypes/ScheduledMerges.hs
```diff
@@ -416,6 +420,14 @@ update tr (LSMHandle scr lsmr) k op = do
       else
         writeSTRef lsmr (LSMContent wb' ls)
 
+supply :: LSM s -> Credit -> ST s ()
+supply (LSMHandle scr lsmr) credits = do
+    LSMContent _ ls <- readSTRef lsmr
+    modifySTRef' scr (+1)
+    supplyCredits credits ls
+    ok <- invariant ls
+    assert ok $ return ()
```
mheinzel, line 429: I exposed this to make the test case simpler, just supplying merge credits instead of having to do inserts that are not relevant to the problem.

mheinzel (commented):
The regression test and ability to show the tracer log with a failure are not strictly necessary, but I created them to track down the bug and thought it doesn't hurt to commit them.

jorisdral (approved):
LGTM! Nice catch

PR #281, opened 2024-07-05
Add lockstep tests for MergeReaders

mheinzel:
As discussed in https://github.com/IntersectMBO/lsm-tree/pull/215#discussion_r1608330935.

Perhaps not the ideal use case for quickcheck-lockstep, e.g. not making use of variables, but probably better than not having tests. Will help with ensuring the correctness of future optimisations.

### DIFF
--- a/test/Test/Database/LSMTree/Internal/RunReaders.hs
+++ b/test/Test/Database/LSMTree/Internal/RunReaders.hs
@@ -1,0 +1,351 @@
+{-# LANGUAGE LambdaCase   #-}
+{-# LANGUAGE TypeFamilies #-}
+
+module Test.Database.LSMTree.Internal.RunReaders (tests) where
+
+import           Control.Exception (assert)
+import           Control.Monad (zipWithM)
+import           Control.Monad.IO.Class (liftIO)
+import           Control.Monad.Trans.Reader (ReaderT (..))
+import           Control.Monad.Trans.State (StateT (..), get, put)
+import           Data.Bifunctor (bimap, first)
+import           Data.Foldable (traverse_)
+import qualified Data.Map.Strict as Map
+import           Data.Proxy (Proxy (..))
+import           Data.Word (Word64)
+import           Database.LSMTree.Extras (showPowersOf)
+import           Database.LSMTree.Extras.Generators (KeyForIndexCompact,
+                     TypedWriteBuffer (..))
+import           Database.LSMTree.Internal.BlobRef
+import           Database.LSMTree.Internal.Entry
+import qualified Database.LSMTree.Internal.Paths as Paths
+import qualified Database.LSMTree.Internal.Run as Run
+import qualified Database.LSMTree.Internal.RunReader as Reader
+import           Database.LSMTree.Internal.RunReaders
+                     (HasMore (Drained, HasMore), Readers)
+import qualified Database.LSMTree.Internal.RunReaders as Readers
+import           Database.LSMTree.Internal.Serialise
+import qualified Database.LSMTree.Internal.WriteBuffer as WB
+import qualified System.FS.API as FS
+import qualified System.FS.Sim.MockFS as MockFS
+import qualified System.FS.Sim.STM as FsSim
+import qualified Test.QuickCheck as QC
+import           Test.Tasty (TestTree, testGroup)
+import           Test.Tasty.QuickCheck
+import           Test.Util.Orphans ()
+
+import           Test.QuickCheck.StateModel
+import           Test.QuickCheck.StateModel.Lockstep
+import qualified Test.QuickCheck.StateModel.Lockstep.Defaults as Lockstep
+import qualified Test.QuickCheck.StateModel.Lockstep.Run as Lockstep
+
+tests :: TestTree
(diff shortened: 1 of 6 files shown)


test/Test/Database/LSMTree/Internal/RunReaders.hs
```diff
@@ -0,0 +1,338 @@
...
+    -> [Any (LockstepAction ReadersState)]
+  shrinkWithVars _ _ = \case
+      New wbs -> Some . New <$> shrink wbs
+      Pop n   -> Some . Pop <$> shrink n
+      _       -> []
+
+  -- TODO: tag, e.g.
+  -- - fully draining the Readers
+  -- - DropWhileKey dropping multiple entries
+  tagStep _ _ _ = []
```
jorisdral, line 215: I think it's important to add this in a follow-up PR, because otherwise I'm not sure we're exercising some scenarios, like draining readers or `DropWhileKey` dropping anything

test/Test/Database/LSMTree/Internal/RunReaders.hs
```diff
@@ -0,0 +1,338 @@
...
+data Op a b where
+  OpId   :: Op a a
+  OpUnit :: Op a ()
+
+instance Operation Op where
+  opIdentity = OpId
+
+instance InterpretOp Op (ModelValue ReadersState) where
+  intOp OpId   x = Just x
+  intOp OpUnit _ = Just (MUnit ())
```
jorisdral, line 247: My intuition says that this isn't needed, and you can use the existing [SumProd](https://hackage.haskell.org/package/quickcheck-lockstep-0.5.0/docs/Test-QuickCheck-StateModel-Lockstep-Op-SumProd.html) instead

mheinzel: In fact I wasn't using the type at all, just deleted it. :smile:

test/Test/Database/LSMTree/Internal/RunReaders.hs
mheinzel: I added it now, should cover the most interesting cases.

```
    prop_lockstep: OK (8.61s)
      +++ OK, passed 100 tests.
      
      Action polarity (3036 in total):
      100.00% +
      
      Actions (3036 in total):
      45.32% +Pop
      28.89% +DropWhileKey
      20.95% +PeekKey
       4.84% +New
      
      Tags (1772 in total):
      32.67% NewEntriesKeyDuplicates 2 <= n < 4
      28.95% DropWhileKeyDropped 1 <= n < 2
      13.49% DropWhileKeyDropped n == 0
       6.32% NewEntriesKeyDuplicates 4 <= n < 8
       5.47% DropWhileKeyDropped 2 <= n < 4
       5.14% NewEntries 100 <= n < 1000
       3.22% ReadersFullyDrained
       2.99% NewEntries 10 <= n < 100
       1.58% DropWhileKeyDropped 4 <= n < 8
       0.17% NewEntries 1 <= n < 10
```

mheinzel: I could tweak it to drain the readers more frequently, but at least it happens.

jorisdral: One action you could add is something like `Drain`, which pops until drained. Regardless, this tagging looks like a step in the right direction!

jorisdral (approved):

test/Test/Database/LSMTree/Internal/RunReaders.hs
mheinzel: I wanted to make sure the last few entries are still checked properly (including looking at the results), as there could be edge cases there.

I guess for `Pop n`, which is currently generated with `n` uniformly between 1 and the number of entries, we could have a special case generating `n` just slightly less than the number of entries. I just tried that, brought the number of `ReadersFullyDrained` from 3% to 7% (or ~47 to ~120 in total numbers), but it also increases the test run time from 7s to 13s, as we are running through a lot more runs that get created.

mheinzel: I increased the frequency of all the other actions a little, might be a reasonable tradeoff:

```
    prop_lockstep: OK (9.86s)
      +++ OK, passed 100 tests.
      
      Action polarity (2344 in total):
      100.00% +
      
      Actions (2344 in total):
      44.37% +Pop
      24.23% +PeekKey
      23.85% +DropWhileKey
       7.55% +New
      
      Tags (1536 in total):
      38.54% NewEntriesKeyDuplicates 2 <= n < 4
      23.31% DropWhileKeyDropped 1 <= n < 2
       7.62% NewEntriesKeyDuplicates 4 <= n < 8
       6.71% NewEntries 100 <= n < 1000
       6.64% DropWhileKeyDropped n == 0
       5.73% ReadersFullyDrained
       4.62% DropWhileKeyDropped 2 <= n < 4
       4.36% NewEntries 10 <= n < 100
       1.82% DropWhileKeyDropped 4 <= n < 8
       0.46% NewEntries 1 <= n < 10
       0.20% NewEntriesKeyDuplicates 8 <= n < 16
```

PR #293, opened 2024-07-12
Implement arena allocator

phadej:


phadej:
```
Benchmark                                                                                                             main      arena           
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 negative lookups, NoCache/Bloomfilter query             0.788e-5  0.785e-5  -0.35%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 negative lookups, NoCache/Compact index search          1.919e-6  1.873e-6  -2.38%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 negative lookups, NoCache/Lookup preparation in memory  0.990e-5  0.984e-5  -0.63%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 negative lookups, NoCache/Lookups in IO                 1.247e-3  0.126e-3 -89.90%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 negative lookups, NoCache/Perform intra-page lookups    2.829e-6  2.697e-6  -4.68%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 negative lookups, NoCache/Submit IOOps                  1.060e-4  1.066e-4  +0.56%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 negative lookups/Bloomfilter query                      0.792e-5  0.775e-5  -2.19%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 negative lookups/Compact index search                   1.895e-6  1.872e-6  -1.17%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 negative lookups/Lookup preparation in memory           0.979e-5  0.977e-5  -0.16%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 negative lookups/Lookups in IO                          1.158e-3  0.029e-3 -97.46%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 negative lookups/Perform intra-page lookups             1.483e-6  1.553e-6  +4.71%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 negative lookups/Submit IOOps                           2.220e-5  2.250e-5  +1.35%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 positive lookups, NoCache/Bloomfilter query             1.568e-5  1.559e-5  -0.58%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 positive lookups, NoCache/Compact index search          1.952e-4  0.964e-4 -50.63%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 positive lookups, NoCache/Lookup preparation in memory  2.495e-4  1.131e-4 -54.66%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 positive lookups, NoCache/Lookups in IO                 0.394e-2  0.201e-2 -48.86%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 positive lookups, NoCache/Perform intra-page lookups    1.414e-4  1.083e-4 -23.42%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 positive lookups, NoCache/Submit IOOps                  1.831e-3  1.796e-3  -1.92%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 positive lookups/Bloomfilter query                      1.576e-5  1.577e-5  +0.10%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 positive lookups/Compact index search                   1.970e-4  0.978e-4 -50.36%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 positive lookups/Lookup preparation in memory           2.501e-4  1.134e-4 -54.67%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 positive lookups/Lookups in IO                          2.135e-3  0.555e-3 -73.98%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 positive lookups/Perform intra-page lookups             0.536e-4  0.380e-4 -29.16%
Bench.Database.LSMTree.Internal.Lookup/2_000_000 entries, 256 positive lookups/Submit IOOps                           0.381e-3  0.380e-3  -0.27%
Geometric mean                                                                                                        0.614e-4  0.378e-4 -38.49%
```

cabal.project
```diff
@@ -30,7 +30,7 @@ benchmarks: True
 constraints: bloomfilter <0
 
 -- comment me if you are benchmarking
-import: cabal.project.debug
+-- import: cabal.project.debug
```
mheinzel, line 33: This should not be committed.

src/Data/Arena.hs
```diff
@@ -15,74 +17,169 @@ module Data.Arena (
...
 instance NFData (Arena s) where
-  rnf (Arena mvar) = rwhnf mvar
+  rnf (Arena !_ !_ !_) = ()
 
 type Size      = Int
 type Offset    = Int
 type Alignment = Int
 
+blockSize :: Int
+blockSize = 0x100000
```
mheinzel, line 60: Might be worth tuning, but seems like a good default for now. :+1:

src/Data/Arena.hs
```diff
@@ -15,74 +17,169 @@ module Data.Arena (
...
 
+-- | Allocate a slice of mutable byte array from the arena.
 allocateFromArena :: PrimMonad m => Arena (PrimState m)-> Size -> Alignment -> m (Offset, MutableByteArray (PrimState m))
-allocateFromArena (Arena mvar) !size !alignment = do
-    mba <- newAlignedPinnedByteArray size alignment
-    atomicModifyMutVar' mvar $ \mbas -> (mba : mbas, ())
-    return (0, mba)
+allocateFromArena !arena !size !alignment =
+    assert (popCount alignment == 1) $ -- powers of 2
+    assert (size <= blockSize) $ -- not to large allocations
```
mheinzel, line 142:
```suggestion
    assert (size <= blockSize) $ -- not too large allocations
```

src/Data/Arena.hs
```diff
@@ -15,74 +17,169 @@ module Data.Arena (
...
+    allocateFromArena' arena size alignment
+
+-- TODO!? this is not async exception safe
+allocateFromArena' :: PrimMonad m => Arena (PrimState m)-> Size -> Alignment -> m (Offset, MutableByteArray (PrimState m))
```
mheinzel, line 146: Might not be too bad, but tell me if I misunderstood something. :thinking: 

If it fails before putting back the MVar, the arena becomes unusable. Future calls will deadlock, but at least not silently corrupt things.
And AFAIU the way we use the arena for lookups, that should be fine, since the exception will bubble up (without a call to `closeArena`) and might only be caught outside of `lookup`. Retrying there will give you a new arena. The unusable one should at some point get garbage collected? So it's a bit wasteful, but only in exceptional situations.

Maybe that's good enough and we just want to document it?

src/Data/Arena.hs
```diff
@@ -15,74 +17,169 @@ module Data.Arena (
...
+            return Arena {..}
+
+closeArena :: PrimMonad m => ArenaManager (PrimState m) -> Arena (PrimState m) -> m ()
+closeArena (ArenaManager arenas) arena = do
+    scrambleArena arena
+
+    -- reset the arena to clear state
+    resetArena arena
+
+    atomicModifyMutVar' arenas $ \xs -> (arena : xs, ())
```
mheinzel, line 96: In debug mode, we could even not put back the arena, always handing out fresh arenas. This way, the next user can't overwrite the scrambled blocks again with something that looks like sensible data. On the other hand, this would mean that `newBlockWithFree` doesn't get good test coverage. What do you think?

mheinzel (commented):
Looks good to me, but async exception safety is an interesting point, even if just to document the current assumptions/guarantees.

src/Data/Arena.hs
phadej: > What do you think?

Exactly, as you

> On the other hand, this would mean that newBlockWithFree doesn't get good test coverage. 

---

I think that if new stuff overwrites something old, some test should fail without scrambling too. So  I chose the approach which is most similar between debug and non-debug mode.

src/Data/Arena.hs
phadej: It really depends whether the code is run in IO or ST. Your analysis is probably true for ST usecases, for IO it's not.

src/Data/Arena.hs
phadej:
```
Prelude Numeric> showHex (4096 * 256) ""
"100000"
```

So it's close. For better test coverage it should be a lot less, for real use case it might sense to double.

src/Data/Arena.hs
mheinzel: Makes sense!

src/Data/Arena.hs
mheinzel: Ah, interesting. What might go wrong in IO?

phadej: there is `bracket`, so one could actually recover. (This is why I don't like having `newArena/closeArena` exported, I see that as a deficit in benchmarking library, that it doesn't support `with*` pattern resource allocation).

mheinzel: Do you mean using `bracket _ closeArena` in `withArena`? This would require us to also use `bracket`/`withMVar` within `allocateFromArena`, too, right (if we don't put back the MVar, `closeArena` will hang)? But then we have to be careful with the state of the other vars. E.g. we put back `curr'`, but we had already added it to `full` as well. Or we got something from `free`, but end up just dropping it. Should be possible, but not trivial.

phadej: @mheinzel I do think that *not* using `bracket` in `withArena` makes stuff simpler. In case of exception an invalid state arena gets GCd. Current interface is suboptimal.

Getting `allocateFromArena` exception safe might not be easy (I'd really like to use TVars, but those are not possible in ST - or then one needs another abstraction, TVars for IO and ordinary vars for ST).

mheinzel: Okay, thanks. Even with an `ST` interface, someone could do something like `bracket (stToIO newArena) (stToIO . closeArena)`. And it indeed seems complicated to make `allocateFromArena` exception safe. So maybe a reasonable way forward would be to make sure non-test code always used `withArena` and document the situation:

- export `newArena` and `closeArena` under the "Test helpers" section, document that the arena won't be reused, so non-test code should use `withArena`, maybe rename them to make it really clear
- document `allocateFromArena` to say that async exceptions will leave the arena unusable, so it is suggested to let them bubble up to escape `withArena`

mheinzel: For `closeArena`, it's only used in the cleanup of the lookup micro benchmark where it could even lead to a deadlock I think?

```haskell
          perRunEnvWithCleanup
            ( newArena arenaManager >>= \arena ->
              stToIO (prepLookups arena blooms indexes kopsFiles ks) >>= \(rkixs, ioops) ->
              FS.submitIO hasBlockIO ioops >>= \ioress ->
              pure (rkixs, ioops, ioress, arena)
            )
            (\(_, _, _, arena) -> closeArena arena) $ \ ~(rkixs, ioops, ioress, _) -> do
              !_ <- intraPageLookups resolveV rs ks rkixs ioops ioress
              pure ()
```

I guess we want to reuse the arena in the happy path here. But we could close the arena below `intraPageLookups`, so it's skipped on exception.

jorisdral: I think there is value in exposing `newArena` and `closeArena`, even if it is only for unsafe uses. We can prefix them with `unsafe` instead.

Regarding `closeArena` hanging, maybe we should then have an `unsafeTryCloseArena`

phadej: So should I change something here, or can we merge this?

jorisdral (approved):

PR #296, opened 2024-07-15
Add general-purpose chunk handling

jeltsch:
This adds a general-purpose module for dealing with chunks, which comprises a chunk type and support for chunk production suitable for incremental index serialization.

Note that this pull request does not include tests. These are called for by #297.

### DIFF
--- a/src/Database/LSMTree/Internal/Chunk.hs
+++ b/src/Database/LSMTree/Internal/Chunk.hs
@@ -1,0 +1,103 @@
+{- HLINT ignore "Avoid restricted alias" -}
+
+module Database.LSMTree.Internal.Chunk
+(
+    -- * Chunks
+    Chunk (Chunk),
+    fromChunk,
+
+    -- * Balers
+    Baler,
+    createBaler,
+    feedBaler,
+    unsafeEndBaler
+)
+where
+
+import           Prelude hiding (length)
+
+import           Control.Exception (assert)
+import           Control.Monad.ST.Strict (ST)
+import           Data.STRef.Strict (STRef, newSTRef, readSTRef, writeSTRef)
+import           Data.Vector.Primitive (Vector, length, unsafeCopy,
+                     unsafeFreeze)
+import           Data.Vector.Primitive.Mutable (MVector)
+import qualified Data.Vector.Primitive.Mutable as Mutable (drop, length, slice,
+                     take, unsafeCopy, unsafeNew)
+import           Data.Word (Word8)
+
+-- * Chunks
+
+-- | A chunk of bytes, typically output during incremental index serialisation.
+newtype Chunk = Chunk (Vector Word8)
+
+fromChunk :: Chunk -> Vector Word8
+fromChunk (Chunk content) = content
+
+-- * Balers
+
+{-|
+    An object that receives blocks of bytes and repackages them into chunks such
+    that all chunks except for a possible remnant chunk at the end are of at
+    least a given minimum size.
+-}
+data Baler s = Baler
+                   !(MVector s Word8) -- Buffer storing queued bytes
+                   !(STRef s Int)     -- Reference to the number of queued bytes
+
+-- | Creates a new baler.
+createBaler :: Int            -- ^ Minimum chunk size in bytes
+            -> ST s (Baler s) -- ^ Creation of the baler
+createBaler minChunkSize = assert (minChunkSize > 0)              $
+                           Baler                                 <$>
+                           Mutable.unsafeNew (pred minChunkSize) <*>
+                           newSTRef 0
+
+{-|
+    Feeds a baler a block of bytes.
+
+    Bytes received by a baler are generally queued for later output, but if
+    feeding new bytes makes the accumulated content exceed the minimum chunk
+    size then a chunk containing all the accumulated content is output.
+-}
+feedBaler :: Vector Word8 -> Baler s -> ST s (Maybe Chunk)
+feedBaler block (Baler buffer remnantSizeRef) = do
+    remnantSize <- readSTRef remnantSizeRef
+    let
+
+        totalSize :: Int
+        !totalSize = remnantSize + length block
+
+    if totalSize <= Mutable.length buffer
+        then do
+                 unsafeCopy (Mutable.slice remnantSize (length block) buffer)
+                            block
+                 writeSTRef remnantSizeRef totalSize
+                 return Nothing
+        else do
+                 protoChunk <- Mutable.unsafeNew totalSize
+                 Mutable.unsafeCopy (Mutable.take remnantSize protoChunk)
+                                    (Mutable.take remnantSize buffer)
+                 unsafeCopy (Mutable.drop remnantSize protoChunk) block
+                 writeSTRef remnantSizeRef 0
+                 chunk <- Chunk <$> unsafeFreeze protoChunk
+                 return (Just chunk)
+
+{-|
+    Returns the bytes still queued in a baler, if any, thereby invalidating the
+    baler. Executing @unsafeEndBaler baler@ is only safe when @baler@ is not
+    used afterwards.
+-}
+unsafeEndBaler :: forall s . Baler s -> ST s (Maybe Chunk)
+unsafeEndBaler (Baler buffer remnantSizeRef) = do
+    remnantSize <- readSTRef remnantSizeRef
+    if remnantSize == 0
+        then return Nothing
+        else do
+                 let
+
+                     protoChunk :: MVector s Word8
+                     !protoChunk = Mutable.take remnantSize buffer
+
+                 chunk <- Chunk <$> unsafeFreeze protoChunk
+                 return (Just chunk)
--- a/lsm-tree.cabal
+++ b/lsm-tree.cabal
@@ -60,6 +60,7 @@ library
     Database.LSMTree.Internal.BlobRef
     Database.LSMTree.Internal.BloomFilter
     Database.LSMTree.Internal.ByteString
+    Database.LSMTree.Internal.Chunk
     Database.LSMTree.Internal.CRC32C
     Database.LSMTree.Internal.Entry
     Database.LSMTree.Internal.IndexCompact


src/Database/LSMTree/Internal/Chunk.hs
```diff
@@ -0,0 +1,98 @@
...
+-- * Chunks
+
+-- | A chunk of bytes, typically output during incremental index serialisation.
+newtype Chunk = Chunk (Vector Word8)
+
+-- * Balers
+
+-- Mutable state of a baler.
+data BalerState = BalerState
+                      ![Vector Word8] -- Reverse list of queued blocks
```
jorisdral, line 30:
```suggestion
                      [Vector Word8] -- Reverse list of queued blocks
```

src/Database/LSMTree/Internal/Chunk.hs
```diff
@@ -0,0 +1,98 @@
...
+    BalerState remnantBlocksRev remnantSize <- readSTRef stateRef
+    let
+
+        totalSize :: Int
+        totalSize = remnantSize + length block
+
+    if totalSize < maxChunkSize
+        then do
+                 writeSTRef stateRef
+                            (BalerState (block : remnantBlocksRev) totalSize)
```
jorisdral, line 64:
```suggestion
                 writeSTRef stateRef $!
                            (BalerState (block : remnantBlocksRev) totalSize)
```

src/Database/LSMTree/Internal/Chunk.hs
```diff
@@ -0,0 +1,98 @@
...
+                 return []
+        else do
+                 let
+
+                     newRemnantSize :: Int
+                     newRemnantSize = totalSize `mod` maxChunkSize
```
jorisdral, line 70:
```suggestion
                     !newRemnantSize = totalSize `mod` maxChunkSize
```

src/Database/LSMTree/Internal/Chunk.hs
```diff
@@ -0,0 +1,98 @@
...
+{-|
+    An object that receives blocks of bytes and repackages them into chunks of
+    preferably a given maximum size.
+-}
+data Baler s = Baler
+                   !Int                  -- Maximum chunk size
+                   !(STRef s BalerState) -- Reference to the mutable state
+
+-- | Creates a new baler.
+createBaler :: Int            -- ^ Maximum chunk size
```
jorisdral, line 42:
```suggestion
createBaler :: Int            -- ^ Maximum chunk size in bytes
```

src/Database/LSMTree/Internal/Chunk.hs
```diff
@@ -0,0 +1,98 @@
...
+
+                     output :: Vector Word8
+                     output = concat (reverse (lastOutput : remnantBlocksRev))
+
+                     chunks :: [Chunk]
+                     chunks = Chunk                                        <$>
+                              (\ start -> slice start maxChunkSize output) <$>
+                              init [0, maxChunkSize .. totalSize]
+
+                 writeSTRef stateRef (BalerState [newRemnant] newRemnantSize)
```
jorisdral, line 84:
```suggestion
                 writeSTRef $! stateRef (BalerState [newRemnant] newRemnantSize)
```

src/Database/LSMTree/Internal/Chunk.hs
```diff
@@ -0,0 +1,98 @@
...
+                 return chunks
+
+{-|
+    Reads the bytes queued in a bailer.
+
+    Note that this will not dispose of the bytes in the bailer. To actually get
+    rid of them, you have to get rid of the bailer itself.
+-}
+readBalerRemnant :: Baler s -> ST s (Maybe Chunk)
+readBalerRemnant (Baler _ stateRef) = do
```
jorisdral, line 94: What is this function going to be used for?

src/Database/LSMTree/Internal/Chunk.hs
```diff
@@ -0,0 +1,98 @@
...
+                      ![Vector Word8] -- Reverse list of queued blocks
+                      !Int            -- Total size of queued blocks
+
+{-|
+    An object that receives blocks of bytes and repackages them into chunks of
+    preferably a given maximum size.
+-}
+data Baler s = Baler
+                   !Int                  -- Maximum chunk size
+                   !(STRef s BalerState) -- Reference to the mutable state
```
jorisdral, line 39: Seeing as the maximum chunk size is already known, you could pre-allocate a mutable vector of the appropriate size and copy the blocks into it. I think that might be a nicer alternative to the block queue

src/Database/LSMTree/Internal/Chunk.hs
```diff
@@ -0,0 +1,98 @@
...
+                     (lastOutput, newRemnant)
+                         = splitAt (length block - newRemnantSize) block
+
+                     output :: Vector Word8
+                     output = concat (reverse (lastOutput : remnantBlocksRev))
+
+                     chunks :: [Chunk]
+                     chunks = Chunk                                        <$>
+                              (\ start -> slice start maxChunkSize output) <$>
+                              init [0, maxChunkSize .. totalSize]
```
jorisdral, line 82: Maybe it would be fine to return a larger chunk if the current block would fill up multiple chunks. I know the compact index would return multiple chunks too, but I wonder if it is necessary

src/Database/LSMTree/Internal/Chunk.hs
```diff
@@ -0,0 +1,98 @@
...
+
+{-|
+    Feeds a baler a block of bytes.
+
+    Bytes received by a bailer are generally queued for later output, but if
+    feeding new bytes makes the queued content exceed the maximum chunk size
+    then as many chunks of maximum size as possible are output.
+-}
+feedBaler :: Vector Word8 -> Baler s -> ST s [Chunk]
+feedBaler block (Baler maxChunkSize stateRef) = do
```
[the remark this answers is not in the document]
jeltsch: Either a serialized key or the bytes representing the size of a serialized key.

src/Database/LSMTree/Internal/Chunk.hs
```diff
@@ -0,0 +1,98 @@
...
+    BalerState remnantBlocksRev remnantSize <- readSTRef stateRef
+    let
+
+        totalSize :: Int
+        totalSize = remnantSize + length block
```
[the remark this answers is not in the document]
jeltsch: Does GHC really not figure out that `totalSize` will definitely be used? After all, `totalSize` is compared to `maxChunkSize` in the condition of the `if` expression that immediately follows. I’m not at all against ensuring that variables like `totalSize` get evaluated promptly, but I’d like to avoid spreading `!`-annotations beyond necessity to ensure that the code stays as clear as possible.

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: Now it was on me to add the `!`. 😉 My motivation was that making this field strict is sensible already from a purely denotational standpoint (so not even considering evaluation, unboxing, and such): this field becoming ⊥ wouldn’t make sense. Actually, it would be even more sensible to make the whole list strict, but this would unnecessarily complicate the code. What would be your motivation for removing the `!`?

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: Wouldn’t this use of `($!)` be superfluous? The expression of what to put into the mutable variable is an application of a data constructor and therefore in head normal form. No thunk should be created; the value should be in evaluated form from the outset.

jeltsch: Hmm, on second thought I’m not so sure anymore. The point is that `BalerState` has strict fields and therefore is not an ordinary data constructor. In fact, it should effectively be defined internally as follows, where `BalerState'` denotes the actual, underlying data constructor:
```
BalerState remnantBlocksRev remnantSize = remnantBlocksRev `seq` remnantSize `seq` BalerState' remnantBlocksRev remnantSize
```
Ironically, this would make the above `writeSTRef` application less strict in a sense, as its second argument wouldn’t be a bare data constructor application anymore. Is this how it is?

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: I was inclined to write the same as I wrote [above](https://github.com/IntersectMBO/lsm-tree/pull/296#discussion_r1680779780) regarding `totalSize`, but then I figured that it is much harder to convince yourself that `newRemnantSize` will be evaluated promptly (in fact, it might not even be evaluated promptly without adding the `$!` you suggested to add [above](https://github.com/IntersectMBO/lsm-tree/pull/296#discussion_r1680797032)). Maybe it is indeed better to be more generous with `!`s and make intended strictness more explicit.

However, shouldn’t then some other patterns get `!`s added as well, for example `output`? What is actually the general policy regarding what should be strict and what not? I guess that things stored in mutable variables should usually contain no thunks, as such thunks would stay beyond the execution of the `ST` (or `IO`) action that created them with no chance for outside code to force their evaluation. But what about things like the list of chunks that `feedBaler` creates? Should this be lazy? I’m inclined to say “yes”, but I guess that it would be a good idea to make sure that the underlying `output` vector has been evaluated by the time the execution of `feedBaler` ends. What is your take on these issues, and is there a general concept that is followed throughout the `lsm-tree` package?

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: I deliberately left out this bit, because the documentation of `Chunk` already says that the chunks we’re talking about here are chunks of bytes. However, it might be good to be a bit more explicit. I’ll make this change and also extend the comment in the definition of `Baler` in the same way.

jeltsch: Done: 50c287f4531529a921d7e7a22cc0dfd3d228b7da.

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: I guess the discussion of [the other suggestion to add `$!`](https://github.com/IntersectMBO/lsm-tree/pull/296#discussion_r1680797032) can be applied here as well, with the additional note that here you placed `$!` right after `writeSTRef`, which was probably not intended 🙂.

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: It’s for implementing ending of the ordinary index (and potentially similar things in potential other applications of this module).

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: I don’t think that this alternative would be nicer. It would certainly use a more imperative programming style. Furthermore, it would involve a data structure (the chunk buffer) that contains a part (the not yet written part of the buffer) that doesn’t represent any of the actual data, so that there would be a clear mismatch between the data structure and the data that it represents. All in all, this approach looks quite low-level to me.

This alternative might be preferable if it has clear benefits in terms of memory or time usage, but I even have doubts that it does. With both your and my approach the data has to be copied once, and with your approach the baler always has the whole buffer allocated, while with my approach it can often share data with other parts of the software (when a baler is used for an ordinary index, the serialized keys are shared).

As a result, I’m not really convinced that changing the implementation in the above-mentioned way is what we should do. What do you think?

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: That’s a good question. I’d say that it depends on how the output generated during incremental serialization is used. Are there uses that require chunks of a fixed size?

In any case, I think that this question should be further considered but doesn’t really belong to this particular pull request, as it touches more fundamental design issues and affects indexes in general, not just the ordinary index. Would you perhaps like to create an issue that calls for further discussion of this topic and optionally a change of the interfaces and implementations of all indexes?

src/Database/LSMTree/Internal/Chunk.hs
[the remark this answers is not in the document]
jorisdral: GHC *might* figure that out, but in the case of most GHC optimisations it is a good rule of thumb is to not trust that GHC will *always* fire optimisations, even if it looks like a simple one. It is better to be explicit

src/Database/LSMTree/Internal/Chunk.hs
jorisdral: Yup, typo

src/Database/LSMTree/Internal/Chunk.hs
jorisdral: AFAICS none of our code relies on the code being a specific size. https://github.com/IntersectMBO/lsm-tree/issues/303

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: Yes.

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: I know that `writeSTRef` is not strict. My point was that a mere application of a data constructor would be in head normal form from the start, so that there would be nothing to evaluate, although my later considerations above suggest that this is likely to be different for data constructors with strict fields. Because of the latter, adding `$!` seems to make sense indeed.

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: This all sounds very sensible to me. 🙂

src/Database/LSMTree/Internal/Chunk.hs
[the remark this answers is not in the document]
jeltsch: Changed in 9c8dec880f539cc47027a16e5872696bc4240850. 🙂

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: I’ve made the suggested change in 9c8dec880f539cc47027a16e5872696bc4240850.

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: The change you suggested is in d0384bfa5bf54d4d1debea8ca52098e3a48e5c7b.

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: The `writeSTRef` strictness issue is now fixed in 43c4f491b1d6b9460294cf3aed6c5c5b85c55f83.

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: The `writeSTRef` strictness issue is now fixed in 43c4f491b1d6b9460294cf3aed6c5c5b85c55f83.

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: > The intent of the `lsm-tree` package and its sub-components is to be fast, so it is a reasonable trade-off to write lower-level code in favour of performance. I understand that nicer code is probably what we'd all prefer to write, but we can use unsafe functions/features and mutable data structures without sacrificing "nicety" by ensuring safety at the API boundary of this module. Moreover, if we have to sacrifice generality in favour of performance, for example by making `Baler` more specific to the `lsm-tree` use case, then I also think that is a road we should take

A thing to consider here is that the `Chunk` module is primarily intended to support the ordinary index, which should be a general-purpose solution, unlike the compact index, which is to be tailored towards the specific use case of the Cardano blockchain.

jeltsch: > > Furthermore, it would involve a data structure (the chunk buffer) that contains a part (the not yet written part of the buffer) that doesn’t represent any of the actual data, so that there would be a clear mismatch between the data structure and the data that it represents.
> 
> I'm not sure what you mean with this mismatch. They are the same bytes, aren't they?

The unwritten part of the chunk buffer would contain bytes that do not correspond to any bytes that are scheduled to be part of the next chunk; so it would contain redundant information. However, I’ve realized meanwhile that the queue data structure also contains redundant information: it remembers where the boundaries between the input blocks are. As a result, the redundancy argument isn’t a valid one.

jeltsch: > I think there are benefits: […]

These arguments make sense to me for the most part; so I think I’ll go for the solution with the chunk buffer.

jeltsch: Actually, when changing the implementation to use a chunk buffer, I’ll already change it also according to #303.

jeltsch: The new implementation is available in 5713127cfc0b9b15fde5f5d04c4ab8edde696dd3.

src/Database/LSMTree/Internal/Chunk.hs
```diff
@@ -29,81 +35,67 @@ fromChunk (Chunk content) = content
...
         !totalSize = remnantSize + length block
 
-    if totalSize < maxChunkSize
+    if totalSize <= Mutable.length buffer
         then do
-                 let
-
-                     state' :: BalerState
-                     !state' = BalerState (block : remnantBlocksRev) totalSize
+                 copy (Mutable.slice remnantSize (length block) buffer) block
```
jorisdral, line 72: `unsafeCopy`

src/Database/LSMTree/Internal/Chunk.hs
```diff
@@ -29,81 +35,67 @@ fromChunk (Chunk content) = content
...
+                 writeSTRef remnantSizeRef totalSize
+                 return Nothing
+        else do
+                 protoChunk <- Mutable.unsafeNew totalSize
+                 Mutable.copy (Mutable.take remnantSize protoChunk)
```
jorisdral, line 77: `unsafeCopy`

src/Database/LSMTree/Internal/Chunk.hs
```diff
@@ -29,81 +35,67 @@ fromChunk (Chunk content) = content
...
+                              (Mutable.take remnantSize buffer)
+                 copy (Mutable.drop remnantSize protoChunk) block
```
jorisdral, line 79: `unsafeCopy`

jorisdral (approved):

src/Database/LSMTree/Internal/Chunk.hs
```diff
@@ -29,81 +35,67 @@ fromChunk (Chunk content) = content
...
+                   !(STRef s Int)     -- Reference to the number of queued bytes
 
 -- | Creates a new baler.
-createBaler :: Int            -- ^ Maximum chunk size in bytes
+createBaler :: Int            -- ^ Minimum chunk size in bytes
             -> ST s (Baler s) -- ^ Creation of the baler
-createBaler maxChunkSize = Baler maxChunkSize <$> (newSTRef $! BalerState [] 0)
+createBaler minChunkSize = assert (minChunkSize > 0)              $
+                           Baler                                 <$>
+                           Mutable.unsafeNew (pred minChunkSize) <*>
```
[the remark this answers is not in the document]
jeltsch: As soon as there are as many bytes as the minimum chunk size, we construct the chunk, and the block that made us reach the minimum chunk size will not get queued. Therefore, one byte less is enough.

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: Oh, I didn’t know about this. I’ll made this change and the other ones mentioned below.

jeltsch: Fixed in b9c3c88968b37a78bdf3beba6338225f2d12bcfd.

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: Fixed in b9c3c88968b37a78bdf3beba6338225f2d12bcfd.

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: Fixed in b9c3c88968b37a78bdf3beba6338225f2d12bcfd.

PR #299, opened 2024-07-17
Add incremental functionality for the ordinary index

jeltsch:
This adds support for incremental construction and serialization of ordinary indexes.

Note that the branch of this pull request is currently stacked onto the branch of #296, as it depends on it. Once #296 is merged, I shall rebase the branch of the present pull request onto the then current `main`.

### DIFF
--- a/src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
+++ b/src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
@@ -1,0 +1,124 @@
+{- HLINT ignore "Avoid restricted alias" -}
+
+{-|
+    Incremental construction functionality for the general-purpose fence pointer
+    index.
+-}
+module Database.LSMTree.Internal.IndexOrdinaryAcc
+(
+    IndexOrdinaryAcc,
+    new,
+    append,
+    unsafeEnd
+)
+where
+
+import           Prelude hiding (take)
+
+import           Control.Exception (assert)
+import           Control.Monad.ST.Strict (ST, runST)
+import           Data.Primitive.ByteArray (newByteArray, unsafeFreezeByteArray,
+                     writeByteArray)
+import           Data.STRef.Strict (STRef, newSTRef, readSTRef, writeSTRef)
+import           Data.Vector (force, take, unsafeFreeze)
+import           Data.Vector.Mutable (MVector)
+import qualified Data.Vector.Mutable as Mutable (unsafeNew, write)
+import qualified Data.Vector.Primitive as Primitive (Vector, length)
+import           Data.Word (Word16, Word8)
+import           Database.LSMTree.Internal.Chunk (Baler, Chunk, createBaler,
+                     feedBaler, unsafeEndBaler)
+import           Database.LSMTree.Internal.IndexCompactAcc
+                     (Append (AppendMultiPage, AppendSinglePage))
+import           Database.LSMTree.Internal.IndexOrdinary
+                     (IndexOrdinary (IndexOrdinary))
+import           Database.LSMTree.Internal.Serialise
+                     (SerialisedKey (SerialisedKey'))
+import           Database.LSMTree.Internal.Vector (mkPrimVector)
+
+{-|
+    A general-purpose fence pointer index under incremental construction.
+
+    A value @IndexOrdinaryAcc buffer keyCountRef baler@ denotes a partially
+    constructed index with the following properties:
+
+      * The keys that the index assigns to pages are stored as a prefix of the
+        mutable vector @buffer@.
+      * The reference @keyCountRef@ points to the number of those keys.
+      * The @baler@ object is used by the index for incremental output of the
+        serialised key list.
+-}
+data IndexOrdinaryAcc s = IndexOrdinaryAcc
+                              !(MVector s SerialisedKey)
+                              !(STRef s Int)
+                              !(Baler s)
+
+-- | Creates a new, initially empty, index.
+new :: Int                       -- ^ Maximum number of keys
+    -> Int                       -- ^ Minimum chunk size in bytes
+    -> ST s (IndexOrdinaryAcc s) -- ^ Construction of the index
+new maxKeyCount minChunkSize = assert (maxKeyCount >= 0)      $
+                               IndexOrdinaryAcc              <$>
+                               Mutable.unsafeNew maxKeyCount <*>
+                               newSTRef 0                    <*>
+                               createBaler minChunkSize
+
+{-|
+    Appends keys to the key list of an index and outputs newly available chunks
+    of the serialised key list.
+
+    __Warning:__ Appending keys whose length cannot be represented by a 16-bit
+    word may result in a corrupted serialised key list.
+-}
+append :: Append -> IndexOrdinaryAcc s -> ST s (Maybe Chunk)
+append instruction (IndexOrdinaryAcc buffer keyCountRef baler)
+    = case instruction of
+          AppendSinglePage _ key -> do
+              keyCount <- readSTRef keyCountRef
+              Mutable.write buffer keyCount key
+              writeSTRef keyCountRef $! succ keyCount
+              feedBaler (keyListElem key) baler
+          AppendMultiPage key overflowPageCount -> do
+              keyCount <- readSTRef keyCountRef
+              let
+
+                  pageCount :: Int
+                  !pageCount = succ (fromIntegral overflowPageCount)
+
+                  keyCount' :: Int
+                  !keyCount' = keyCount + pageCount
+
+              mapM_ (flip (Mutable.write buffer) key)
+                    [keyCount .. pred keyCount']
+              writeSTRef keyCountRef $! keyCount'
+              feedBaler (concat (replicate pageCount (keyListElem key))) baler
+    where
+
+    keyListElem :: SerialisedKey -> [Primitive.Vector Word8]
+    keyListElem (SerialisedKey' keyBytes) = [keySizeBytes, keyBytes] where
+
+        keySize :: Int
+        !keySize = Primitive.length keyBytes
+
+        keySizeAsWord16 :: Word16
+        !keySizeAsWord16 = assert (keySize <= fromIntegral (maxBound :: Word16)) $
+                           fromIntegral keySize
+
+        keySizeBytes :: Primitive.Vector Word8
+        !keySizeBytes = mkPrimVector 0 2 $
+                        runST $ do
+                            rep <- newByteArray 2
+                            writeByteArray rep 0 keySizeAsWord16
+                            unsafeFreezeByteArray rep
+
+{-|
+    Returns the constructed index, along with a final chunk in case the
+    serialised key list has not been fully output yet, thereby invalidating the
+    index under construction. Executing @unsafeEnd index@ is only safe when
+    @index@ is not used afterwards.
+-}
+unsafeEnd :: IndexOrdinaryAcc s -> ST s (Maybe Chunk, IndexOrdinary)
+unsafeEnd (IndexOrdinaryAcc buffer keyCountRef baler) = do
+    keyCount <- readSTRef keyCountRef
+    keys <- force <$> take keyCount <$> unsafeFreeze buffer
+    remnant <- unsafeEndBaler baler
+    return (remnant, IndexOrdinary keys)
--- a/lsm-tree.cabal
+++ b/lsm-tree.cabal
@@ -119,6 +119,7 @@ library
     Database.LSMTree.Internal.IndexCompact
     Database.LSMTree.Internal.IndexCompactAcc
     Database.LSMTree.Internal.IndexOrdinary
+    Database.LSMTree.Internal.IndexOrdinaryAcc
     Database.LSMTree.Internal.Lookup
     Database.LSMTree.Internal.Managed
     Database.LSMTree.Internal.Merge
--- a/src/Database/LSMTree/Internal/Chunk.hs
+++ b/src/Database/LSMTree/Internal/Chunk.hs
@@ -18,6 +18,7 @@ import           Prelude hiding (length)
 
 import           Control.Exception (assert)
 import           Control.Monad.ST.Strict (ST)
+import           Data.List (scanl')
 import           Data.STRef.Strict (STRef, newSTRef, readSTRef, writeSTRef)
 import           Data.Vector.Primitive (Vector, length, unsafeCopy,
                      unsafeFreeze)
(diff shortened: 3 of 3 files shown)


src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
```diff
@@ -0,0 +1,100 @@
...
+          (IndexOrdinaryAcc lastKeysRevRef baler)
+    = do
+          modifySTRef lastKeysRevRef (lastKey :)
+          chunksIntoLastKeySize <- feedBaler lastKeySizeBytes baler
+          chunksIntoLastKey <- feedBaler lastKeyBytes baler
+          return (chunksIntoLastKeySize ++ chunksIntoLastKey)
+    where
+
+    lastKeySize :: Int
+    lastKeySize = Primitive.length lastKeyBytes
```
jorisdral, line 66:
```suggestion
    !lastKeySize = Primitive.length lastKeyBytes
```

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
```diff
@@ -0,0 +1,100 @@
...
+
+    lastKeySizeAsWord16 :: Word16
+    lastKeySizeAsWord16
```
jorisdral, line 69:
```suggestion
    !lastKeySizeAsWord16
```

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
```diff
@@ -0,0 +1,100 @@
...
+        | lastKeySize <= fromIntegral (maxBound :: Word16)
+            = fromIntegral lastKeySize
+        | otherwise
+            = error "Serialised key too large"
+
+    lastKeySizeBytes :: Primitive.Vector Word8
+    lastKeySizeBytes = mkPrimVector 0 2 $
```
jorisdral, line 76:
```suggestion
    !lastKeySizeBytes = mkPrimVector 0 2 $
```

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
```diff
@@ -0,0 +1,100 @@
...
+                       byteArrayFromListN 1 [lastKeySizeAsWord16]
```
jorisdral, line 77: It's minor, but this will allocate a list `Cons` and `Nil` for every key. We can probably get around that by creating the byte array ourselves without `byteArrayFromListN`.

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
```diff
@@ -0,0 +1,100 @@
...
+{-|
+    Appends keys to the key list of an index and outputs newly available chunks
+    of the serialised key list.
+-}
+append :: Append -> IndexOrdinaryAcc s -> ST s [Chunk]
+append (AppendSinglePage _ lastKey)    index = appendKey lastKey index
+append (AppendMultiPage key pageCount) index = fmap concat                $
+                                               sequence                   $
+                                               genericReplicate pageCount $
+                                               appendKey key index
```
jorisdral, line 88: This seems suboptimal: I think this should have a specialised function that does everything in one go instead of replicating the single-page case

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
```diff
@@ -0,0 +1,100 @@
...
+
+    A value @IndexOrdinaryAcc lastKeysRevRef baler@ denotes a partially
+    constructed index with the following properties:
+
+      * The keys pointed to by @lastKeysRevRef@ are the keys that the index
+        assigns to pages, but in reverse order.
+      * The @baler@ object is used by the index for incremental output of the
+        serialised key list.
+-}
+data IndexOrdinaryAcc s = IndexOrdinaryAcc !(STRef s [SerialisedKey]) !(Baler s)
```
jorisdral, line 44: Lists are probably not the right accumulator type. It would be better if we could use mutable vectors or something like that. See my comments on #296

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
```diff
@@ -0,0 +1,100 @@
...
+
+{-
+    Appends a single key to the key list of an index and outputs newly available
+    chunks of the serialised key list.
+-}
+appendKey :: SerialisedKey -> IndexOrdinaryAcc s -> ST s [Chunk]
+appendKey lastKey@(SerialisedKey' lastKeyBytes)
+          (IndexOrdinaryAcc lastKeysRevRef baler)
+    = do
+          modifySTRef lastKeysRevRef (lastKey :)
```
jorisdral, line 59:
```suggestion
          modifySTRef' lastKeysRevRef (lastKey :)
```

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
```diff
@@ -0,0 +1,100 @@
...
+          chunksIntoLastKeySize <- feedBaler lastKeySizeBytes baler
+          chunksIntoLastKey <- feedBaler lastKeyBytes baler
```
jorisdral, line 61: One could write a `feedBaler2` function that gets two arguments, instead of calling `feedBaler` twice

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
```diff
@@ -0,0 +1,100 @@
...
+
+    lastKeySize :: Int
+    lastKeySize = Primitive.length lastKeyBytes
+
+    lastKeySizeAsWord16 :: Word16
+    lastKeySizeAsWord16
+        | lastKeySize <= fromIntegral (maxBound :: Word16)
+            = fromIntegral lastKeySize
+        | otherwise
+            = error "Serialised key too large"
```
jorisdral, line 73: The index does not have to check this. It should be checked by something like the `PageAcc`. https://github.com/IntersectMBO/lsm-tree/issues/304

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
```diff
@@ -0,0 +1,100 @@
...
+    End an incremental construction of an index.
+
+    This returns the constructed index, along with a final chunk in case the
+    serialised key list has not been fully output yet.
+-}
+end :: IndexOrdinaryAcc s -> ST s (Maybe Chunk, IndexOrdinary)
+end (IndexOrdinaryAcc lastKeysRevRef baler) = do
+    lastKeysRev <- readSTRef lastKeysRevRef
+    remnant <- readBalerRemnant baler
+    return (remnant, IndexOrdinary (fromList (reverse lastKeysRev)))
```
jorisdral, line 100: `fromList (reverse lastKeysRev))` seems like a relatively costly operation. It might be better to use a mutable vector, as mentioned before.

jorisdral: Note: we know the *maximum* number of keys beforehand. It's a parameter given to the `RunAcc`, and from there it could also be sent to this construction code. As such, we can pre-allocate the mutable vector

https://github.com/IntersectMBO/lsm-tree/blob/main/src/Database/LSMTree/Internal/RunAcc.hs#L85-L86

jorisdral:
> Note that the branch of this pull request is currently stacked onto the branch of https://github.com/IntersectMBO/lsm-tree/pull/296, as it depends on it. Once https://github.com/IntersectMBO/lsm-tree/pull/296 is merged, I shall rebase the branch of the present pull request onto the then current main.

You can set #296 as the target branch for this PR, and then the PR only shows the changes with respect to #296. When #296 is merged, the target branch automatically changes to `main` and you won't have to rebase

jeltsch:
> You can set #296 as the target branch for this PR, and then the PR only shows the changes with respect to #296.

Done. 🙂

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
jeltsch: Fixed in 02cf840c1fbe6a37cef8c21e2088e0210e6a0d66.

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
jeltsch: Fixed in 02cf840c1fbe6a37cef8c21e2088e0210e6a0d66.

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
jeltsch: Fixed in 02cf840c1fbe6a37cef8c21e2088e0210e6a0d66.

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
jeltsch: I think that this module should still provide a somewhat safe interface, despite being internal. The `SerialisedKey` type allows serialized keys to have any size that can be represented by an `Int`, so that the interface of this module doesn’t prevent keys from being larger than what can be represented by a `Word16`. What makes matters worse is that the `fromIntegral` implementation for `Word16` computes remainders modulo 2¹⁶, so that, without a check, submitting too large keys will silently corrupt the index, in a somewhat subtle way.

I think that there should at least be an assertion in this place. The API documentation should then clearly state the size constraint for keys, as assertion checking may be switched off.

jeltsch: I made a corresponding change in f1e84da3bad7307b0468eee806552dd8cdddb7c6. Please tell me whether you’re happy with it or not. 🙂

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
jeltsch: I made a respective change in 87a9d4d788e980707917730d630425e18ff90fd1. Does this correspond to what you had in mind?

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
jeltsch: How tight is this upper bound expected to be? Can we be reasonably sure that we won’t waste too much space with this approach? Is the bound maybe even so tight that we can just use the contents of that mutable vector for the finally constructed index without copying?

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
jeltsch: Hmm, having a function for the very special case of 2 blocks seems to go a bit the wrong way in my opinion. However, what about changing `feedBaler` such that it doesn’t accept a single block but a list of blocks?

I think that this change would be advisable anyhow. Currently, the baler implementation already respects #303 by having `feedBaler` output at most one chunk. However, the ordinary index can’t fully profit from this yet, because it calls `feedBaler` several times per appending. Having `feedBaler` accept an arbitrary number of blocks but still produce at most one chunk would enable `append` to also output at most one chunk, even in the case of a multi-page appending.

What are your thoughts on this?

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
jorisdral: This upper bound isn't super tight. During a merge, resolving `n` duplicate keys produces *at most* 1 key -- for intermediate levels, it's going to be exactly `1`, on the last level it can be `0` or `1`. 

We can avoid wasting too much space by doing a [`force`](https://hackage.haskell.org/package/vector-0.13.1.0/docs/Data-Vector-Primitive.html#v:force)

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
jorisdral: Hmm a list is probably nicest, yes. It would also mean that we can use that in the `AppendMultiPage` case

jeltsch: The `feedBaler` function has been changed accordingly in `4d9463ff6f0c5f3cfafbcc4e82caa4a992229488`.

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
jeltsch: Changed in f84d4de5af557746cf3b9471acc790f1dd6e7218.

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
jeltsch: Changed in f84d4de5af557746cf3b9471acc790f1dd6e7218 and 33ceb2d0bbb019f6543ebd982943bdc6a116588a.

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
jeltsch: Fixed in 916a5e84be04684ef87a3cf60b78233c2ef80361.

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
jeltsch: With 916a5e84be04684ef87a3cf60b78233c2ef80361, `append` fully makes use of the extended `feedBaler` functionality, producing at most one chunk.

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
```diff
@@ -64,47 +63,51 @@ new maxKeyCount minChunkSize = assert (maxKeyCount >= 0)      $
...
-                                               appendKey key index
+append :: Append -> IndexOrdinaryAcc s -> ST s (Maybe Chunk)
+append instruction (IndexOrdinaryAcc buffer keyCountRef baler)
+    = case instruction of
+          AppendSinglePage _ key -> do
+              keyCount <- readSTRef keyCountRef
+              Mutable.write buffer keyCount key
+              writeSTRef keyCountRef $! succ keyCount
+              feedBaler (keyListElem key) baler
+          AppendMultiPage key pageCount -> do
```
jorisdral, line 81: `pageCount` is not the number of pages, it's the number of *overflow pages*. For `AppendMultiPage _ noverflow`, `pagecount = noverflow + 1`.

jorisdral (approved):
LGTM. Let's merge after you've squashed the commits, and after you've resolved my last comment.

src/Database/LSMTree/Internal/IndexOrdinaryAcc.hs
jeltsch: Oops, you’re right. Fixed in a4fcb497a9d386a2a6aba4e9528b2e27fb4c48de.

jeltsch:
> LGTM. Let's merge after you've squashed the commits, and after you've resolved my last comment.

Squashed into 1b3158bebfa4122de1491ef115d50b0d6ccce5ea. Ready to be merged.

PR #300, opened 2024-07-18
Add file locking to `blockio-api`

### LINKED ISSUE #240: Test failure in `twiceOpenSession` on Windows
https://github.com/IntersectMBO/lsm-tree/actions/runs/9387087243/job/25849161692#step:15:506

jorisdral:
Should resolve #240

### DIFF
--- a/blockio-api/src/System/FS/BlockIO/API.hs
+++ b/blockio-api/src/System/FS/BlockIO/API.hs
@@ -23,12 +23,19 @@ module System.FS.BlockIO.API (
   , Advice (..)
   , hAdviseAll
   , hDropCacheAll
+    -- * File locks
+  , GHC.LockMode (..)
+  , GHC.FileLockingNotSupported (..)
+  , LockFileHandle (..)
+  , tryLockFileIO
     -- * Re-exports
   , ByteCount
   , FileOffset
   ) where
 
 import           Control.DeepSeq
+import           Control.Monad.Class.MonadThrow (MonadCatch (bracketOnError),
+                     MonadThrow (..), bracketOnError, try)
 import           Control.Monad.Primitive (PrimMonad (PrimState))
 import           Data.Primitive.ByteArray (MutableByteArray)
 import qualified Data.Vector as V
@@ -38,8 +45,13 @@ import qualified Data.Vector.Primitive as VP
 import qualified Data.Vector.Unboxed as VU
 import qualified Data.Vector.Unboxed.Mutable as VUM
 import           GHC.IO.Exception (IOErrorType (ResourceVanished))
+import qualified GHC.IO.Handle.Lock as GHC
 import           GHC.Stack (HasCallStack)
-import           System.FS.API
+import qualified System.FS.API as FS
+import           System.FS.API (BufferOffset, FsError (..), FsPath, Handle (..),
+                     HasFS, SomeHasFS (..))
+import           System.FS.IO (HandleIO)
+import qualified System.IO as GHC
 import           System.IO.Error (ioeSetErrorString, mkIOError)
 import           System.Posix.Types (ByteCount, FileOffset)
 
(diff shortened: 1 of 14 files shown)


dcoutts (approved):
LGTM.

Is there anything remaining to move this from draft status? (apart from rebasing on main of course)

jorisdral:
> LGTM.
> 
> Is there anything remaining to move this from draft status? (apart from rebasing on main of course)

I don't remember, but it's probably just a matter of cleaning up some code and the commit messages. I parked this PR for a while in favour of more pressing PRs, but I'll soon get around to fixing up this PR

blockio-api/src/System/FS/BlockIO/API.hs
```diff
@@ -170,3 +210,40 @@ hAdviseAll hbio h advice = hAdvise hbio h 0 0 advice -- len=0 implies until the
...
+rethrowFsErrorIO :: HasCallStack => HasFS IO HandleIO -> FsPath -> IO a -> IO a
+rethrowFsErrorIO hfs fp action = do
+    res <- try action
+    case res of
+      Left err -> handleError err
+      Right a  -> return a
+  where
+    handleError :: HasCallStack => IOError -> IO a
+    handleError ioErr =
+      throwIO $ FS.ioToFsError (FS.mkFsErrorPath hfs fp) ioErr
```
mheinzel, line 249: It's not exported there, so not sure how to avoid duplicating it. Also, I wonder why it's not just doing ``action `catch` handleError``.

blockio-api/src/System/FS/BlockIO/API.hs
```diff
@@ -170,3 +210,40 @@ hAdviseAll hbio h advice = hAdvise hbio h 0 0 advice -- len=0 implies until the
...
+    -- | Release a file lock acquired using 'tryLockFile'.
+    hUnlock :: m ()
+  }
+
+tryLockFileIO :: HasFS IO HandleIO -> FsPath -> GHC.LockMode -> IO (Maybe (LockFileHandle IO))
+tryLockFileIO hfs fsp mode = do
+    fp <- FS.unsafeToFilePath hfs fsp -- shouldn't fail
+    rethrowFsErrorIO hfs fsp $
+      bracketOnError (GHC.openFile fp GHC.WriteMode) GHC.hClose $ \h -> do
+        bracketOnError (GHC.hTryLock h mode) (\_ -> GHC.hUnlock h) $ \b -> do
```
mheinzel, line 229: Do we need the second `bracketOnError` with `hUnlock` or is it enough to rely on `hClose`, which should also release the lock? Same for the unlock+close below.

We can just leave it as is, though. It's probably clearer.

mheinzel (approved):
Looks good, just a few questions that might lead to minor changes if you think they make sense.

src/Database/LSMTree/Internal.hs
```diff
@@ -245,21 +242,18 @@ openSession hfs hbio dir = do
...
     -- is included in the returned Session.
     bracketOnError
-      (acquireLock lf)
-      (traverse_ System.hClose) -- also releases the lock
+      acquireLock
+      releaseLock
       $ \case
-          Nothing -> throwIO (SessionDirLocked (FS.mkFsErrorPath hfs lockFilePath))
-          Just sessionFileLock ->
+          Left _e       -> throwIO (SessionDirLocked (FS.mkFsErrorPath hfs lockFilePath))
```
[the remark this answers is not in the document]
jorisdral: Sounds like a good idea. We should be able to distinguish between expected exceptions and unexpected ones. I'll add that to the PR

blockio-api/src/System/FS/BlockIO/API.hs
jorisdral: Good question, I don't know :stuck_out_tongue:

blockio-api/src/System/FS/BlockIO/API.hs
jorisdral: I think closing the handle *should* release the lock, but alas I'm not sure. It probably doesn't hurt to be extra safe by doing `bracketOnError` twice

mheinzel (approved):
Nice, looks good!

PR #312, opened 2024-07-26
`DiskCachePolicy` option for WP8 macro-benchmark

jorisdral:


src/Database/LSMTree/Internal.hs
```diff
@@ -1100,13 +1104,34 @@ open sesh label snap = do
...
-      with (openLevels hfs hbio (confDiskCachePolicy conf) runPaths) $ \lvls -> do
+      with (openLevels hfs hbio (confDiskCachePolicy conf') runPaths) $ \lvls -> do
         am <- newArenaManager
-        (newWith sesh seshEnv conf am WB.empty lvls)
+        (newWith sesh seshEnv conf' am WB.empty lvls)
+
+data TableConfigOverride = TableConfigOverride {
+      -- | Override for 'confDiskCachePolicy'
+      confOverrideDiskCachePolicy  :: Maybe DiskCachePolicy
+    }
```
mheinzel, line 1119: This could use a comment giving the reasoning behind the type.

What I gather is that most parts of the table config are fixed, so they should stay the same even after creating and opening a snapshot. But some config options can safely be changed, such as the disk cache poilcy (write buffer and bloom filter alloc should probably also qualify). So we want to ensure that when the config is read from disk, most of it can't be changed by the user, but some options can. Makes sense.

An alternative would be to split the type into the options that are inherent properties of a table and the ones that can be changed on the fly, and then only store the first part in a snapshot and always require the second one when opening. Although that might cause some churn, now and also when moving other options between the two parts (it's not always obvious whether something can be changed or not). :thinking:

src/Database/LSMTree/Internal.hs
```diff
@@ -1100,13 +1104,34 @@ open sesh label snap = do
...
+
+configNoOverride :: TableConfigOverride
+configNoOverride = TableConfigOverride {
+      confOverrideDiskCachePolicy = Nothing
+    }
+
+configOverrideDiskCachePolicy :: DiskCachePolicy -> TableConfigOverride
+configOverrideDiskCachePolicy pol = TableConfigOverride {
+      confOverrideDiskCachePolicy = Just pol
+    }
```
mheinzel, line 1134: This does not compose, there is no way to create an override that overrides two settings. Maybe add a monoid instance using `Last` for the fields? This would also give you a convenient way of specifying no overrides.

jorisdral: I've added it in

src/Database/LSMTree/Internal.hs
jorisdral: > What I gather is that most parts of the table config are fixed, so they should stay the same even after creating and opening a snapshot. But some config options can safely be changed, such as the disk cache poilcy (write buffer and bloom filter alloc should probably also qualify). So we want to ensure that when the config is read from disk, most of it can't be changed by the user, but some options can. Makes sense.

Right, that is indeed the reasoning

> An alternative would be to split the type into the options that are inherent properties of a table and the ones that can be changed on the fly, and then only store the first part in a snapshot and always require the second one when opening. Although that might cause some churn, now and also when moving other options between the two parts (it's not always obvious whether something can be changed or not). 🤔

I think the advantage of an override is that you aren't required to provide values for things that you don't want to change. I'm also foreseeing a use case where one could change some of these options dynamically on a live table, and not just when using `new`/`open`

PR #313, opened 2024-07-30
Start using the `BloomFilterAlloc` configuration option

jorisdral:


src/Database/LSMTree/Internal.hs
```diff
@@ -1347,6 +1345,10 @@ data BloomFilterAlloc =
 -- persistence of the config to disk.
 deriving stock instance Read BloomFilterAlloc
 
+bloomFilterAllocForLevel :: BloomFilterAlloc -> LevelNo -> RunBloomFilterAlloc
+bloomFilterAllocForLevel (AllocFixed n) _        = RunAllocFixed n
+bloomFilterAllocForLevel (AllocRequestFPR fpr) _ = RunAllocRequestFPR fpr
```
mheinzel, line 1350: Ah, so this is why there are two types. Once we add support for Monkey calculations, there will be a third option for `BloomFilterAlloc`, but that will translate to one of the existing constructors in `RunBloomFilterAlloc`, right?

src/Database/LSMTree/Internal/Run.hs
```diff
@@ -196,11 +197,12 @@ fromMutable fs hbio caching refCount builder = do
 fromWriteBuffer :: HasFS IO h
                 -> HasBlockIO IO h
                 -> RunDataCaching
+                -> RunBloomFilterAlloc
```
mheinzel, line 200: We could bundle these two arguments (and in the future potentially more things) into a single run creation config/setting type (also in `Merge.new`, `mergeRuns`), but maybe it's best to wait and only do it once we really add a third thing.

src/Database/LSMTree/Internal/Monkey.hs
```diff
@@ -37,8 +39,8 @@ monkeyFPR numBits numEntries =
 monkeyBits ::
      Int    -- ^ Number of entries inserted into the bloom filter.
   -> Double -- ^ False positive rate.
-  -> Int
-monkeyBits numEntries fpr = ceiling $
+  -> Word64
+monkeyBits numEntries fpr = max 1 $ ceiling $
   (- fromIntegral numEntries) * (log fpr / (log 2 ** 2))
```
mheinzel, line 44: This leads to underflow for FPR > 1, which shouldn't happen, but maybe best to add a guard or at least switch `max` and `ceiling`?

src/Database/LSMTree/Internal/RunAcc.hs
```diff
@@ -71,13 +74,24 @@ data RunAcc s = RunAcc {
...
 -- output run.
-new :: NumEntries -> ST s (RunAcc s)
-new (NumEntries nentries) = do
-    mbloom <- Bloom.Easy.easyNew 0.02 nentries -- TODO(optimise): tune bloom filter
+new :: NumEntries -> RunBloomFilterAlloc -> ST s (RunAcc s)
+new (NumEntries nentries) alloc = do
+    mbloom <- case alloc of
+      RunAllocFixed bitsPerEntry    ->
+        let !nbits = bitsPerEntry * fromIntegral nentries
+        in  MBloom.new (monkeyHashFuncs nbits nentries) nbits
```
mheinzel, line 92: Just to be sure, despite the naming, this function is not specific to the monkey calculations, right? It's just the optimal number of hash functions for a given number of bits and entries.

mheinzel (approved):
Mostly straightforward plumbing, just had a few small comments/questions.

src/Database/LSMTree/Internal.hs
jorisdral: I might add a separate constructor to `RunBloomFilterAlloc`, but yes the types `BloomFilterAlloc` and `RunBloomFilterAlloc` would be slightly different

src/Database/LSMTree/Internal/RunAcc.hs
jorisdral: Yup, the name currently only refers to `monkey` because that is where I got the equation from

src/Database/LSMTree/Internal/Monkey.hs
jorisdral: Good catch -- I've switched them around

PR #315, opened 2024-07-30
Implement monoidal table API

mheinzel:
~~Depends on #314. Before un-drafting this, I want to make sure all the tests are enabled and run successfully.~~

As a bonus, the `Internal` module now doesn't know anything about the normal/monoidal distinction. We prevent opening a normal snapshot as monoidal through the labels (although we might iterate on the exact design of that).

I haven't touched any of the TODOs from #314 yet, we should first discuss what we want to do there.

The tests all seem to be enabled now, unless I missed something.

### DIFF
--- a/src/Database/LSMTree/Internal.hs
+++ b/src/Database/LSMTree/Internal.hs
@@ -27,6 +27,7 @@ module Database.LSMTree.Internal (
   , close
   , lookups
   , toNormalLookupResult
+  , toMonoidalLookupResult
   , updates
     -- * Snapshots
   , SnapshotLabel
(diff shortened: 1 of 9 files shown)


jorisdral:
Just a quick remark that you should probably enable some of the monoidal class tests manually, because currently they pass the `expectFailure`s with the error `ErrExpectedMonoidalTable`

mheinzel:
Yes, that's what I mean by making sure the tests are enabled.

jorisdral (commented):
LGTM!

> I haven't touched any of the TODOs from https://github.com/IntersectMBO/lsm-tree/pull/314 yet, we should first discuss what we want to do there.

Could you maybe list here which TODOs those are?

jorisdral (approved):

mheinzel:
> Could you maybe list here which TODOs those are?

There are a few open questions about `ResolveValue`:

- revisit totality requirement
- potentially move it to another module
- potentially add `SerialiseValue` as a superclass

PR #316, opened 2024-07-30
Add missing specialisations in Normal API

jorisdral:


mheinzel (approved):
I left a note to do the same! :smile:

PR #318, opened 2024-08-01
Revive `ghc-8.10.7` support

jorisdral:


test/Test/Database/LSMTree/Normal/StateMachine.hs
```diff
@@ -1030,7 +1031,7 @@ shrinkActionWithVars ::
   -> LockstepAction (ModelState h) a
   -> [Any (LockstepAction (ModelState h))]
 shrinkActionWithVars _ _ = \case
-    New @k @v @blob conf -> [ Some $ New @k @v @blob conf' | conf' <- QC.shrink conf ]
+    New p conf -> [ Some $ New p conf' | conf' <- QC.shrink conf ]
```
jorisdral, line 1034: Annoying, but we can't bind type variables like this in `ghc-8.10.7`, so I added a `Proxy` argument instead. Seemed the simplest solution

test/Test/Database/LSMTree/Normal/StateMachine.hs
```diff
@@ -328,7 +328,8 @@ instance ( Show (Class.TableConfig h)
   data instance Action (Lockstep (ModelState h)) a where
     -- Tables
     New :: C k v blob
-        => Class.TableConfig h
+        => Proxy (k, v, blob)
```
phadej, line 331: `{-# UNPACK #-} !Proxy...` to avoid having it at runtime. Or use `Proxy#`.

test/Test/Database/LSMTree/Normal/StateMachine.hs
```diff
@@ -393,8 +394,8 @@ instance ( Eq (Class.TableConfig h)
   x == y = go x y
     where
       go :: LockstepAction (ModelState h) a -> LockstepAction (ModelState h) a -> Bool
-      go (New conf1)                (New conf2) =
-          conf1 == conf2
+      go (New p1 conf1)                (New p2 conf2) =
+          Just p1 == cast p2 && conf1 == conf2
```
phadej, line 398: `p1 == p2` doesn't do anything. Proxies are always equal.

test/Test/Database/LSMTree/Normal/StateMachine.hs
```diff
@@ -1113,7 +1114,7 @@ updateStats action result =
     -- === Tags
 
     updNewTableTypes stats = case action of
-      New @k @v @blob _ -> stats {
+      New (Proxy :: Proxy (k, v, blob)) _ -> stats {
```
phadej, line 1117: FYI, you can also say `Proxy @(k, v, blob)`.

src/Database/LSMTree/Internal/ByteString.hs
```diff
@@ -55,20 +51,32 @@ tryGetByteArray :: BS.ByteString -> Either String (ByteArray, Int)
...
       Foreign.MallocPtr {} ->
         Left ("unsupported MallocPtr (length " <> show n <> ")")
       Foreign.PlainForeignPtr {} ->
         Left ("unsupported PlainForeignPtr (length " <> show n <> ")")
+#if __GLASGOW_HASKELL__ >= 902
+      Foreign.FinalPtr | n == 0 ->
+        -- We can also handle empty bytestrings ('BS.empty' uses 'FinalPtr').
+        Right (emptyByteArray, 0)
+      Foreign.FinalPtr ->
+          Left ("unsupported FinalPtr (length "  <> show n <> ")")
```
phadej, line 68: indentation is off

lsm-tree.cabal
```diff
@@ -452,7 +505,7 @@ benchmark rocksdb-bench-wp8
   hs-source-dirs: bench/macro src-extras
   main-is:        rocksdb-bench-wp8.hs
 
-  if !(os(linux) && flag(rocksdb))
+  if (!(os(linux) && flag(rocksdb)) || !impl(ghc >=9.2.0))
```
phadej, line 508: `if !os(linux) || !flag(rocksdb) || !impl(ghc >=9.2)` is easier to understand; or
`if !(os(linux) && flag(rocksdb) && impl(ghc >=9.2)`.

test/Test/Database/LSMTree/Normal/StateMachine.hs
jorisdral: I'm casting beforehand 
```
ghci> cast (Proxy @Int) == Just (Proxy @Char)
False
ghci> cast (Proxy @Int) == Just (Proxy @Int)
True
```

phadej: The cast adds type-check which didn't exist before. It's cast which fails, not the equality.

phadej: To clarify: there weren't type-equality check before, so this is semantic change, not a compatibility change.

jorisdral: Okay, I get your point -- I'm now using `eqT` instead and checking that the result of `Just Refl`. The test looks like it can't fail because the type of `New` mentions `k`, `v` and `blob`, but it doesn't hurt to keep it in there

mheinzel (approved):
Looks good to me.

PR #319, opened 2024-08-01
Minor refactoring to add compatability with GHC 9.10

recursion-ninja:
On my machine, `ghc-9.10` is the default compiler. Hence my first attempt at build the project failed... but barely. I notice that the project could easily build with the newest version of GHC with very minor changes to disambiguate some calls to `foldl'`. This PR adds future-proofing compatibility with GHC 9.10.

### DIFF
--- a/bench/macro/lsm-tree-bench-bloomfilter.hs
+++ b/bench/macro/lsm-tree-bench-bloomfilter.hs
@@ -13,7 +13,7 @@ import           Data.BloomFilter (Bloom)
 import qualified Data.BloomFilter as Bloom
 import qualified Data.BloomFilter.Hash as Bloom
 import qualified Data.BloomFilter.Mutable as MBloom
-import           Data.List (foldl')
+import qualified Data.Foldable as Fold
 import           Data.Time
 import           Data.Vector (Vector)
 import qualified Data.Vector as Vector
@@ -264,6 +264,6 @@ benchElemCheapHashes !bs !rng !n =
     let k :: Word256
         (!k, !rng') = uniform rng
         !kh =  Bloom.makeHashes (serialiseKey k)
-     in foldl' (\_ b -> Bloom.elemHashes kh b `seq` ()) () bs
+     in Fold.foldl' (\_ b -> Bloom.elemHashes kh b `seq` ()) () bs
   `seq` benchElemCheapHashes bs rng' (n-1)
 
--- a/blockio-sim/src/System/FS/BlockIO/Sim.hs
+++ b/blockio-sim/src/System/FS/BlockIO/Sim.hs
@@ -56,7 +56,7 @@ simErrorHasBlockIO ::
   -> StrictTVar m Errors
   -> m (HasFS m HandleMock, HasBlockIO m HandleMock)
 simErrorHasBlockIO fsVar errorsVar = do
-    let hfs = mkSimErrorHasFS fsVar errorsVar
+    let hfs = simErrorHasFS fsVar errorsVar
     hbio <- fromHasFS hfs
     pure (hfs, hbio)
 
@@ -66,6 +66,6 @@ simErrorHasBlockIO' ::
   -> Errors
   -> m (HasFS m HandleMock, HasBlockIO m HandleMock)
 simErrorHasBlockIO' mockFS errs = do
-    hfs <- mkSimErrorHasFS' mockFS errs
+    hfs <- simErrorHasFS' mockFS errs
     hbio <- fromHasFS hfs
     pure (hfs, hbio)
--- a/cabal.project
+++ b/cabal.project
@@ -12,11 +12,11 @@ repository cardano-haskell-packages
 
 index-state:
   -- Bump this if you need newer packages from Hackage
-  -- current date: crc32c-0.2.2
-  , hackage.haskell.org 2024-06-05T04:25:30Z
+  -- current date: support ghc-9.10.1
+  , hackage.haskell.org 2024-08-01T00:00:00Z
   -- Bump this if you need newer packages from CHaP
-  -- current date: fs-api-0.2.0.1, fs-sim-0.2.1.1
-  , cardano-haskell-packages 2023-11-30T09:59:24Z
+  -- current date: support ghc-9.10.1
+  , cardano-haskell-packages 2024-08-01T00:00:00Z
 
 packages: .
 
@@ -44,7 +44,13 @@ import: cabal.project.blockio-uring
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/fs-sim
-  tag: 47879aa5edfd3a3f8824d61687e85b8f1586e010
+  tag: 235d5dd10aadb70defa5a2e843a4cadf5d6eb18e
   subdir:
     fs-api
     fs-sim
+
+-- TODO: this relaxing of dependency bounds on @base@ is currently required in order to
+-- build the Haddock documentation with the @scripts/haddock.sh@ script.
+if impl(ghc >=9.10)
+  allow-newer:
+    quickcheck-lockstep:base,
--- a/lsm-tree.cabal
+++ b/lsm-tree.cabal
@@ -12,7 +12,8 @@ category:           Database
 build-type:         Simple
 extra-doc-files:    CHANGELOG.md
 extra-source-files: README.md
-tested-with:        GHC ==8.10 || ==9.2 || ==9.4 || ==9.6 || ==9.8
+tested-with:
+  GHC ==8.10 || ==9.2 || ==9.4 || ==9.6 || ==9.8 || ==9.10
 
 source-repository head
   type:     git
@@ -147,7 +148,7 @@ library
     Database.LSMTree.Normal
 
   build-depends:
-    , base                  >=4.14      && <4.20
+    , base                  >=4.14      && <4.21
     , bitvec                ^>=1.1
     , bytestring            ^>=0.11.4.0 || ^>=0.12.1.0
     , containers
@@ -155,15 +156,15 @@ library
     , deepseq               ^>=1.4      || ^>=1.5
     , filepath
     , fs-api                ^>=0.2
-    , io-classes            ^>=1.4
+    , io-classes            ^>=1.5
     , lsm-tree:blockio-api
     , lsm-tree:bloomfilter
     , lsm-tree:kmerge
     , lsm-tree:monkey
     , nothunks              ^>=0.2
     , primitive             ^>=0.9
-    , strict-mvar           ^>=1.4
-    , strict-stm            ^>=1.4
+    , strict-mvar           ^>=1.5
+    , strict-stm            ^>=1.5
     , vector                ^>=0.13
     , vector-algorithms     ^>=0.9
 
@@ -336,8 +337,8 @@ test-suite lsm-tree-test
     , filepath
     , fs-api
     , fs-sim                   >=0.2
-    , io-classes
-    , io-sim                   >=1.4
+    , io-classes               >=1.5
+    , io-sim                   >=1.5
     , lsm-tree
     , lsm-tree:blockio-api
     , lsm-tree:blockio-sim
@@ -453,7 +454,7 @@ benchmark lsm-tree-bench-lookups
     , bytestring
     , deepseq
     , fs-api
-    , io-classes
+    , io-classes            >=1.5
     , lsm-tree
     , lsm-tree:blockio-api
     , lsm-tree:bloomfilter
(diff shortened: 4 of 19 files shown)


cabal.project
```diff
@@ -48,3 +48,16 @@ source-repository-package
...
+allow-newer:
+  fs-api:base,
+  fs-sim:base,
+  io-classes:base,
+  io-sim:base,
+  quickcheck-lockstep:base,
+  si-timers:base,
+  strict-mvar:base,
+  strict-stm:base,
+  unix-bytestring:base,
```
jorisdral, line 63: Let's make this conditional on the GHC version

```suggestion
-- TODO: this relaxing of dependency bounds on @base@ is currently required in order to
-- build the Haddock documentation with the @scripts/haddock.sh@ script.
if impl(ghc >=9.10)
  allow-newer:
    fs-api:base,
    fs-sim:base,
    io-classes:base,
    io-sim:base,
    quickcheck-lockstep:base,
    si-timers:base,
    strict-mvar:base,
    strict-stm:base,
    unix-bytestring:base,
```

jorisdral (commented):
First contribution! Nice. Some comments:

I'd expect that fewer `allow-newer` stanzas are required if we consider more recent versions of packages on `Hackage` and `CHaP` (a Cardano-specific Hackage instance fyi). There are probably already a few releases that support `ghc-9.10.1` among them.

```diff
diff --git a/cabal.project b/cabal.project
index cf07834..e265d56 100644
--- a/cabal.project
+++ b/cabal.project
@@ -12,11 +12,11 @@ repository cardano-haskell-packages
 
 index-state:
   -- Bump this if you need newer packages from Hackage
-  -- current date: crc32c-0.2.2
-  , hackage.haskell.org 2024-06-05T04:25:30Z
+  -- current date: support ghc-9.10.1
+  , hackage.haskell.org 2024-08-01T00:00:00Z
   -- Bump this if you need newer packages from CHaP
-  -- current date: fs-api-0.2.0.1, fs-sim-0.2.1.1
-  , cardano-haskell-packages 2023-11-30T09:59:24Z
+  -- current date: support ghc-9.10.1
+  , cardano-haskell-packages 2024-08-01T00:00:00Z
 
 packages: .
```

I'd also suggest this change to `.github/workflows/haskell.yml` to ensure our CI also builds/tests with `ghc-9.10.1`

```diff
diff --git a/.github/workflows/haskell.yml b/.github/workflows/haskell.yml
index f7f1045..41d94fd 100644
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["9.2.8", "9.4.8", "9.6.4", "9.8.2"]
+        ghc: ["9.2.8", "9.4.8", "9.6.4", "9.8.2", "9.10.1"]
         cabal: ["3.10.2.1"]
         os: [ubuntu-latest, windows-latest, macOS-latest]
         cabal-flags: [""]
@@ -36,6 +36,10 @@ jobs:
             os: windows-latest
           - ghc: "9.8.2"
             os: macOS-latest
+          - ghc: "9.10.1"
+            os: windows-latest
+          - ghc: "9.10.1"
+            os: macOS-latest
         include:
           - ghc: "9.6.4"
             cabal: "3.10.2.1"
```

bench/macro/lsm-tree-bench-bloomfilter.hs
```diff
@@ -13,7 +13,7 @@ import           Data.BloomFilter (Bloom)
 import qualified Data.BloomFilter as Bloom
 import qualified Data.BloomFilter.Hash as Bloom
 import qualified Data.BloomFilter.Mutable as MBloom
-import           Data.List (foldl')
+import qualified Data.List as List
```
phadej, line 16: This is actually "wrong". `import qualified Data.Foldable as Foldable`, as we use it on vector, not the list.

phadej (changes requested):
The comment applies to other occurrences of the import too

bench/macro/lsm-tree-bench-bloomfilter.hs
phadej: (alternatively, `Vector.foldl'`)

cabal.project
recursion-ninja: > Let's make this conditional on the GHC version

I forgot the `cabal.project` format permitted the conditional keywords. This is a good catch!

Still, I'll see if I can remove the need for most (hopefully all) of the dependencies in the `allow-newer` stanza by tweaking their upper bounds in the package description file.

recursion-ninja: > I'd also suggest this change to .github/workflows/haskell.yml to ensure our CI also builds/tests with ghc-9.10.1

I had refrained from updating the CI pipeline because I wasn't sure if the branch would be culled and didn't want to allocate excessive (probably overly cautious) resources in the CI pipeline. Since the changes are well received this makes sense. Thanks!

bench/macro/lsm-tree-bench-bloomfilter.hs
recursion-ninja: Really, I'm surprised it compiled, the tests ran, and the benchmarks ran if it was wrong. Maybe I misunderstand what you mean by "wrong" here? Regardless of any misunderstanding, I can easily swap the `foldl'` references to the polymorphic definition from `Data.Foldable`.

cabal.project
jorisdral: FYI, you probably don't have to update many bounds in the package description file, just the `index-state` stanzas (see https://github.com/IntersectMBO/lsm-tree/pull/319#pullrequestreview-2213903964)

bench/macro/lsm-tree-bench-bloomfilter.hs
mheinzel: `Data.List` exports the `Foldable`-polymorphic versions, so it compiles for vectors as well. It's just a weird/misleading place to get them from if you're not using them for lists.

recursion-ninja: Ah, got it. That makes sense. I recall the GHC warning that this will change in an upcoming version; i.e. that `Data.List.foldl'` will be monomorphic to lists.

cabal.project
recursion-ninja: I think it turned out to be a little bit more involved (see the new changes) than the index state since neither `fs-api` nor `fs-sim` have a version which supports GHC 9.10. I'm not sure why, but the updated `index-state` is not having the [IntersectMBO/cardano-haskell-packages](https://github.com/IntersectMBO/cardano-haskell-packages) repository forward the GHC 9.10 compatible `fs-api` source code from the [input-output-hk/fs-sim](https://github.com/input-output-hk/fs-sim) repository. I would like to see how much of the `fs-api:*` and `fs-sim:*` items can be removed from `allow-newer` if I can get the newest source code, but I'm not savvy enough to navigate the plumbing of the TOML files and cabal revisions.

lsm-tree.cabal
```diff
@@ -94,22 +94,22 @@ library
...
+    , base                  >=4.16      && <4.21
     , bitvec                ^>=1.1
     , bytestring            ^>=0.11.4.0 || ^>=0.12.1.0
     , containers
     , crc32c                ^>=0.2.1
     , deepseq               ^>=1.4      || ^>=1.5
     , filepath
     , fs-api                ^>=0.2
-    , io-classes            ^>=1.4
+    , io-classes            ^>=1.4      || ^>=1.5
```
phadej, line 105: is there a reason to allow older `io-classes`? If there isn't, it's better to drop it as CI won't test it. Ditto for other dependencies.

recursion-ninja: `io-classes` will build for all supported GHC versions if and only if we remove the conditional dependency relaxation from the `cabal.project` file:
```
-- if impl(ghc >=9.10)                                                                                          
```

Otherwise, the `fs-api` and `fs-sim` packages reject the newer version of `io-classes` specified in the package description.
```
$ cabal build --with-compiler=ghc-9.2.8
Resolving dependencies...
Error: [Cabal-7107]
Could not resolve dependencies:
[__0] trying: fs-api-0.2.0.1 (user goal)
[__1] next goal: io-classes (dependency of fs-api)
[__1] rejecting: io-classes-1.5.0.0 (conflict: fs-api => io-classes>=0.3 && <1.5)
[__1] trying: io-classes-1.4.1.0
[__2] next goal: lsm-tree (user goal)
[__2] rejecting: lsm-tree-0.1.0.0 (conflict: io-classes==1.4.1.0, lsm-tree => io-classes^>=1.5)
```

I can either leave the expanded version ranges in the package description, or relax the constraints in the `cabal.project` file until the upper bounds of `base` in `fs-api` and `fs-sim` get bumped.

recursion-ninja: For the time being I made the`allow-newer` stanza, which was previous conditional on GHC 9.10, now apply universally to all GHC version with the intention of removing the stanza after a new version of `fs-api` and `fs-sim` are accessible.

recursion-ninja: Never-mind, I didn't notice that the `fs-api` and `fs-sim` versions were explicitly pinned to a commit hash. I can update the commit hash to a newer version which supports GHC 9.10.

cabal.project
recursion-ninja: Never-mind, I didn't notice that the `fs-api` and `fs-sim` versions were explicitly pinned to a commit hash. I can update the commit hash to a newer version which supports GHC 9.10.

bench/macro/lsm-tree-bench-bloomfilter.hs
recursion-ninja: The requested changes have been made.

blockio-sim/src/System/FS/BlockIO/Sim.hs
recursion-ninja, line 1: This is required after changing the pinned dependency hash of `fs-api` and `fs-sim` as there was a name change of the function in the upstream API.

cabal.project
```diff
@@ -44,7 +44,13 @@ import: cabal.project.blockio-uring
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/fs-sim
-  tag: 47879aa5edfd3a3f8824d61687e85b8f1586e010
+  tag: 235d5dd10aadb70defa5a2e843a4cadf5d6eb18e
```
recursion-ninja, line 47: Updated to a newer hash of `fs-api` and `fs-sim` to permit building with `ghc-9.10.1` without an excessively large `allow-newer` stanza
.

test/Database/LSMTree/ModelIO/Monoidal.hs
recursion-ninja, line 1: An explicit import list was required to avoid a name clash with locally defined `writeTMVar`.

recursion-ninja (commented):
I think the PR looks ready to merge into `main`.

test/Database/LSMTree/ModelIO/Monoidal.hs
jorisdral: You can probably just remove the local one. It was only there because `writeTMVar` was missing from `io-classes`

src/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -230,7 +231,14 @@ data ByteCountDiscrepancy = ByteCountDiscrepancy {
...
 -- should pointwise match with the vectors of runs.
 lookupsIO ::
-     forall m h. (PrimMonad m, MonadThrow m, MonadST m)
+     forall m h.
+     (
+#if MIN_VERSION_io_classes(1,5,0)
+-- Since io-classes-1.5.0.0 the MonadST type-class constraint provides MonadPrim m => MonadST m
+#else
+     PrimMonad m,
+#endif
```
jorisdral, line 240: The CPP is redundant, since we're always using `io-classes-1.5.0.0`

jorisdral (approved):
LGTM. Let's address the two new comments I inculded, and then we can merge

src/Database/LSMTree/Internal/Lookup.hs
recursion-ninja: > You can probably just remove the local one. It was only there because writeTMVar was missing from io-classes

I removed the local definition of `writeTMVar` in favor of the exported definition from `io-classes`.

> The CPP is redundant, since we're always using io-classes-1.5.0.0

An excellent observation. It is best to remove CPP usage wherever possible. I have indeed removed the superfluous usage in the `Database.LSMTree.Internal.Lookup` module.

dcoutts (approved):
LGTM. Make sure the "merge branch main" patch gets squashed into the main patch, before the PR is merged.

PR #322, opened 2024-08-05
Add configuration option for the fence pointer index

jorisdral:
The configuration option currently has no effect: the Compact Index is always used.


dcoutts (approved):
LGTM!

PR #331, opened 2024-08-09
Correcting false positive of HLint passing GitHub workflow

recursion-ninja:
Our GitHub action for running HLint on the code was previously emitting an alarming warning:
```
  Running hlint -j --json -- src/ test/
  Warning: Can't find file "MachDeps.h" in directories
  	src/Database/LSMTree/Internal
  	.
    Asked for by: src/Database/LSMTree/Internal/RawBytes.hs  at line 8 col 1
  hlint completed with status code 0
```

The `haskell.yml` workflow file's entry for the HLint action had previously specified only `fail-on: suggestion` whereas this PR increases the failure condition to `fail-on: warning`. This change causes our CI pipeline to have caught and require correcting the previously ignored warning.

Additionally, the PR corrects the warning by conditionally importing the offending file.

### DIFF
--- a/bench/macro/lsm-tree-bench-wp8.hs
+++ b/bench/macro/lsm-tree-bench-wp8.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE OverloadedStrings     #-}
-
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 {- Benchmark requirements:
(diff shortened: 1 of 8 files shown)


recursion-ninja:
While reviewing the subject of HLint, is there a reason we only include the `src` and `test` directories as part of the HLint coverage action and exclude the following directories from HLint coverage?

- `bench`
- `blockio-api`
- `blockio-sim`
- `bloomfilter`
- `prototypes`
- `src-extras`
- `src-fcntl-nocache`
- `src-kmerge`
- `src-mcg`
- `src-monkey`
- `src-rocksdb`
- `xxhash`

jorisdral (approved):
Looks good

mheinzel:
The 9.10 job didn't run because it didn't exist before, but it's required now. Maybe just rebase and push again to trigger CI and the auto-merge?

recursion-ninja:
I `rebase`ed, so the `ghc-9.10` checks should run now.

I added all the `src-*` directories to the HLint checks.

I *not not add* the `bench` directory to the HLint checks because there are many suggestions and warnings raised by HLint which would change the evaluation semantics of the benchmarks in undesirable ways. Essentially, within the `bench` directory, HLint gives too many "false positives."

recursion-ninja:
> Instead of adding `{-# HLINT ignore ... #-}`, ignore the warnings in the `hlint.yaml` file

I removed the `{-# HLINT ignore ... #-}` pragmas from all the modules and I added the corresponding `ignore` statements from the `src-*` directories to the `hlint.yaml` file. I did *not* add the `{-# HLINT ignore ... #-}` pragmas from the `bench` directory to the `hlint.yaml` file since those were related to evaluation semantics false positives.

I think the PR is ready to be merged.

phadej (dismissed):

bench/macro/lsm-tree-bench-wp8.hs
```diff
@@ -392,20 +391,20 @@ doDryRun gopts opts = do
 doDryRun' :: GlobalOpts -> RunOpts -> IO ()
 doDryRun' gopts opts = do
     -- calculated some expected statistics for generated batches
-    id $ do
```
phadej, line 395: the block makes new scope for `n`, `d` etc. See warnings below. It's  the same trick as 

```c
int main() {

   {
     // new scope in C
     
   }
```

though in C++ one can also use it for RAII.

---

Maybe the `id` is not needed, so you could write just `

```
doDryRun' gopts opts = do

   do
       -- inner scope
       
```

recursion-ninja: I applied your suggestions. Thanks for elucidating "nested `do`-block" usage. Now that I'm better informed, I think you are right to leave the scoping limitation. I added a comment on the matter in case other eyes find the code confusing.

PR #332, opened 2024-08-09
Performing TODO: Create newtype

recursion-ninja:
This PR performs the TODO task of changing the type-synonym `NumPages` to a proper `newtype`.

The `NumPages` is now a `newtype` wrapper of the `Word` data-type, as there cannot ever be a negative number of pages.

Additionally, I added documentation for `NumPages` and corrected a few typos in the module's existing documentation.

### DIFF
--- a/src/Database/LSMTree/Internal/IndexCompact.hs
+++ b/src/Database/LSMTree/Internal/IndexCompact.hs
@@ -395,7 +395,7 @@ newtype NumPages = NumPages Word
   deriving newtype (NFData)
 
 -- | A type-safe "unwrapper" for 'NumPages'. Use this accessor whenever you want
--- to convert 'NumPages' to a more versitile number type.
+-- to convert 'NumPages' to a more versatile number type.
 {-# INLINE getNumPages #-}
 getNumPages :: Integral i => NumPages -> i
 getNumPages (NumPages w) = fromIntegral w
--- a/src/Database/LSMTree/Internal/IndexCompact.hs
+++ b/src/Database/LSMTree/Internal/IndexCompact.hs
@@ -7,6 +7,7 @@ module Database.LSMTree.Internal.IndexCompact (
     IndexCompact (..)
   , PageNo (..)
   , NumPages
+  , getNumPages
     -- * Queries
   , PageSpan (..)
   , singlePage
@@ -69,7 +70,7 @@ import           Database.LSMTree.Internal.Vector
 
   Given a serialised target key @k@, an index can be 'search'ed to find a disk
   page @i@ that /might/ contain @k@. Fence-pointer indices offer no guarantee of
-  whether the page contains the key, but the indices do guarentee that no page
+  whether the page contains the key, but the indices do guarantee that no page
   other than page @i@ could store @k@.
 
   === Intro
@@ -143,7 +144,7 @@ import           Database.LSMTree.Internal.Vector
 
   The intuition behind the compact index is this: often, we don't need to look
   at full bit-strings to compare independently generated hashes against
-  eachother. The probability of the @n@ most significant bits of two
+  each other. The probability of the @n@ most significant bits of two
   independently generated hashes matching is @(1/(2^n))@, and if we pick @n@
   large enough then we can expect a very small number of collisions. More
   generally, the expected number of @n@-bit hashes that have to be generated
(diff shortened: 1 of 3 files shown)


src/Database/LSMTree/Internal/IndexCompact.hs
```diff
@@ -385,8 +385,13 @@ newtype PageNo = PageNo { unPageNo :: Int }
...
 
--- TODO: Turn into newtype
-type NumPages = Int
+-- | The number of pages contained by an index or other paging data-structure.
+--
+-- Note: This is a 0-based number; take care to ensure arithmetic underflow
+-- does not occur during subtraction operations!
+newtype NumPages = NumPages Word
+  deriving stock (Eq, Ord, Show)
+  deriving newtype (Enum, Integral, NFData, Num, Real)
```
mheinzel, line 394: I would argue that providing `Num`, `Real` and `Enum` instances removes a significant part of the advantages of having a newtype. They provide a lot of operations that don't make sense for this type, like `(*) :: NumPages -> NumPages -> NumPages` and can make conversions "too easy". It's safer to be more explicit, manually defining the small number of required operations, which are hopefully not too many. This of course might require more changes to the places where `NumPages` is used, but IMO it's often worth it.

mheinzel: But generally I think it makes sense that you decided to work on this! It's a good idea to get more familiar with the code base by resolving TODOs and improving code quality.

jorisdral: I concur -- let's remove the instances and wrap/unwrap the newtype where necessary

recursion-ninja: I updated the PR by removing the permissive type-class instances and adds `getNumPages` as a polymorphic "unwrapper function."

src/Database/LSMTree/Internal/IndexCompact.hs
```diff
@@ -385,8 +386,22 @@ newtype PageNo = PageNo { unPageNo :: Int }
...
+newtype NumPages = NumPages Word
+  deriving stock (Eq, Ord, Show)
+  deriving newtype (NFData)
+
+-- | A type-safe "unwrapper" for 'NumPages'. Use this accessor whenever you want
+-- to convert 'NumPages' to a more versitile number type.
+{-# INLINE getNumPages #-}
+{-# SPECIALIZE getNumPages :: NumPages -> Int #-}
+{-# SPECIALIZE getNumPages :: NumPages -> Word #-}
+{-# SPECIALIZE getNumPages :: NumPages -> Word64 #-}
```
jorisdral, line 402: Just the `INLINE` should be sufficient for specialisation to happen at the use site (in most cases)

jorisdral (approved):
LGTM. Just one last comment, and then we can merge

src/Database/LSMTree/Internal/IndexCompact.hs
jorisdral: Also, AFAIK the `INLINE` pragma doesn't apply to the specialised versions of the function

recursion-ninja: I wasn't sure so I leaned towards the side of caution. Is there a downside to leaving the specializations (larger object file)?

recursion-ninja: I removed the `SPECIALIZE` statements.

src/Database/LSMTree/Internal/IndexCompact.hs
```diff
@@ -385,8 +386,19 @@ newtype PageNo = PageNo { unPageNo :: Int }
...
+-- | The number of pages contained by an index or other paging data-structure.
+--
+-- Note: This is a 0-based number; take care to ensure arithmetic underflow
+-- does not occur during subtraction operations!
+newtype NumPages = NumPages Word
+  deriving stock (Eq, Ord, Show)
+  deriving newtype (NFData)
+
+-- | A type-safe "unwrapper" for 'NumPages'. Use this accessor whenever you want
+-- to convert 'NumPages' to a more versitile number type.
```
mheinzel, line 398:
```suggestion
-- to convert 'NumPages' to a more versatile number type.
```

mheinzel (approved):

PR #333, opened 2024-08-12
Remove Totality requirement for ResolveValue

mheinzel:
We can revisit this once we focus on robustness, including error handling during run merges.

### DIFF
--- a/src/Database/LSMTree/Monoidal.hs
+++ b/src/Database/LSMTree/Monoidal.hs
@@ -99,7 +99,6 @@ module Database.LSMTree.Monoidal (
   , resolveDeserialised
     -- ** Properties
   , resolveValueValidOutput
-  , resolveValueTotality
   , resolveValueAssociativity
 
     -- * Utility types
@@ -111,9 +110,9 @@ import           Control.Monad (void, (<$!>))
 import           Data.Bifunctor (Bifunctor (..))
 import           Data.Coerce (coerce)
 import           Data.Kind (Type)
+import           Data.Monoid (Sum (..))
 import           Data.Typeable (Proxy (Proxy), Typeable)
 import qualified Data.Vector as V
-import           Data.Word (Word64)
 import           Database.LSMTree.Common (IOLike, Range (..), SerialiseKey,
                      SerialiseValue (..), Session (..), SnapshotName,
                      closeSession, deleteSnapshot, listSnapshots, openSession,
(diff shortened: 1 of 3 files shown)


src/Database/LSMTree/Monoidal.hs
jorisdral, line 1: It might be better to move the `ResolveValue Word64` instance to the test suite, since it might/will conflict with user instance. Or, alternatively, you can use a newtype in the style of `Data.Monoid.Sum` as an example, and it might be useful as a modifier as well

jorisdral (changes requested):
LGTM. Small but unrelated comment below

src/Database/LSMTree/Monoidal.hs
mheinzel: I initially also wanted to provide an instance for `Sum`, but it's trickier than I thought. `resolveDeserialised` requires an instance `SerialiseValue (Sum a)` (and I didn't want to coerce things based on `SerialiseValue a` because the instances might disagree), but I'm not sure we generally want to have `instance SerialiseValue a => SerialiseValue (Sum a)`. But if we don't, users would have to define an orphan instance for e.g. `SerialiseValue (Sum Foo)`, which seems wrong.

mheinzel: I can provide instances `SerialiseValue (Sum Word64)` and  `ResolveValue (Sum Word64)` specifically, maybe that's an okay compromise?

recursion-ninja: I added the `SerialiseValue (Sum Word64)` and `ResolveValue (Sum Word64)` instances. I squashed the commit and re-based with `main` in anticipation of merging.

recursion-ninja: Per a discussion with @mheinzel:

Regarding the instance `SerialiseValue a => SerialiseValue (Sum a)`, it essentially states that the `Sum` newtype is transparent to serialization. This assertion is not unreasonable, as it is highly improbable one would serialize `Sum a` differently than `a`.

Hence we can have the more useful polymorphic instance: 
```
instance (Num a, SerialiseValue a) => ResolveValue (Sum a)
```

test/Test/Database/LSMTree/Monoidal.hs
```diff
@@ -1,35 +1,32 @@
...
 
+-- | An example /(orphan)/ instance used to facilitate the property tests.
+instance ResolveValue Word64 where
+  resolveValue = resolveDeserialised (+)
+
 tests :: TestTree
 tests = testGroup "Test.Database.LSMTree.Monoidal"
-    [ testGroup "Word64" (allProperties @Word64 False)
-      -- TODO: revisit totality (drop requirement or fix @SerialiseValue Word64@)
+    [ testGroup "Word64" (allProperties @Word64)
```
mheinzel, line 19: Ah, I thought we would just use the existing instance for `ResolveValue (Sum Word64)` instead of defining an orphan instance.

mheinzel (commented):
One last detail, then we can merge.

PR #334, opened 2024-08-13
Replace `IORef` by `MutVar`

jorisdral:


src/Database/LSMTree/Internal.hs
```diff
@@ -787,14 +787,14 @@ flushWriteBuffer conf@TableConfig{confDiskCachePolicy}
 -- | Note that the invariants rely on the fact that levelling is only used on
 -- the last level.
 --
-levelsInvariant :: forall s h. TableConfig -> Levels h -> ST s Bool
+levelsInvariant :: forall s1 s2 h. TableConfig -> Levels s2 h -> ST s1 Bool
```
phadej, line 790: this doesnt look right, why different state threads?

jorisdral: The current invariant check doesn't actually use any mutable references, but it will once we add merge scheduling, so I'll change it to account for that in this PR.

jorisdral:
BTW, I'm running the micro-benchmarks locally to see if the changes lead to a regression

jorisdral:
> BTW, I'm running the micro-benchmarks locally to see if the changes lead to a regression

Okay, as far as I can see there is no noticeable regression.

src/Database/LSMTree/Internal.hs
```diff
@@ -868,13 +869,16 @@ addRunToLevels ::
...
-  -> m (Levels (Handle h))
+  -> Levels (PrimState m) (Handle h)
+  -> m (Levels (PrimState m) (Handle h))
 addRunToLevels conf@TableConfig{..} resolve hfs hbio root uniqC r0 reg levels = do
     ls' <- go (LevelNo 1) (V.singleton r0) levels
-    assert (runST $ levelsInvariant conf ls') $ return ls'
+#ifdef NO_IGNORE_ASSERTS
+    void $ stToIO $ _levelsInvariant conf ls'
+#endif
+    return ls'
```
mheinzel, line 881: Interesting, what is the advantage of using CPP here?

Is it because the invariant is not really a predicate (always returning `True`), so it would be weird to use `assert`? In that case, we could also change it to return `()` now and get rid of the `void`.

mheinzel (approved):
The type parameter spreads into quite a lot of types, but we will need this for `io-sim`, right?

src/Database/LSMTree/Internal.hs
jorisdral: The current `levelsInvariant` doesn't actually read mutable variables, so we can run it purely, but once we add ongoing merges we'll have to read mutable variables to inspect the state of an ongoing merge. As such, we can't use `runST` anymore, and we have to instead embed the invariant check into `IO`

jorisdral:
> The type parameter spreads into quite a lot of types, but we will need this for `io-sim`, right?

Yes, we'll need either a state parameter `s` or a monad parameter `m`. For now, we're sticking with `s`

src/Database/LSMTree/Internal.hs
mheinzel: And reading the variable won't be possible in ST, like with `MutVar`? Okay, we'll see what it looks like then.

jorisdral: The problem is that `runST` is polymorphic over `s`, while in this `addRunToLevels` function we have `s ~ RealWorld` (or more precisely, `s ~ PrimState m`), so we have to embed using `stToIO`

PR #344, opened 2024-08-21
Define cursor API

mheinzel:
# Description

Defines the public API for cursors. The implementation will follow soon, but I decided to split the work into small PRs, so individual aspects can be reviewed and merged separately. For example, based on this PR, one could already start integrating the new cursor API into the quickcheck-lockstep tests.

I decided to rename RangeLookupResult, since it will be used for cursor reads, too. Generally it is useful for returning entries when their keys are not already known, which is everything except for (point) lookups.

# Checklist

### DIFF
--- a/src/Database/LSMTree/Normal.hs
+++ b/src/Database/LSMTree/Normal.hs
@@ -65,6 +65,12 @@ module Database.LSMTree.Normal (
   , rangeLookup
   , Range (..)
   , QueryResult (..)
+    -- ** Cursor
+  , Cursor
+  , withCursor
+  , newCursor
+  , closeCursor
+  , readCursor
     -- ** Updates
   , inserts
   , deletes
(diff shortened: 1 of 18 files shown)


src/Database/LSMTree/Normal.hs
jorisdral, line 1: We should extend/update the `$resourcemanagement`, `$exception-safety`, and `$concurrency` documentation sections.

jorisdral (changes requested):
Just some small things to address, and then we can merge

test/Test/Database/LSMTree/Class/Monoidal.hs
```diff
@@ -217,8 +217,8 @@ evalRange :: Ord k => Range k -> k -> Bool
 evalRange (FromToExcluding lo hi) x = lo <= x && x < hi
 evalRange (FromToIncluding lo hi) x = lo <= x && x <= hi
 
-rangeLookupResultKey :: RangeLookupResult k v -> k
-rangeLookupResultKey (FoundInRange k _)             = k
+rangeLookupResultKey :: QueryResult k v -> k
+rangeLookupResultKey (FoundInQuery k _) = k
```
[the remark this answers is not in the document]
mheinzel: Ah, yes, I didn't grep for the lower case version. Thanks!

src/Database/LSMTree/Normal.hs
mheinzel: I don't know how helpful it would be to add "and cursors" every time the docs talk about table handles there. Maybe in the beginning once say "(and similarly for cursors)" or at the end have a short sentence that all of this also applies to cursors?

jorisdral: True, there is some duplication there, but I think it's minor

dcoutts (commented):
API looks good.

PR #345, opened 2024-08-21
Implement cursor type, new, close

mheinzel:
# Description

Defines `Cursor` and implements `newCursor` and `closeCursor`, with a focus on resource management. The next PR will implement `readCursor`.

# Checklist

### DIFF
--- a/src/Database/LSMTree/Internal.hs
+++ b/src/Database/LSMTree/Internal.hs
@@ -27,6 +27,11 @@ module Database.LSMTree.Internal (
   , close
   , lookups
   , updates
+    -- ** Cursor API
+  , Cursor (..)
+  , withCursor
+  , newCursor
+  , closeCursor
     -- * Snapshots
   , SnapshotLabel
   , snapshot
@@ -72,6 +77,7 @@ import qualified Database.LSMTree.Internal.Paths as Paths
 import           Database.LSMTree.Internal.Run (Run)
 import qualified Database.LSMTree.Internal.Run as Run
 import           Database.LSMTree.Internal.RunNumber
+import qualified Database.LSMTree.Internal.RunReaders as Readers
 import           Database.LSMTree.Internal.Serialise (SerialisedBlob,
                      SerialisedKey, SerialisedValue)
 import           Database.LSMTree.Internal.TempRegistry
@@ -99,14 +105,17 @@ data LSMTreeError =
     -- | The session directory is malformed: the layout of the session directory
     -- contains unexpected files and/or directories.
   | SessionDirMalformed FsErrorPath
-    -- | All operations on a closed session or tables within a closed session
-    -- will throw this exception. There is one exception (pun intended) to this
-    -- rule: the idempotent operation 'Database.LSMTree.Common.closeSession'.
+    -- | All operations on a closed session as well as tables or cursors within
+    -- a closed session will throw this exception. There is one exception (pun
+    -- intended) to this rule: the idempotent operation
+    -- 'Database.LSMTree.Common.closeSession'.
   | ErrSessionClosed
-    -- | All operations on a closed table will throw this exception. There is
-    -- one exception (pun intended) to this rule: the idempotent operation
-    -- 'Database.LSMTree.Common.close'.
+    -- | All operations on a closed table will throw this exception, except for
+    -- the idempotent operation 'Database.LSMTree.Common.close'.
   | ErrTableClosed
+    -- | All operations on a closed cursor will throw this exception, except for
+    -- the idempotent operation 'Database.LSMTree.Common.closeCursor'.
+  | ErrCursorClosed
   | ErrSnapshotExists SnapshotName
   | ErrSnapshotNotExists SnapshotName
   | ErrSnapshotWrongType SnapshotName
@@ -202,6 +211,9 @@ data SessionEnv m h = SessionEnv {
     --   tables without restrictions, even concurrently with 'closeSession'.
     --   This is safe because 'close' is idempotent'.
   , sessionOpenTables  :: !(StrictMVar m (Map Word64 (TableHandle m h)))
+    -- | Similarly to tables, open cursors are tracked so they can be closed
+    -- once the session is closed. See 'sessionOpenTables'.
+  , sessionOpenCursors :: !(StrictMVar m (Map Word64 (Cursor m h)))
   }
 
 -- | 'withOpenSession' ensures that the session stays open for the duration of the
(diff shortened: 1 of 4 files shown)


src/Database/LSMTree/Internal.hs
```diff
@@ -439,20 +453,24 @@ closeSession Session{sessionState, sessionTracer} = do
...
-        -- Close tables first, so that we know none are open when we release
-        -- session-wide resources.
-        --
-        -- If any table from this set has been closed already by a different
-        -- thread, the idempotent 'close' will act like a no-op, and so we are
-        -- not in trouble.
-        mapM_ close tables
+        modifyMVar_ (sessionOpenTables seshEnv) $ \tables -> do
+          mapM_ close tables
+          pure Map.empty
```
mheinzel, line 470: I think this is slightly nicer, since if there is an error while closing tables, `sessionOpenTables` will keep the old values instead of being empty. Better to close twice than not at all.

src/Database/LSMTree/Internal.hs
```diff
@@ -1157,6 +1175,146 @@ mergeRuns resolve hfs hbio caching alloc runPaths mergeLevel runs = do
...
+        -- (same for closing table handles)
+
+        -- Since we have a write lock on the table state, we know that we are
+        -- the only thread currently closing the table. We can safely make the
+        -- session forget about this table.
+        modifyMVar_ (sessionOpenCursors cursorSessionEnv) $
+          pure . Map.delete cursorId
+        Readers.close hfs hbio cursorReaders
+        V.mapM_ (Run.removeReference hfs hbio) cursorRuns
+        return CursorClosed
```
mheinzel, line 1316: This now works the same as table handle close, but it doesn't seem right. Still thinking about it. Thoughts?

src/Database/LSMTree/Normal.hs
```diff
@@ -315,7 +315,8 @@ rangeLookup = undefined
 -- Once a cursor has been created, updates to the referenced table don't affect
 -- the cursor.
 type Cursor :: (Type -> Type) -> Type -> Type -> Type -> Type
-data Cursor m k v blob
+-- TODO: tweak? needs Typeable?
+data Cursor m k v blob = forall h. Typeable h => Cursor !(Internal.Cursor m h)
```
mheinzel, line 319: Same as for `TableHandle` and `BlobRef`. Do we currently use `Typeable` for anything there? Is the idea that we can later want to add runtime assertions?

src/Database/LSMTree/Internal.hs
```diff
@@ -1157,6 +1175,146 @@ mergeRuns resolve hfs hbio caching alloc runPaths mergeLevel runs = do
...
+    -- 'sessionOpenTables'.
+    withOpenSession cursorSession $ \_ -> do
+      -- only put the real state once creation completed successfully
+      cursorState <- newMVar CursorClosed
+      let !cursor = Cursor cursorState
+      -- Make sure that if now an exception is raised (and all resources get
+      -- freed by the registry) the cursor state remains 'CursorClosed',
+      -- so the session (which might track the cursor) doesn't later find it
+      -- open and double frees resources.
+      modifyWithTempRegistry (pure CursorClosed) (void <$> swapMVar cursorState) $ \reg _ -> do
```
mheinzel, line 1254: Does the problem become clear? AFAICT the implementation of  `duplicate` is unsafe in that regard.

I had another approach before, which used `tempFree` instead of `modifyWithTempRegistry`, see commit "other approach to resource management".

src/Database/LSMTree/Internal.hs
mheinzel: Ah, but it causes a deadlock, since `close` wants to remove the handle from `sessionOpenTables`: https://github.com/IntersectMBO/lsm-tree/actions/runs/10489456896/job/29054138188?pr=345#step:16:1583

mheinzel: Putting back closed tables is not bad, as `close` is idempotent. But I agree that this generally needs `TempRegistry`.

src/Database/LSMTree/Internal.hs
mheinzel: Improved it, although probably still not perfect.

src/Database/LSMTree/Normal.hs
mheinzel: We're going to need it then, since reading entries from the cursor needs to wrap the `BlobRefs`. We can remove it in `Monoidal`, though, for TableHandle as well. I'll do that.

src/Database/LSMTree/Internal.hs
jorisdral: It can lead to a memory leak for long-lived programs, because only `sessionClose` would get rid of the closed table from `sessionOpenTables`. It's a contrived scenario, but probably one we should avoid

src/Database/LSMTree/Internal.hs
```diff
@@ -1157,6 +1176,145 @@ mergeRuns resolve hfs hbio caching alloc runPaths mergeLevel runs = do
...
+            freeTableContent hfs hbio reg cursorRuns
+            Cursor <$> newMVar CursorDrained
+
+          Just cursorReaders -> do
+            cursorId <- incrUniqCounter (sessionUniqCounter cursorSessionEnv)
+            cursor <- Cursor <$> newMVar (CursorOpen CursorEnv {..})
+            -- Track cursor, but careful: If now an exception is raised, all
+            -- resources get freed by the registry, so if the session still
+            -- tracks 'cursor' (which is 'CursorOpen'), it later double frees.
+            -- Therefore, we only track the cursor if 'withOpenSession' exits
```
jorisdral, line 1268:
```suggestion
            -- Therefore, we only track the cursor if 'withTempRegistry' exits
```

src/Database/LSMTree/Internal.hs
```diff
@@ -1157,6 +1176,145 @@ mergeRuns resolve hfs hbio caching alloc runPaths mergeLevel runs = do
...
+
+        mReaders <-
+          allocateMaybeTemp reg
+            (Readers.new hfs hbio (V.toList cursorRuns))
+            (Readers.close hfs hbio)
+
+        case mReaders of
+          Nothing -> do
+            freeTableContent hfs hbio reg cursorRuns
+            Cursor <$> newMVar CursorDrained
```
jorisdral, line 1260: It would be nice if we can release runs early when the cursor is drained, but I think this code is at odds with how blob references work. If a blob reference is returned by a cursor read, but the cursor is drained in the process, then the blob references might be immediately invalidated if the cursor had the only reference to the runs. Maybe we should always let the user explicitly close the cursor to reduce reference counts on the runs

jorisdral (changes requested):
Looks good so far. I think we should address the conflict with blob references that the current version has, and then we can probably merge

src/Database/LSMTree/Internal.hs
mheinzel: That's a good point. I'll still mark the cursor as drained, but it should keep holding the runs open.

jorisdral (approved):

PR #348, opened 2024-08-22
Remove `GetUniq` in favour of `UniqCounter`

jorisdral:


dcoutts (approved):
LGTM. Just replacing an interface (action to return a unique) with a concrete data type and known function.

PR #351, opened 2024-08-22
Fix #350 by properly cleaning up RunBuilders in tests

### LINKED ISSUE #350: [BUG] `RunBuilder` tests take multiple minutes on Windows
**Describe the bug**
`RunBuilder` tests take multiple minutes on Windows

**Expected behavior**
Running times for the `RunBuilder` tests should be the same on Linux and MacOS as on Windows.

**Output**
CI output: https://github.com/IntersectMBO/lsm-tree/actions/runs/10515089091/job/29134408774#step:16:1220

 - OS: Windows
 - GHC: all

jorisdral:
# Description

Resolves #350

# Checklist

### DIFF
--- a/lsm-tree.cabal
+++ b/lsm-tree.cabal
@@ -364,6 +364,7 @@ test-suite lsm-tree-test
     , semialign
     , split
     , stm
+    , strict-stm
     , tasty
     , tasty-hunit
     , tasty-quickcheck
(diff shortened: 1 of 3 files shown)


mheinzel:
If we ran these tests in a MockFS, we could assert that all handles get closed properly. Would be sensible to do, maybe in addition to the real FS ones?

jorisdral:
> If we ran these tests in a MockFS, we could assert that all handles get closed properly. Would be sensible to do, maybe in addition to the real FS ones?

Good idea. I've added the check. Maybe we can use `withSimHasFS` in other places as well

dcoutts (commented):
Looks ok but in the first patch I don't see what is the fix for #350. What actual change are we making, why does that fix the running time problem on Windows?

jorisdral:
> Looks ok but in the first patch I don't see what is the fix for #350. What actual change are we making, why does that fix the running time problem on Windows?

I'm not so sure why the fix works, but what is clear is that the `RunBuilder` tests were leaking file descriptors, and it *looks* like that leak was causing a slowdown (or it's a spurious correlation, I don't really have a way to check)

dcoutts (approved):

PR #359, opened 2024-08-27
Start using `quickcheck-lockstep-0.5.1`

jorisdral:


dcoutts (approved):
LGTM. It will probably need conflicts resolving with #357.

PR #361, opened 2024-08-30
Add run number field to TableContent

phadej:
Also add TODO to change the reprentation of closed table handle these should not need to have any `TableContent`.

### DIFF
--- a/src/Database/LSMTree/Internal.hs
+++ b/src/Database/LSMTree/Internal.hs
@@ -528,6 +528,7 @@ newWith ::
   -> m (TableHandle m h)
 newWith sesh seshEnv conf !am !wb !levels = do
     tableId <- incrUniqCounter (sessionUniqCounter seshEnv)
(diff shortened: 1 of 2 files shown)


dcoutts (approved):
Looks reasonable to me.

And I agree it'd be nice for a closed table to contain less.

jorisdral:
I'm not sure I understand the TODO, is the current definition of `TableHandleState` not already doing what the TODO describes?

phadej:
@jorisdral maybe, but not 100% as emptyTableContent is used. See https://github.com/IntersectMBO/lsm-tree/blob/6d075bd19af82be96c63a39e705448c4affcd21a/src/Database/LSMTree/Internal.hs#L552

jorisdral (approved):
LGTM as well -- just needs a rebase to fix a merge conflict

dcoutts (approved):

phadej:
Rebased, make the field strict (now as its possible).

PR #362, opened 2024-08-30
Add tests for the ordinary index

jeltsch:
This adds tests for the ordinary index.

Currently, the property `prop_searchForUnmentionedKeyBeyondRangeWorks` tends to discard too many test cases. I will address this and then drop the draft status of this pull request.

### DIFF
--- a/src/Database/LSMTree/Internal/IndexOrdinary.hs
+++ b/src/Database/LSMTree/Internal/IndexOrdinary.hs
@@ -4,6 +4,7 @@
 module Database.LSMTree.Internal.IndexOrdinary
 (
     IndexOrdinary (IndexOrdinary),
+    toLastKeys,
 
     -- * Search
     search,
@@ -57,6 +58,10 @@ supportedTypeAndVersion = 0x0101
     to return a valid page span.
 -}
 newtype IndexOrdinary = IndexOrdinary (Vector SerialisedKey)
+    deriving stock (Eq, Show)
+
+toLastKeys :: IndexOrdinary -> Vector SerialisedKey
+toLastKeys (IndexOrdinary lastKeys) = lastKeys
 
 -- * Search
 
@@ -89,7 +94,7 @@ search key (IndexOrdinary lastKeys) = assert (length lastKeys > 0) result where
                 = let
 
                       start :: Int
-                      !start = maybe 0 succ $
+                      !start = maybe 0 succ                  $
                                findIndexR (/= last lastKeys) $
                                lastKeys
 
--- a/test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
+++ b/test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
@@ -1,0 +1,519 @@
+{-# LANGUAGE MagicHash #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+{- HLINT ignore "Avoid restricted alias" -}
+
+module Test.Database.LSMTree.Internal.IndexOrdinary (tests) where
+
+import           Prelude hiding (all, head, last, length, notElem, splitAt,
+                     tail, takeWhile)
+
+import           Control.Arrow (first, (>>>))
+import           Control.Monad.ST.Strict (runST)
+import           Data.ByteString.Short (ShortByteString (SBS))
+import qualified Data.ByteString.Short as ShortByteString (length, pack)
+import           Data.Either (isLeft)
+import           Data.List (genericReplicate)
+import qualified Data.List as List (tail)
+import           Data.Maybe (catMaybes)
+import           Data.Primitive.ByteArray (ByteArray (ByteArray), ByteArray#)
+import           Data.Vector (Vector, all, fromList, head, last, length,
+                     notElem, splitAt, tail, takeWhile, toList, (!))
+import qualified Data.Vector.Primitive as Primitive (Vector (Vector), concat,
+                     force, length, singleton)
+import           Data.Word (Word16, Word32, Word64, Word8)
+import           Database.LSMTree.Extras.Generators (LogicalPageSummaries,
+                     toAppends)
+import           Database.LSMTree.Internal.Chunk (fromChunk)
+import           Database.LSMTree.Internal.Entry (NumEntries (NumEntries))
+import           Database.LSMTree.Internal.IndexCompact (PageNo (PageNo),
+                     PageSpan (PageSpan))
+import           Database.LSMTree.Internal.IndexCompactAcc
+                     (Append (AppendMultiPage, AppendSinglePage))
+import           Database.LSMTree.Internal.IndexOrdinary
+                     (IndexOrdinary (IndexOrdinary), fromSBS, search,
+                     toLastKeys)
+import           Database.LSMTree.Internal.IndexOrdinaryAcc (append, new,
+                     unsafeEnd)
+import           Database.LSMTree.Internal.Serialise
+                     (SerialisedKey (SerialisedKey'))
+import           Database.LSMTree.Internal.Vector (byteVectorFromPrim)
+import           Test.Database.LSMTree.Internal.Chunk ()
+import           Test.QuickCheck (Arbitrary (arbitrary, shrink), Gen,
+                     NonNegative (NonNegative), Property,
+                     Small (Small, getSmall), Testable, chooseInt,
+                     counterexample, frequency, getPositive, getSorted,
+                     shrinkMap, suchThat, vector, (.&&.), (===), (==>))
+import           Test.QuickCheck.Instances.ByteString ()
+                     -- for @Arbitrary ShortByteString@
+import           Test.Tasty (TestTree, testGroup)
+import           Test.Tasty.QuickCheck (testProperty)
+
+-- * Tests
+
+tests :: TestTree
+tests = testGroup "Test.Database.LSMTree.Internal.IndexOrdinary" $
+        [
+            testGroup "Search" $
+            [
+                testProperty
+                    "Search for mentioned key works"
+                    prop_searchForMentionedKeyWorks,
+                testProperty
+                    "Search for unmentioned key in range works"
+                    prop_searchForUnmentionedKeyInRangeWorks,
+                testProperty
+                    "Search for unmentioned key beyond range works"
+                    prop_searchForUnmentionedKeyBeyondRangeWorks
+            ],
+            testGroup "Deserialisation" $
+            [
+                testProperty
+                    "Number of entries from serialised index works"
+                    prop_numberOfEntriesFromSerialisedIndexWorks,
+                testProperty
+                    "Index from serialised index works"
+                    prop_indexFromSerialisedIndexWorks,
+                testProperty
+                    "Too short input makes deserialisation fail"
+                    prop_tooShortInputMakesDeserialisationFail,
+                testProperty
+                    "Type-and-version error makes deserialisation fail"
+                    prop_typeAndVersionErrorMakesDeserialisationFail,
+                testProperty
+                    "Partial key size block makes deserialisation fail"
+                    prop_partialKeySizeBlockMakesDeserialisationFail,
+                testProperty
+                    "Partial key block makes deserialisation fail"
+                    prop_partialKeyBlockMakesDeserialisationFail
+            ],
+            testGroup "Incremental construction" $
+            [
+                testProperty
+                    "Incremental index construction works"
+                    prop_incrementalIndexConstructionWorks,
+                testProperty
+                    "Incremental serialised key list construction works"
+                    prop_incrementalSerialisedKeyListConstructionWorks
+            ]
+        ]
+
+-- * Tested type and version
+
+-- The type and version of the index to which these tests refer.
+testedTypeAndVersion :: Word32
+testedTypeAndVersion = 0x0101
+
+-- * Utilities
+
+-- ** Simple index construction
+
+{-
+    Constructs an index from a list of keys. The keys have the same meaning as
+    in the vector given to the 'IndexOrdinary' data constructor.
+-}
+indexFromLastKeyList :: [SerialisedKey] -> IndexOrdinary
+indexFromLastKeyList = IndexOrdinary . fromList
+
+-- ** Partitioning according to search
+
+{-
+    Invokes the 'search' function to obtain a page span and partitions the key
+    list of the index into a prefix, selection, and suffix part, which
+    correspond to the pages before, within, and after the page span,
+    respectively.
+-}
+searchPartitioning
+    :: SerialisedKey
+    -> IndexOrdinary
+    -> (Vector SerialisedKey, Vector SerialisedKey, Vector SerialisedKey)
+searchPartitioning key index@(IndexOrdinary lastKeys)
+    = (prefix, selection, suffix)
+    where
+
+    start, end :: Int
+    PageSpan (PageNo start) (PageNo end) = search key index
+
+    prefix, selection, suffix :: Vector SerialisedKey
+    ((prefix, selection), suffix) = first (splitAt start) $
+                                    splitAt (succ end)    $
+                                    lastKeys
+
+-- Adds a search partitioning to a counterexample.
+searchPartitioningCounterexample
+    :: Testable prop
+    => Vector SerialisedKey
+    -> Vector SerialisedKey
+    -> Vector SerialisedKey
+    -> prop
+    -> Property
+searchPartitioningCounterexample prefix selection suffix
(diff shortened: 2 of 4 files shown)


test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,525 @@
...
+    selectionHead :: SerialisedKey
+    selectionHead = head selection
+
+prop_searchForUnmentionedKeyBeyondRangeWorks :: SerialisedKey
+                                             -> SearchableIndex
+                                             -> Property
+prop_searchForUnmentionedKeyBeyondRangeWorks key (SearchableIndex index)
+    = not (null lesserLastKeys) ==> all (<  selectionHead) prefix           .&&.
+                                    all (== selectionHead) (tail selection) .&&.
+                                    all (>  selectionHead) suffix
```
mheinzel, line 324: Ah, at first I didn't see how that ensure that the key is unmentioned and beyond range, but we construct new index to have that. Makes sense.

But why do we still check the same properties? Couldn't we just say that `prefix === lesserLastKeys` and the rest is empty?

test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,525 @@
...
+      isLeft (fromSBS input)
+    where
+
+    statedSizeBlock :: Primitive.Vector Word8
+    statedSizeBlock = byteVectorFromPrim statedSize
+
+    input :: ShortByteString
+    input = potentialSerialisedIndex
+                testedTypeAndVersionBlock
+                (lastKeysBlocks lastKeys ++ [statedSizeBlock, partialKeyBlock])
```
mheinzel, line 420: Yes, nice. Maybe one more test with a completely random key block would make sense, where we don't care about Left/Right, just that the result is `QC.total`. But this here should already cover the case of a key length going out of range. :+1:

test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,525 @@
...
+    shrink = shrinkMap (NumEntries . getNonNegative)
+                       (\ (NumEntries entryCount) -> NonNegative entryCount)
+
+instance Arbitrary SerialisedKey where
+
+    arbitrary = SerialisedKey' <$>
+                arbitrary `suchThat` (Primitive.length >>> (< 0x10000))
+
+    -- This shrinker relies on lists not becoming longer by shrinking.
+    shrink = shrinkMap SerialisedKey' (\ (SerialisedKey' bytes) -> bytes)
```
mheinzel, line 451: We already have these instances in `src-extras/../Generators.hs`. The distribution for the second one might be a bit different, but I don't think it's a problem.

test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,525 @@
...
+        genKeyCount :: Gen Int
+        genKeyCount = frequency $
+                      [(4, pure 1), (1, getSmall <$> getPositive <$> arbitrary)]
+
+    shrink (SearchableIndex (IndexOrdinary lastKeysVec))
+        = [
+              SearchableIndex (indexFromLastKeys lastKeys') |
+                  lastKeys' <- shrink (toList lastKeysVec),
+                  not (null lastKeys'),
+                  and (zipWith (<=) lastKeys' (List.tail lastKeys'))
```
mheinzel, line 479: neat

test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,525 @@
...
+          [
+              AppendMultiPage key' overflowPageCount |
+                  key' <- shrink key
+          ]
+          ++
+          [
+              AppendMultiPage key overflowPageCount' |
+                  overflowPageCount' <- shrinkMap getSmall Small $
+                                        overflowPageCount
+          ]
```
mheinzel, line 525: Another way of doing this (which also the compact index tests use) would be to use the existing generators for `LogicalPageSummaries`, which has a function `toAppends :: _ -> [Append]`.

mheinzel (commented):
Looks good, a few suggestions.

mheinzel:
Also, would be nice to see a rebase making the tests pass in CI. They pass for me locally when I rebase, but would be nice to see it here as well.

jeltsch:
> Also, would be nice to see a rebase making the tests pass in CI. They pass for me locally when I rebase, but would be nice to see it here as well.

Sure thing, and, since #363 is finally merged, I performed the rebasing. Let’s see if the builds succeed now.

jeltsch:
> Let’s see if the builds succeed now.

They do. 🙂

src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -57,6 +58,10 @@ supportedVersion = 1
     to return a valid page span.
 -}
 newtype IndexOrdinary = IndexOrdinary (Vector SerialisedKey)
+    deriving stock (Eq, Show)
+
+fromIndexOrdinary :: IndexOrdinary -> Vector SerialisedKey
+fromIndexOrdinary (IndexOrdinary lastKeys) = lastKeys
```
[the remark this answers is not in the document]
jeltsch: Indeed. I wasn’t aware of the intention of importing this module with qualification only, but it makes perfect sense. Fixed in f0f096297545a200d7b23680715994a5d511b1c8.

test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,525 @@
+{-# LANGUAGE MagicHash #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+{- HLINT ignore "Avoid restricted alias" -}
```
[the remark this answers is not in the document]
jeltsch: My problem with these particular HLint rules is that using local module names like `PV` has the following issues:

* Since these names use single-letter abbreviations for words like “primitive” or “mutable”, they can be hard to understand for the uninitiated.
* These names don’t follow the standard Haskell approach of using dots to delimit components of a module name but rather tie all components (single letters in this case) directly together. From what I know, it is indeed possible to use “hierarchical” module names also as aliases introduced via `as`.

Throughout the modules I’ve been contributing, I’ve been consistently following a style that avoids these problems:

* Since virtually all list-like operations are performed with vectors in these modules, the fact that vectors are used is not mentioned in names, resulting in particular in the absence of qualification for operations using standard vectors.
* The use of special kinds of vectors is signaled by meaningful name components, like `Primitive`, resulting in readable names like `Primitive.length` or, even better, `Primitive.Vector`.

To be honest, I would have nothing against dropping the above-mentioned HLint rules.

This issue here is an instance of a more general problem I see in lots of Haskell code: People are keen to use a consistent style, which is good, but this often results in them using a suboptimal style, because this style happens to be customary and possibly encouraged or even enforced by certain tools.

test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -0,0 +1,525 @@
...
+
+prop_searchForUnmentionedKeyInRangeWorks :: SerialisedKey
+                                         -> SearchableIndex
+                                         -> Property
+prop_searchForUnmentionedKeyInRangeWorks key (SearchableIndex index)
+    = key `notElem` lastKeys && key <= last lastKeys ==>
+      key < selectionHead                     .&&.
+      all (<  key)           prefix           .&&.
+      all (== selectionHead) (tail selection) .&&.
+      all (>  selectionHead) suffix
```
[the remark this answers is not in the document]
jeltsch: I like the solution with adding the search partitioning to what is shown as the counterexample: simple, yet effective. It would indeed have saved me some work when investigating the bug fixed in #393. 😉 I implemented this solution in 7141645ae03955e250b0214949a0fa1177637a72.

I looked into using `conjoin` instead of `(.&&.)` but found the results not really convincing and therefore left things as they were in this regard.

test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: No, because what should be returned is some arbitrary but valid page span, and so we have to check that a valid page span is indeed returned. Actually, the selection being empty would mean that the page span is not valid. By “valid” I mean here that the page span comprises a non-overflow page plus any overflow pages that belong to it.

jeltsch: > Ah, at first I didn't see how that ensure that the key is unmentioned and beyond range, but we construct new index to have that. Makes sense.

This is my little trick to avoid the problem that the initial solution had: it used a conditional property to only consider those cases where the key was large enough by accident, and this resulted in too many test cases being discarded.

test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: I think that a key length being too large is really the only error that can occur in a serialized key list beside the error where the serialized key list ends in a partial key size specification. The keys themselves can be arbitrary sequences of bytes; so, as long as the key length specifications are correct, any sequence of bytes is a valid serialized key list.

Just imagine how you can get the keys out of some given sequence of bytes: you take the size information from the beginning, chop off correspondingly many bytes to get the first key, then take the next size information, chop off correspondingly many bytes to get the second key, and so on. If you have an empty sequence of bytes at the end, everything is fine.

So the bottom line is that *any* random sequence of bytes is already covered by the tests, as it falls into one of the following three cases:

* Valid serialized key list
* Key list with an incomplete key size specification at the end
* Key list with a final key size specification that is too large

test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: Checking the corresponding code, I see that the `Arbitrary` instantiations of `NumEntries` and `SerialisedKey` reuse the instantiations of `Int` and `RawBytes`, respectively. However, what does the `Arbitrary` instantiation of `RawBytes` do? I didn’t find any documentation on this, and I must admit that the code looked complicated enough to discourage me from trying to understand it. Could you perhaps provide me with some sort of informal specification of the generator and the shrinker for `RawBytes`?

test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: Hmm, not sure about this. It seems to me that using the `LogicalPageSummaries` infrastructure for my tests would be overkill, in particular as that infrastructure deals with first and last keys of pages, while ordinary indexes only maintain last keys. Leaving the above code as it is seems the better alternative to me, at least for the time being.

Why does the compact index actually store both first and last keys? To me, this rather seems like maintaining redundant information, with the potential of making various things, tests for example, more complicated than needed.

test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: Currently, the builds fail, because I rebased this pull request onto #406. The problem that makes them fail is that the `Arbitrary` instantiations of `NumEntries` and `SerialisedKey` from `Database.LSMTree.Extras.Generators` are now imported and clash with my own instantiations. I understand that you would like me to use the instantiations from `Database.LSMTree.Extras.Generators` anyhow, but I currently don’t want to remove my own instantiations, as there are still questions that should be answered before.

Beside the above question regarding `RawBytes`, there is also the question how one should ensure that arbitrary values of `NumPages` fit into 64 bits and arbitary serialized keys have lengths that fit into 16 bits. How do the tests for the compact index address this? Are they just based on the assumption that, with the Quick Check versions and processor architectures used, nothing will go wrong?

test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
mheinzel: > Actually, the selection being empty would mean that the page span is not valid

Ah, you're right. There is of course still a selection. We could still have more specific (and tighter?) conditions we check here, but there is an argument to be made for using the same as in the other cases.

test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
mheinzel: That is true, as long as the generators for each of these cases cover the full space of inputs for that case. Looks good.

test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: I don’t think that we can make the condition tighter. Currently, it says that the page span must be valid, and this is really all we can require.

test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
mheinzel: > Why does the compact index actually store both first and last keys?

It doesn't.

`LogicalPageSummaries` does, because it is a representation of all information available from adding pages to an index. It's isomorphic to `[Append]`, so it seemed redundant to have separate `Arbitrary` instances. The compact index tests also use `LogicalPageSummaries`.

test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
mheinzel: > Could you perhaps provide me with some sort of informal specification of the generator and the shrinker for `RawBytes`?

At it's core it's a sequence of bytes, but the type has some extra cases that could be handled incorrectly in the code, so the tests make sure to cover them:

1. the underlying byte array being pinned or unpinned
2. only using a slice of out of a byte array

So the tests generate all these cases and also have an additional way of shrinking by just taking a smaller slice of the same byte array.

> how one should ensure that arbitrary values of NumPages fit into 64 bits and arbitary serialized keys have lengths that fit into 16 bits

Ints fit into 64 bits. For the length of serialised keys it's not a straightforward, but it depends on `QC.listof`, where the length is bounded by the size parameter (I think up to 100 by default?). So there shouldn't be issues with `SerialisedKey` being too long, but maybe one even wants to allow longer ones?

Either way, if you want to have local Arbitrary instances, it's probably a good idea to define a newtype and provide the instance for that, so you don't need orphan instances and don't have the risk of conflicting instances.

test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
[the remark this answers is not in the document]
mheinzel: I understand your points. When working in a team, I personally prefer consistent style over optimal style (whatever that means), but in this project we are generally not very strict on that.

I'm personally in favour of adhering to the hlint rules, but I can't speak for the others.

test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: Thanks for the explanation of the `Arbitrary RawBytes` instance. Makes sense to me. 🙂

jeltsch: I think that there’s no real guarantee that `Int` values fit into 64 bits. From what I remember, the Haskell Report only specifies that `Int` covers at least the interval [−2²⁷ + 1, +2²⁷ − 1]. I can imagine that there’s no architecture currently on which GHC would actually use more than the 64-bit range for `Int`, but what about all those future 128-bit architectures? 😉

jeltsch: I think I’ll use the instantiations from `Database.LSMTree.Extras.Generators` for both `NumPages` and `SerialisedKey` and add comments that specify the assumptions I’m making for the tests being correct: non-negative `Int` values being representable as 64-bit words and key lengths being representable as 16-bit words.

jeltsch: I’ve made the corresponding changes in f9fb2816cc8b88aa4dca701885857fed58766b56.

test/Test/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: I’ve made corresponding changes in 7619f450a8f541fa176edf83974e2259e3c81b1e.

mheinzel (approved):
Looks good!

Not sure why one CI run failed on the docspec stuff, seems like it's flaky. If you squash your commits together a bit and push again, that might get it to pass.

PR #363, opened 2024-08-31
Include the index type in the index versioning information

### LINKED ISSUE #263: Include the index type in the index versioning information
The serialization formats of both the compact and the ordinary index reserve the first 32 bits for storing the version of the index. Currently, we number versions of the compact and the ordinary index independently, which means that the type of index, compact or ordinary, cannot be detected from those first 32 bits. However, there is also no other part of a serialized index that would reliably tell its type. To get out of this unfortunate situation, @joris and I [propose](https://github.com/IntersectMBO/lsm-tree/pull/244#discussion_r1639687153) to use the first 32 bits of a serialized index also to indicate its type.

So far, we’ve been using the serialized version number also to capture the endianness used by a serialized index, which has been made possible by picking version numbers appropriately. It is important to preserve the ability to detect the endianness when including information about the index type. To this end, I propose the following structure for the 32-bit number to be stored at the beginning of an index:

* Bits 0–7 contain the actual version number.
* Bits 8–15 contain a number denoting the type of the index.
* Bits 16–31 are zero.

Here, bit numbers refer to significance. To put things differently, the initial 32-bit number can be written in hexadecimal with 4-digits of which the first two denote the type of the index and the last two denote the actual version number. If version numbers are never zero, this structure guarantees that little-endian and big-endian can always be told apart. [Mixed-endian](https://en.wikipedia.org/wiki/Endianness#Middle-endian) could still cause problems, but my understanding is that virtually no architecture in use today supports this kind of endianness.

I furthermore propose the following encodings of index types:

* 0 for the compact index
* 1 for the ordinary index

With this scheme, nothing has to be changed for the compact index, while the ordinary index must have its version number 1 changed to the type–version indicator 0101h.

If this approach of storing the index type and the index version is adopted, we should perform the following tasks:

* [x] Adapt the code of the ordinary index as described above
* [x] Rename `indexVersion` in the code of the compact index such that its new, expanded meaning is expressed
* [x]  Document this approach

jeltsch:
This resolves #263.

### DIFF
--- a/doc/format-run.md
+++ b/doc/format-run.md
@@ -204,7 +204,7 @@ whole key per page: we can typically store just 8 bytes (64bits) per page. This
 is a factor of 4 saving for 32 byte keys.
 
 The type of compact indexes is denoted by the number 0. The type–version
-indicater in a serialised compact index is padded to 64 bits.
+indicator in a serialised compact index is padded to 64 bits.
 
 For version 1, the representation after the type–version indicator consists of
 1. a primary array of 64bit words, one entry per page in the index
@@ -269,7 +269,7 @@ this, the following constraints on the sequence of stored keys follow:
 * The elements of the sequence must be non-decreasing.
 
 The type of compact indexes is denoted by the number 1. The type–version
-indicater in a serialised ordinary index is not padded.
+indicator in a serialised ordinary index is not padded.
 
 For version 1, the file contents after the type–version indicator consist of the
 following:
--- a/doc/format-run.md
+++ b/doc/format-run.md
@@ -185,11 +185,11 @@ bit numbers refer to significance:
 * Bits 16–31 are zero.
 
 As a result, a type–version indicator can be written in hexadecimal with only
-4-digits of which the first two denote the type and the last two constitute the
-version number. Version numbers are assigned independently for each particular
-type. They must not be zero, which ensures that from the way the type–version
-indicator is stored one can determine whether the concrete serialised index
-stores numbers in little-endian or big-endian.
+4-digits of which the two most significant denote the type and the two least
+significant constitute the version number. Version numbers are assigned
+independently for each particular type. They must not be zero, which ensures
+that from the way the type–version indicator is stored one can determine whether
+the concrete serialised index stores numbers in little-endian or big-endian.
 
 ### Compact index
 
--- a/doc/format-run.md
+++ b/doc/format-run.md
@@ -187,9 +187,10 @@ bit numbers refer to significance:
 As a result, a type–version indicator can be written in hexadecimal with only
 4-digits of which the two most significant denote the type and the two least
 significant constitute the version number. Version numbers are assigned
-independently for each particular type. They must not be zero, which ensures
-that from the way the type–version indicator is stored one can determine whether
-the concrete serialised index stores numbers in little-endian or big-endian.
+independently for each particular type. A version number must not be zero, which
+ensures that from the way the type–version indicator is stored one can determine
+whether the concrete serialised index stores numbers in little-endian or
+big-endian.
 
 ### Compact index
 
--- a/doc/format-run.md
+++ b/doc/format-run.md
@@ -175,6 +175,22 @@ ordinary index type.
 The type of index in use is metadata that must be known for the whole LSM table
 handle. It does not vary per run.
 
+Independently of the concrete index serialization format, the type and the
+version of the format are stored at the beginning as a 32-bit number called the
+type–version indicator. A type–version indicator has the following shape, where
+bit numbers refer to significance:
+
+* Bits 0–7 contain the version number.
+* Bits 8–15 contain a number denoting the type of the index.
+* Bits 16–31 are zero.
+
+As a result, a type–version indicator can be written in hexadecimal with only
+4-digits of which the first two denote the type and the last two constitute the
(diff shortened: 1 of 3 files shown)


doc/format-run.md
```diff
@@ -253,11 +268,10 @@ this, the following constraints on the sequence of stored keys follow:
 * The sequence must be non-empty.
 * The elements of the sequence must be non-decreasing.
 
-Like a compact-index file, an ordinary-index file starts with a 32 bit number
-that states the format version and, as a side effect, also determines the
-endianness used in the file.
+The type of compact indexes is denoted by the number 1. The type–version
+indicater in a serialised ordinary index is not padded.
```
jorisdral, line 272:
```suggestion
indicator in a serialised ordinary index is not padded.
```

doc/format-run.md
```diff
@@ -175,6 +175,22 @@ ordinary index type.
 The type of index in use is metadata that must be known for the whole LSM table
 handle. It does not vary per run.
 
+Independently of the concrete index serialization format, the type and the
+version of the format are stored at the beginning as a 32-bit number called the
+type–version indicator. A type–version indicator has the following shape, where
+bit numbers refer to significance:
```
jorisdral, line 181: This does not tell the reader which number indicates highest significance

doc/format-run.md
```diff
@@ -175,6 +175,22 @@ ordinary index type.
 The type of index in use is metadata that must be known for the whole LSM table
 handle. It does not vary per run.
 
+Independently of the concrete index serialization format, the type and the
+version of the format are stored at the beginning as a 32-bit number called the
```
jorisdral, line 179: What do you mean by the first part of the sentence before the comma? The index serialization format is not stored anywhere

doc/format-run.md
```diff
@@ -175,6 +175,22 @@ ordinary index type.
...
+type–version indicator. A type–version indicator has the following shape, where
+bit numbers refer to significance:
+
+* Bits 0–7 contain the version number.
+* Bits 8–15 contain a number denoting the type of the index.
+* Bits 16–31 are zero.
+
+As a result, a type–version indicator can be written in hexadecimal with only
+4-digits of which the first two denote the type and the last two constitute the
```
jorisdral, line 188: Replace first and last with most and least significant

doc/format-run.md
```diff
@@ -175,6 +175,22 @@ ordinary index type.
...
+version number. Version numbers are assigned independently for each particular
+type. They must not be zero, which ensures that from the way the type–version
```
jorisdral, line 190: Does "they" refer to both the version number and type, or one of them?

doc/format-run.md
```diff
@@ -187,11 +203,10 @@ For this important special case, we can do significantly better than storing a
 whole key per page: we can typically store just 8 bytes (64bits) per page. This
 is a factor of 4 saving for 32 byte keys.
 
-Just as the bloom filter, the compact index file format starts with a 32 bit
-format identifier / format version, which determines the format and endianness
-of the rest of the file. This field is padded to 64 bits.
+The type of compact indexes is denoted by the number 0. The type–version
+indicater in a serialised compact index is padded to 64 bits.
```
jorisdral, line 207: indicator

jorisdral (changes requested):
LGTM - I have some suggested changes for the documentation, but the code changes look great

doc/format-run.md
jeltsch: Fixed in 7374994e5e945729c5dfb45d79709e9f51674bcb.

doc/format-run.md
jeltsch: Oh, no! This mistake again. Fixed 7374994e5e945729c5dfb45d79709e9f51674bcb.

doc/format-run.md
jeltsch: Well, bits with higher numbers have higher significance. This is the established convention. 🙂 I don’t think that it is necessary to explicitly mention this here.

doc/format-run.md
jeltsch: I mean that, no matter what particular index serialization format is used, there is this 32-bit number at the beginning or, to put it differently, that all index serialization formats agree in having serialized indexes start with this 32-bit number. Does this need clarification?

doc/format-run.md
jeltsch: It is about how this number is written, and in writing there are first and last digits. Well, if you place the same written representation of a number into some Arabic text, for example, first becomes last and *vice versa*. Maybe I should indeed change this.

jeltsch: Done in 17a3c914063c83f0680ebef15acce01e510f7470.

doc/format-run.md
jeltsch: “They” refers to “version numbers”. I think that the natural interpretation is that this occurrence of “they” refers to only one nominal phrase of the previous sentence, and “version numbers” is then the phrase to pick, because it is the only plural phrase in this sentence and is sort of the default because of being the subject phrase. In my opinion, it should be intuitively rather clear that this is meant, but maybe I can clarify this.

jeltsch: An improved version is in ee678027a27d290045204b7c75b20cda180865ee.

jorisdral (approved):

PR #377, opened 2024-09-04
Comparative benchmark for table scans

jorisdral:
This benchmark can be used to compare table scans performed with cursor reads against table scans performed with range lookups.

WIP: still have to split out the changes a little bit


mheinzel (approved):
LGTM. I'll rebase and merge.

PR #381, opened 2024-09-05
Generalise RunReaders to support write buffers

mheinzel:
# Description

Allows to pass a write buffer into `RunReaders`. This is needed for cursors, which read from all runs and the write buffer.

This requires an additional branch for each entry being read. In the run merge micro benchmark, allocations and run time are unaffected.

# Checklist

### DIFF
--- a/src/Database/LSMTree/Internal/RunReaders.hs
+++ b/src/Database/LSMTree/Internal/RunReaders.hs
@@ -1,35 +1,41 @@
 module Database.LSMTree.Internal.RunReaders (
     Readers (..)
-  , ReaderNumber (..)
-  , ReadCtx (..)
   , new
   , close
   , peekKey
   , HasMore (..)
   , pop
   , dropWhileKey
+    -- * Internals
+  , Reader (..)
+  , ReaderNumber (..)
+  , ReadCtx (..)
   ) where
 
 import           Control.Monad (zipWithM)
 import           Control.Monad.Primitive
 import           Data.Function (on)
+import           Data.Functor ((<&>))
 import           Data.List.NonEmpty (nonEmpty)
 import           Data.Maybe (catMaybes)
 import           Data.Primitive.MutVar
 import           Data.Traversable (for)
+import           Database.LSMTree.Internal.BlobRef (BlobRef)
+import qualified Database.LSMTree.Internal.Entry as Entry
 import           Database.LSMTree.Internal.Run (Run)
 import           Database.LSMTree.Internal.RunReader (RunReader)
 import qualified Database.LSMTree.Internal.RunReader as Reader
 import           Database.LSMTree.Internal.Serialise
+import qualified Database.LSMTree.Internal.WriteBuffer as WB
 import qualified KMerge.Heap as Heap
 import qualified System.FS.API as FS
 import           System.FS.API (HasFS)
 import           System.FS.BlockIO.API (HasBlockIO)
 
--- | Abstraction for the collection of 'RunReader', yielding elements in order.
--- More precisely, that means first ordered by their key, then by the input
--- run they came from. This is important for resolving multiple entries with the
--- same key into one.
+-- | A collection of runs and write buffers being read from, yielding elements
+-- in order. More precisely, that means first ordered by their key, then by the
+-- input run they came from. This is important for resolving multiple entries
+-- with the same key into one.
 --
 -- Construct with 'new', then keep calling 'pop'.
 -- If aborting early, remember to call 'close'!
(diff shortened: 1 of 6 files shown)


jorisdral (approved):
LGTM -- some small suggestions, but then we can merge

src/Database/LSMTree/Internal/RunReaders.hs
```diff
@@ -186,12 +216,18 @@ nextReadCtx ::
...
-nextReadCtx fs hbio readCtxNumber readCtxReader = do
-    res <- Reader.next fs hbio readCtxReader
-    case res of
-      Reader.Empty -> do
-        return Nothing
-      Reader.ReadEntry readCtxHeadKey readCtxHeadEntry ->
-        return (Just ReadCtx {..})
+nextReadCtx fs hbio readCtxNumber readCtxReader =
+    case readCtxReader of
+      ReadRun r -> flip fmap (Reader.next fs hbio r) $ \case
```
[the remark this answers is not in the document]
mheinzel: But then I need to `pure/return`. :shrug: 
Some people use `<&>`, but I don't think we do anywhere. So this seemed like a reasonable solution, especially since the `ReadBuffer` version below looks pretty similar, applying a pure function as well.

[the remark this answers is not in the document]
jorisdral: I overlooked the need for `pure`/`return`. Arguably, `<&>` would be slightly better

src/Database/LSMTree/Internal/RunReaders.hs
```diff
@@ -78,22 +86,51 @@ instance Eq (ReadCtx fhandle) where
...
+-- A more general version is possible, but despite SPECIALISE-ing everything
+-- showed ~100 bytes of extra allocations per entry that is read (which might be
+-- avoidable with some tinkering).
+data Reader m fhandle =
+    ReadRun    !(RunReader m fhandle)
+    -- The list allows to incrementally read from the write buffer without
+    -- having to find the next entry in the Map again (requiring key
+    -- comparisons) or having to copy out all entries.
+    -- TODO: more efficient representation? benchmark!
+  | ReadBuffer !(MutVar (PrimState m) [KOp m fhandle])
```
mheinzel, line 102: This causes [NoThunks test failures], but a thunk probably makes sense here. Should I somehow change the `NoThunks` instance to allow a thunk in this particular place? @jorisdral

jorisdral: I'll have a look

jorisdral: For now, let's just do this:

```diff
diff --git a/src-extras/Database/LSMTree/Extras/NoThunks.hs b/src-extras/Database/LSMTree/Extras/NoThunks.hs
index cee8a61c..597071c7 100644
--- a/src-extras/Database/LSMTree/Extras/NoThunks.hs
+++ b/src-extras/Database/LSMTree/Extras/NoThunks.hs
@@ -525,7 +525,9 @@ instance NoThunks a => NoThunks (MutVar s a) where
   showTypeOf (_ :: Proxy (MutVar s a)) = "MutVar"
   wNoThunks ctx var = do
       x <- unsafeSTToIO $ readMutVar var
-      noThunks ctx x
+      -- TODO: disallow thunk once we find out why atomicModifyMutVar' is
+      -- creating thunks
+      noThunks ctx (AllowThunk x)

 -- TODO: https://github.com/input-output-hk/nothunks/issues/56
 deriving via OnlyCheckWhnf (PrimVar s a)
```

I might have to play around with `MutVar`s to find out why `atomicModifyMutVar'` is creating thunks...

mheinzel: I don't think it is about the `MutVar` here, but the list inside of it. It is intentionally a thunk. We don't want to copy the whole write buffer into a list, but incrementally evaluate the list as we read entries.

I will try tweaking the instance `NoThunks (Reader m (Handle h))`.

mheinzel: Like this?


```haskell
instance (Typeable (PrimState m), Typeable h)
      => NoThunks (Reader m (Handle h)) where
  showTypeOf (_ :: Proxy (Reader m (Handle h))) = "Reader"
  wNoThunks ctx = \case
    ReadRun r      -> noThunks ctx r
    ReadBuffer var -> return Nothing  -- intentionally lazy
```

mheinzel: It doesn't check that the MutVar itself is not a thunk, as that seemed slightly trickier to do without also checking the content of the MutVar. But maybe good enough?

jorisdral:
```diff
- ReadBuffer var -> return Nothing  -- intentionally lazy
+ ReadBuffer var -> noThunks ctx (OnlyCheckWhnf var)

PR #384, opened 2024-09-06
Revive the `extras` sublibrary

jorisdral:
# Description

We've had some discussion in the past about whether (i) we want this as a sublibrary, or whether (ii) we want to include the source files directly in the components that need them. I'd like to revive option (ii), since option (i) degrades my HLS experience due to ambiguity -- I suspect HLS gets confused because a file in the `Extras` hierarchy might be loaded as part of any component that includes the file. It basically means I get no useful HLS feedback as soon as I open an `Extras` file, and it occasionally prevents HLS from giving useful feedback on modules that depend on such an `Extras` module

# Checklist

### DIFF
--- a/lsm-tree.cabal
+++ b/lsm-tree.cabal
@@ -309,18 +309,11 @@ library extras
 test-suite lsm-tree-test
   import:         language, warnings, wno-x-partial
   type:           exitcode-stdio-1.0
-  hs-source-dirs: test src-extras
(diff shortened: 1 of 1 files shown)


mheinzel (approved):
I don't have a strong opinion on this, but the HLS argument makes sense to me.

dcoutts (approved):
I don't mind. I don't use HLS.

PR #393, opened 2024-09-16
Fix a bug in ordinary index search

jeltsch:
This fixes a bug that violated `prop_searchForUnmentionedKeyInRangeWorks`. It also contains a small stylistic improvement in nearby code, which is in line with the style of the fixed code.

### DIFF
--- a/src/Database/LSMTree/Internal/IndexOrdinary.hs
+++ b/src/Database/LSMTree/Internal/IndexOrdinary.hs
@@ -13,7 +13,7 @@ module Database.LSMTree.Internal.IndexOrdinary
 )
 where
 
-import           Prelude hiding (drop, length, takeWhile)
+import           Prelude hiding (drop, last, length)
 
 import           Control.Exception (assert)
 import           Control.Monad (when)
@@ -21,8 +21,8 @@ import           Data.ByteString.Short (ShortByteString (SBS))
 import qualified Data.ByteString.Short as ShortByteString (length)
 import           Data.Primitive.ByteArray (ByteArray (ByteArray),
                      indexByteArray)
-import           Data.Vector (Vector, drop, findIndexR, fromList, length,
-                     takeWhile, (!))
+import           Data.Vector (Vector, drop, findIndex, findIndexR, fromList,
+                     last, length, (!))
 import qualified Data.Vector.Primitive as Primitive (Vector (Vector), drop,
                      force, length, null, splitAt, take)
 import           Data.Word (Word16, Word32, Word64, Word8, byteSwap32)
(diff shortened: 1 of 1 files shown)


src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -72,15 +72,19 @@ search key (IndexOrdinary lastKeys)
...
-        | otherwise
-            = pred pageCount
+    end | protoStart < pageCount = start + overflowPageCount
+        | otherwise              = pred pageCount
+        where
+
+        overflowPageCount :: Int
+        overflowPageCount = length                          $
+                            takeWhile (== lastKeys ! start) $
+                            drop (succ start) lastKeys
```
mheinzel, line 87: If we use `findIndex`, we avoid relying on the optimiser to get rid of the intermediate vector. We also use `findIndexR` above to do something very similar, so it would be consistent.

src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -72,15 +72,19 @@ search key (IndexOrdinary lastKeys)
...
+          | otherwise              = maybe 0 succ                              $
                                      findIndexR (/= lastKeys ! pred pageCount) $
                                      lastKeys
 
     end :: Int
-    end | protoStart < pageCount
-            = start + length (takeWhile (== key) (drop (succ start) lastKeys))
-        | otherwise
-            = pred pageCount
+    end | protoStart < pageCount = start + overflowPageCount
```
mheinzel, line 80: There is the same guard in both of these definitions. Maybe we should put them under a single guard, returning `(Int, Int)` or even the `PageSpan`?

mheinzel (commented):
Good catch!

src/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: Yes, this makes a lot of sense. Done in 1c036b3c2b793fc0c535558077800f55f4e0fd24.

jeltsch: Absolutely. Done as part of 9486ca27fdb2f335729f5f6f1d50982022e306ec.

src/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: Done as part of 9486ca27fdb2f335729f5f6f1d50982022e306ec.

src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -66,27 +66,35 @@ newtype IndexOrdinary = IndexOrdinary (Vector SerialisedKey)
...
+    !result | protoStart < pageCount
+                = let
+
+                      end :: Int
+                      !end = maybe (pred pageCount) (+ protoStart) $
+                             findIndex (/= lastKeys ! protoStart)  $
+                             drop (succ protoStart) lastKeys
+
+                  in PageSpan (PageNo $ protoStart)
+                              (PageNo $ end)
```
mheinzel, line 87: The `$` are not needed here (and similarly for `start` below`), so I would leave them out.

src/Database/LSMTree/Internal/IndexOrdinary.hs
```diff
@@ -66,27 +66,35 @@ newtype IndexOrdinary = IndexOrdinary (Vector SerialisedKey)
...
 
     pageCount :: Int
-    pageCount = length lastKeys
+    !pageCount = length lastKeys
+
+    result :: PageSpan
+    !result | protoStart < pageCount
+                = let
+
+                      end :: Int
```
mheinzel, line 81: I know it's a matter of taste, but the code feels very "airy" and spaced out, taking up a lot of screen space. Here for example, I'd personally not make these two line breaks.

mheinzel (approved):
Looks good. I left some stylistic opinions, but feel free to disregard them.

src/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: Usually, I don’t write such code; in fact, it seems to be the first time that I did. That said, I did so for a reason: The definition of `result` is basically about computing the indexes of the start and the end page, and therefore I wanted these two pieces of data to be clearly stated. However, the occurrences of the data constructors `PageSpan` and `PageNo` are distracting. By separating the actual data from the additional data constructors using `$` and aligning the arguments of `PageSpan` vertically, I could make the actual data somewhat stick out.

src/Database/LSMTree/Internal/IndexOrdinary.hs
jeltsch: I know that I have the tendency of writing “airy” code. 🙂 Well, it’s also a matter of perspective: for example, I personally find some other code quite cramped.

As for the example you quoted: The definition of `end` is, well, a definition of its own, like the including definition of `result` and ultimately the definition of `search`. As such, it has some complexity, which might make it advisable to separate it from the surrounding code by one empty line on each side. For global definitions, this is what virtually everyone is doing, and doing it for local definitions as well has the further advantage of keeping the (implicit) style rules I’m following here simple and consistent.

As for saving screen space: This is the, quite usual, pragmatic argument. 🙂 However, I don’t think that these days there is a shortage of screen space. In fact, I believe that the typical person looking at Haskell code has much more screen space than I have, and I must say that I don’t suffer from looking at my code. 😉

jeltsch:
> Looks good. I left some stylistic opinions, but feel free to disregard them.

I reacted to your opinions above. In the end, I would like to get this branch merged into `main` as it is. However, I can’t add it to the merge queue, as I’m not a code owner. Could you please do this for me? Thanks a lot.

PR #398, opened 2024-09-25
Make the`Refcounter`'s `finaliser` field  less strict.

jorisdral:
In two places, we use `RecursiveDo` to attach a `RefCounter` with finaliser to a `Run`. Because we evaluate the finaliser as much as possible before storing it in the `RefCounter`, we are getting "cyclic evaluation" errors in some cases where we use recursive `do` (i.e., `fixIO`). To prevent these errors from occurring, we instead do not evaluate the finaliser at all before storing it in a `RefCounter`.


mheinzel (approved):
Makes sense.

PR #401, opened 2024-09-26
Make `IndexCompact.search` return valid page spans in some edge cases

jorisdral:
In case `search` was given a key that was smaller than all keys on all pages in the index, then `search` would return `PageSpan 0 0` by default. However, if page 0 contains a large value that overflows into subsequent pages, then `PageSpan 0 0` is not valid because we have to include all the overflow pages as well. Instead, `search` now consults the `LTP` array to ensure that we return a valid page span. In addition, the `search` code has been cleaned up a little, and it's documentation has been improved.

A `TODO` in `RunReader.new` has been addressed, which depended on `search` returning valid page spans.

### DIFF
--- a/src/Database/LSMTree/Internal/IndexCompact.hs
+++ b/src/Database/LSMTree/Internal/IndexCompact.hs
@@ -436,42 +436,63 @@ pageSpanSize :: PageSpan -> NumPages
 pageSpanSize pspan = NumPages . toEnum $
     unPageNo (pageSpanEnd pspan) - unPageNo (pageSpanStart pspan) + 1
 
--- | Given a search key, find the number of the disk page that /might/ contain
--- this key.
+-- |  Searches for a page span that contains a key–value pair with the given key.
 --
--- Note: there is no guarantee that the disk page contains the key. However, it
--- is guaranteed that /no other/ disk page contain the key.
+-- If there is indeed such a pair, the result is the corresponding page span; if
+-- there is no such pair, the result is an arbitrary but valid page span.
 --
 -- See [an informal description of the search algorithm](#search-descr) for more
 -- details about the search algorithm.
---
--- TODO: optimisation ideas: we should be able to get it all unboxed,
--- non-allocating and without unnecessary bounds checking.
---
+search :: SerialisedKey -> IndexCompact -> PageSpan
 -- One way to think of the search algorithm is that it starts with the full page
 -- number interval, and shrinks it to a minimal interval that contains the
--- search key. The code below is annotated with @Pre:@ and @Post:@ comments that
--- describe the interval at that point.
-search :: SerialisedKey -> IndexCompact -> PageSpan
+-- search key. The code below is annotated with [x,y] or [x, y) comments that
+-- describe the known page number interval at that point in the search
+-- algorithm.
 search k IndexCompact{..} =
-    let !primbits = keyTopBits64 k
-    in
-      case unsafeSearchLE primbits icPrimary of
-        Nothing -> singlePage (PageNo 0)  -- Post: @[lb, lb)@ (empty).
-        Just !i ->         -- Post: @[lb, i]@.
-          if unBit $ icClashes VU.! i then
-            -- Post: @[lb, i]@, now in clash recovery mode.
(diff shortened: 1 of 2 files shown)


src/Database/LSMTree/Internal/IndexCompact.hs
```diff
@@ -436,42 +436,67 @@ pageSpanSize :: PageSpan -> NumPages
...
     unPageNo (pageSpanEnd pspan) - unPageNo (pageSpanStart pspan) + 1
 
--- | Given a search key, find the number of the disk page that /might/ contain
--- this key.
+-- |  Searches for a page span that contains a key–value pair with the given key.
 --
--- Note: there is no guarantee that the disk page contains the key. However, it
--- is guaranteed that /no other/ disk page contain the key.
+-- If there is indeed such a pair, the result is the corresponding page span; if
+-- there is no such pair, the result is an arbitrary but valid page span.
```
jorisdral, line 442: INFO: this is now the same as `IndexOrdinary.search`

src/Database/LSMTree/Internal/IndexCompact.hs
```diff
@@ -436,42 +436,67 @@ pageSpanSize :: PageSpan -> NumPages
...
 --
 -- See [an informal description of the search algorithm](#search-descr) for more
 -- details about the search algorithm.
---
--- TODO: optimisation ideas: we should be able to get it all unboxed,
--- non-allocating and without unnecessary bounds checking.
```
jorisdral, line 449: INFO: It's actually already non-allocating! There are probably still other optimisation opportunities, but I'm removing the comment here

src/Database/LSMTree/Internal/IndexCompact.hs
```diff
@@ -436,42 +436,67 @@ pageSpanSize :: PageSpan -> NumPages
...
-        Just !i ->         -- Post: @[lb, i]@.
-          if unBit $ icClashes VU.! i then
-            -- Post: @[lb, i]@, now in clash recovery mode.
-            let !i1  = PageNo $ fromMaybe 0 $
-                  bitIndexFromToRev (BoundInclusive 0) (BoundInclusive i) (Bit False) icClashes
-                !i2  = maybe (PageNo 0) snd $ Map.lookupLE (unsafeNoAssertMakeUnslicedKey k) icTieBreaker
-                PageNo !i3 = max i1 i2 -- Post: the intersection of @[i1, i]@ and @[i2, i].
-                !i4 = bitLongestPrefixFromTo (BoundExclusive i3) (BoundInclusive i) (Bit True) icLargerThanPage
-                      -- Post: [i3, i4]
-            in  if i3 == i4 then singlePage (PageNo i3) else multiPage (PageNo i3) (PageNo i4)
```
jorisdral, line 470: INFO: I noticed that this is an unnecessary check. `multiPage (PageNo i3) (PageNo i4)` provides the same result

src/Database/LSMTree/Internal/IndexCompact.hs
```diff
@@ -436,42 +436,67 @@ pageSpanSize :: PageSpan -> NumPages
...
+        -- [0, n), our page span definitely starts at 0, but we still have to
+        -- consult the LTP array to check whether the value on page 0 overflows
+        -- into subsequent pages.
+        let !i = bitLongestPrefixFromTo (BoundExclusive 0) NoBound (Bit True) icLargerThanPage
+        -- [0, i]
+        in  if i == 0
+            then singlePage (PageNo 0)
+                 -- [0, 0]
+            else multiPage (PageNo 0) (PageNo i)
+                 -- [0, i]
```
mheinzel, line 470: Isn't this the same as https://github.com/IntersectMBO/lsm-tree/pull/401#discussion_r1776696453, so we could remove the check?

mheinzel (approved):
Nice, maybe one small tweak.

src/Database/LSMTree/Internal/IndexCompact.hs
jorisdral: Definitely!

PR #406, opened 2024-09-28
Add `Arbitrary` instantiation for primitive vectors

jeltsch:
This makes the type of primitive vectors an instance of `Arbitrary`. The corresponding instance declaration is added to `Database.LSMTree.Extras.Generators`, which previously had only been used for LSM-tree-specific things. If it should be put elsewhere, the reviewer should say so. 🙂


mheinzel (approved):

recursion-ninja (approved):
Looks good to me!

PR #407, opened 2024-09-30
Replace `TypedWriteBuffer` by the new `RunData`

jorisdral:


jorisdral:
FYI, it's probably useful to hide whitespace changes in the diff view, which should make the changes more readable

dcoutts (approved):
LGTM! And better name now.

PR #410, opened 2024-10-01
Give errors in `addReference` and `removeReference` a callstack

jorisdral:
This makes it slightly easier to see where assertions related to reference count invariants fail.


dcoutts (approved):
LGTM. The redundant constraint warning for the no-debug builds makes some sense, but I'd argue it's a GHC bug. So perhaps we can just suppress that warning for that buggy GHC version, since it doesn't seem to happen in the latest version.

PR #414, opened 2024-10-03
Use cabal haddock-project instead of our custom script

jorisdral:
This generates better documentation than our own script does. `cabal haddock-project` is experimental, but we can always revive our script in the future when needed since it's immortalised in our git history


recursion-ninja (approved):
📤 LGTM. 

Out of curiosity, was there any additionally functionality our custom script provided which `cabal haddock-project` does not?

jorisdral:
> 📤 LGTM.
> 
> Out of curiosity, was there any additionally functionality our custom script provided which `cabal haddock-project` does not?

Not that I know of. The main feature of the script over just using `cabal haddock`, is that the script would provide a top-level `index.html` that would link to all the haddocks for the sub-components. `cabal haddock-project` does the same thing, but nicer

PR #419, opened 2024-10-04
Enable ordinary-index building without a page count bound

jeltsch:
So far, incremental construction of an ordinary index required users to provide an upper bound of the number of pages or equivalently the number of keys stored in the index. However, such an upper bound cannot be reliably computed.

This pull request introduces an implementation of “growing vectors” and adapts ordinary-index construction to make use of it. As a result, it is not necessary anymore to provide an upper bound of the number of keys but only an initial size of the key buffer. For performance reasons, it is good for this initial size to be an upper bound of the number of keys, but index construction will also work if it is not.

This pull request does not contain tests for growing vectors. These will be added later by resolving #420.

### DIFF
--- a/src/Database/LSMTree/Internal/Vector/Growing.hs
+++ b/src/Database/LSMTree/Internal/Vector/Growing.hs
@@ -1,0 +1,120 @@
+{-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
+
+{- HLINT ignore "Avoid restricted alias" -}
+
+-- | Vectors with support for appending elements.
+module Database.LSMTree.Internal.Vector.Growing
+(
+    GrowingVector,
+    new,
+    append,
+    freeze
+)
+where
+
+import           Prelude hiding (init, last, length)
+
+import           Control.Monad (when)
+import           Control.Monad.ST.Strict (ST)
+import           Data.Primitive.PrimVar (PrimVar, newPrimVar, readPrimVar,
+                     writePrimVar)
+import           Data.STRef.Strict (STRef, newSTRef, readSTRef, writeSTRef)
+import           Data.Vector (Vector)
+import qualified Data.Vector as Mutable (freeze)
+import           Data.Vector.Mutable (MVector)
+import qualified Data.Vector.Mutable as Mutable (grow, length, new, set, slice,
+                     take)
+
+{-|
+    A vector with support for appending elements.
+
+    Internally, the elements of a growing vector are stored in a buffer. This
+    buffer is automatically enlarged whenever this is needed for storing
+    additional elements. On each such enlargement, the size of the buffer is
+    multiplied by a power of 2, whose exponent is chosen just big enough to make
+    the final buffer size at least as high as the new vector length.
+
+    Note that, while buffer sizes and vector lengths are represented as 'Int'
+    values internally, the above-described buffer enlargement scheme has the
+    consequence that the largest possible buffer size and thus the largest
+    possible vector length may not be the maximum 'Int' value. That said, they
+    are always greater than half the maximum 'Int' value, which should be enough
+    for all practical purposes.
+-}
+data GrowingVector s a = GrowingVector
+                             !(STRef s (MVector s a)) -- Reference to the buffer
+                             !(PrimVar s Int)         -- Reference to the length
+
+-- | Creates a new, initially empty, vector.
+new :: Int                      -- ^ Initial buffer size
+    -> ST s (GrowingVector s a) -- ^ Construction of the vector
+new illegalInitialBufferSize | illegalInitialBufferSize <= 0
+    = error "Initial buffer size not positive"
+new initialBufferSize
+    = do
+        buffer <- Mutable.new initialBufferSize
+        bufferRef <- newSTRef $! buffer
+        lengthRef <- newPrimVar 0
+        return (GrowingVector bufferRef lengthRef)
+
+{-|
+    Appends a value a certain number of times to a vector. If a negative number
+    is provided as the count, the vector is not changed.
+-}
+append :: forall s a . GrowingVector s a -> Int -> a -> ST s ()
+append _ pseudoCount _ | pseudoCount <= 0
+    = return ()
+append (GrowingVector bufferRef lengthRef) count val
+    = do
+          length <- readPrimVar lengthRef
+          makeRoom
+          buffer' <- readSTRef bufferRef
+          Mutable.set (Mutable.slice length count buffer') val
+    where
+
+    makeRoom :: ST s ()
+    makeRoom = do
+        length <- readPrimVar lengthRef
+        when (count > maxBound - length) (error "New length too large")
+        buffer <- readSTRef bufferRef
+        let
+
+            bufferSize :: Int
+            !bufferSize = Mutable.length buffer
+
+            length' :: Int
+            !length' = length + count
+
+        when (bufferSize < length') $ do
+            let
+
+                higherBufferSizes :: [Int]
+                higherBufferSizes = tail (init ++ [last]) where
+
+                    init :: [Int]
+                    last :: Int
+                    (init, last : _) = span (<= maxBound `div` 2) $
+                                       iterate (* 2) bufferSize
+                    {-NOTE:
+                        In order to prevent overflow, we have to start with the
+                        current buffer size here, although we know that it is
+                        not sufficient.
+                    -}
+
+                sufficientBufferSizes :: [Int]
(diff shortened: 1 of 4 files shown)


src/Database/LSMTree/Internal/Vector/Growing.hs
```diff
@@ -0,0 +1,117 @@
...
+{-|
+    Turns a growing vector into an ordinary vector, thereby invalidating the
+    growing vector. Executing @unsafeFreeze vec@ is only safe when @vec@ is not
+    used afterwards.
+-}
+unsafeFreeze :: GrowingVector s a -> ST s (Vector a)
+unsafeFreeze (GrowingVector bufferRef lengthRef) = do
+    buffer <- readSTRef bufferRef
+    length <- readPrimVar lengthRef
+    force <$> Mutable.unsafeFreeze (Mutable.take length buffer)
```
jorisdral, line 117: If I remember correctly, `force` will always copy the contents into a new vector. As such, this implementation is actually a safe `freeze`

src/Database/LSMTree/Internal/Vector/Growing.hs
```diff
@@ -0,0 +1,117 @@
...
+
+-- | Creates a new, initially empty, vector.
+new :: Int                      -- ^ Initial buffer size
+    -> ST s (GrowingVector s a) -- ^ Construction of the vector
+new illegalInitialBufferSize | illegalInitialBufferSize <= 0
+    = error "Initial buffer size not positive"
+new initialBufferSize
+    = do
+        buffer <- Mutable.new initialBufferSize
+        bufferRef <- newSTRef buffer
```
jorisdral, line 49:
```suggestion
        bufferRef <- newSTRef $! buffer
```

src/Database/LSMTree/Internal/Vector/Growing.hs
```diff
@@ -0,0 +1,117 @@
...
+    = do
+          length <- readPrimVar lengthRef
+          makeRoom
+          buffer' <- readSTRef bufferRef
+          Mutable.set (Mutable.slice length count buffer') val
+    where
+
+    makeRoom :: ST s ()
+    makeRoom = do
+        length <- readPrimVar lengthRef
```
jorisdral, line 70: This length was just read in the `do` block above

src/Database/LSMTree/Internal/Vector/Growing.hs
```diff
@@ -0,0 +1,117 @@
...
+            !bufferSize = Mutable.length buffer
+
+            length' :: Int
+            !length' = length + count
+
+        when (bufferSize < length') $ do
+            let
+
+                higherBufferSizes :: [Int]
+                higherBufferSizes = tail (init ++ [last]) where
```
jorisdral, line 85: I think `maxBound` should be part of the higher buffer sizes as well. `last` could be much less than `maxBound`

src/Database/LSMTree/Internal/Vector/Growing.hs
```diff
@@ -0,0 +1,117 @@
...
+
+                sufficientBufferSizes :: [Int]
+                sufficientBufferSizes = dropWhile (< length') higherBufferSizes
+
+            case sufficientBufferSizes of
+                []
+                    -> error "No sufficient buffer size available"
+                bufferSize' : _
+                    -> Mutable.grow buffer (bufferSize' - bufferSize) >>=
+                       writeSTRef bufferRef
```
jorisdral, line 105:
```suggestion
                    -> Mutable.grow buffer (bufferSize' - bufferSize) >>=
                       (writeSTRef bufferRef $!)
```

jorisdral (changes requested):
Looks great, just a few small suggestions

src/Database/LSMTree/Internal/Vector/Growing.hs
jeltsch: I had wondered about this already before sending you this pull request. I have checked now: From a quick look at the code, I get indeed the impression that `force` will always copy the contents. However, the API documentation says this:

> Yield the argument, but force it not to retain any extra memory, possibly by copying it.

So the specification leaves open the possibility of `force` returning the original vector if this vector uses its underlying array in full. We might hope that future versions of the `vector` package won’t make use of this opportunity, but I’d like to err on the safe side and continue considering the above function unsafe.

src/Database/LSMTree/Internal/Vector/Growing.hs
jeltsch: Hmm, I don’t think this is necessary. Since we’re using the strict version of `ST`, the computation `Mutable.new initialBufferSize` will definitely be performed before `newSTRef buffer`, and I’d be very surprised if `new` wouldn’t always yield a fully evaluated result. Thoughts?

src/Database/LSMTree/Internal/Vector/Growing.hs
jeltsch: You’re absolutely right, and I was fully aware of that when I filed this pull request. 🙂 I’m not quite sure how to handle this situation. I would like to have exactly the code for “making room” be separated, so that it doesn’t clutter the main code. However, a self-contained implementation for making room needs to contain the initial reading of the length, which fulfills a different purpose than the reading of the length in the main code, which is for saving the original length in order to use it later in a computation. An alternative would be to pass the length as an argument to `makeRoom`, but this seems a bit awkward to me. Would you be fine with keeping the code as it is now? Do you perhaps have a different idea how to deal with this problem?

src/Database/LSMTree/Internal/Vector/Growing.hs
jeltsch: I had considered this as well but then dropped this idea. The thing is that, while `maxBound` could be as “low” as `2 ^ 27 - 1` according to the Haskell Report, in practice it is likely to be `2 ^ 63 - 1` and thus so high that the number of pages will probably never get anywhere close to it. With the current implementation, the last element of `higherBufferSizes` is at least `2 ^ 62` and thus still so high that it will probably never be needed. I think that overflow of buffer sizes should be prevented even if it is unlikely to ever occur, simply to be safe and for the code to be correct. However, I don’t think we have to introduce the extra, albeit small, complexity that comes from adding `maxBound` to `higherBufferSizes`, given that doing so probably wouldn’t make any difference in practice. What are your thoughts on this?

src/Database/LSMTree/Internal/Vector/Growing.hs
jeltsch: I think that this is not needed, for the same reasons as in [the invocation of `newSTRef` above](https://github.com/IntersectMBO/lsm-tree/pull/419#discussion_r1789945134).

src/Database/LSMTree/Internal/Vector/Growing.hs
jorisdral: I've created https://github.com/haskell/vector/issues/511 about the specification of `force`.

Different suggestion: use `freeze` instead of `force`+`unsafeFreeze`. The result is the same, and it's less ambiguous whether there is a copy or not. And then we can call the function `freeze` instead of `unsafeFreeze`

jeltsch: Indeed, `freeze` should always generate a copy, and, while its documentation doesn’t specifically mention that it copies only the relevant slice, it would be weird for it not to do so, and the current implementation seems to do it. I’ll change the code to use `freeze` instead of `force` and `unsafeFreeze`.

jeltsch: Done in 49eb151ca766b6c6a33685753b715bd4b1d304d2.

src/Database/LSMTree/Internal/Vector/Growing.hs
jeltsch: Okay. Fixed in d0cc37980fd55b9744549005340726080ca24a3a.

src/Database/LSMTree/Internal/Vector/Growing.hs
jeltsch: Given your comment to my other `$!`-related remark, I’ve made the suggested change here as well, also as part of d0cc37980fd55b9744549005340726080ca24a3a.

src/Database/LSMTree/Internal/Vector/Growing.hs
jeltsch: Done in 65fed7662d71d56d1e90f5f667a626a26efae12b.

jorisdral (approved):

jorisdral:
Let's squash the commits, and then we can merge

PR #424, opened 2024-10-08
Fix adding/removing references on table content

jorisdral:
In some places, we were adding/removing references but not taking into account rollbacks/delays that should be performed when exceptions occur. In particular, correct acquisition and release of write buffer blob files were not guarded by `allocateTemp` and `freeTemp`.

In addition, to make our reference accounting slightly less error prone, we make a few related changes:

* We add new  `addReferenceTableContent` and `removeReferenceTableContent` functions, which replace most occurrences of manually increasing/decreasing reference counts for the components of a  `TableContent`. For example, in `duplicate` and `close` we now use the new functions.

* `newWith` now uses a `TempRegistry`, which fixes a double-free TODO.

* `newWith` now requires a full `TableContent` to be passed in.

### DIFF
--- a/src/Database/LSMTree/Internal.hs
+++ b/src/Database/LSMTree/Internal.hs
@@ -105,9 +105,7 @@ import           Database.LSMTree.Internal.Serialise (SerialisedBlob (..),
                      SerialisedKey, SerialisedValue)
 import           Database.LSMTree.Internal.UniqCounter
 import qualified Database.LSMTree.Internal.Vector as V
-import           Database.LSMTree.Internal.WriteBuffer (WriteBuffer)
 import qualified Database.LSMTree.Internal.WriteBuffer as WB
-import           Database.LSMTree.Internal.WriteBufferBlobs (WriteBufferBlobs)
 import qualified Database.LSMTree.Internal.WriteBufferBlobs as WBB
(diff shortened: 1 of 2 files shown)


src/Database/LSMTree/Internal.hs
```diff
@@ -1390,21 +1404,12 @@ duplicate th = do
...
-                (\_ -> Run.removeReference r)
+            addReferenceTableContent reg content
             pure content
-          WBB.addReference (tableWriteBufferBlobs content)
-          -- TODO: Fix possible double-free! See 'newCursor'.
-          -- In `newWith`, the table handle (in the open state) gets added to
-          -- `sessionOpenTables', even if later an async exception occurs and
-          -- the temp registry rolls back all allocations.
           newWith
+            reg
```
recursion-ninja, line 1410: I think it may be clearer to change lines 1397 and 1399 to `th{..}` and `\thEnv{..}`, respectively,  rather than explicitly calling the record accessor functions here.

recursion-ninja (changes requested):
Seems straight forward enough. 

I had one comment on record the potential use of record wildcards that might improve code clarity.

recursion-ninja (approved):
📤 LGTM (currently).

PR #430, opened 2024-10-14
Implement a more proper snapshot format

jorisdral:
The current snapshot format is only loosely established, so we add a more concrete snapshot format, which is more amenable to refactoring. This snapshot format is still temporary -- we will implement snapshots properly in a later package.

#426 is near to being ready, but it is currently still failing some tests related to snapshots. Establishing a snapshot format makes bug-hunting in the snapshot code easier.

### DIFF
--- a/lsm-tree.cabal
+++ b/lsm-tree.cabal
@@ -152,6 +152,7 @@ library
     Database.LSMTree.Internal.RunReaders
     Database.LSMTree.Internal.Serialise
     Database.LSMTree.Internal.Serialise.Class
+    Database.LSMTree.Internal.Snapshot
     Database.LSMTree.Internal.UniqCounter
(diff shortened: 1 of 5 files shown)


dcoutts (approved):
So unless I missed anything, this is mostly just moving the interim impl of snapshots from `.Internal` (and `.Config` and `.RunNumber`) into `.Snapshot`. Which is fine.

jorisdral:
> So unless I missed anything, this is mostly just moving the interim impl of snapshots from `.Internal` (and `.Config` and `.RunNumber`) into `.Snapshot`. Which is fine.

That, and the (interim) format of a snapshot is more explicit

mheinzel (approved):
That looks like a sensible improvement :+1:

PR #432, opened 2024-10-14
Split module internal

mheinzel:
It got quite large.

I just quickly made it compile, not sure I really just want to re-export the `Resource` type, but I think we want something like it, since just storing `m ()` gets annoying in the `NoThunks` instances.

Main points I would like to hear your opinions on:
1. Just storing the `Resource`s (cleanup actions) in the session (first commit).
2. Where to put and whether to split the `Trace` and `Exception` types. Currently each part (Cursor, TableHandle, Snapshot, ...) knows about all the trace messages and exceptions, so splitting them and then nesting them all into a sum type at the end might make it more modular.


recursion-ninja (approved):
📤LGTM.

I noticed a few type signatures had the `PrimMonad m` constraint removed. Do you think the lack of these constraints will negatively effect performance in some way? If so I imagine it's nothing the `SPECIALISE` pragmas won't be able to recitify.

jorisdral (changes requested):
I'd like to have a look at this PR before we merge it -- I'll review soon

mheinzel:
> I noticed a few type signatures had the PrimMonad m constraint removed.

A few constraints moved around because the point where e.g. `closeCursor` is referenced (not executed!) changed, so its constraints now show up in a different part of the code. That doesn't affect the `m` in any way, but there is a small performance impact because of doing an unknown call to the cleanup action.

jorisdral (changes requested):
Thanks for the work! Did anything about the code change other than it being moved?

Also, I hope you don't mind, but before I approve I'd like to start a discussion about the structuring philosophy. 

What I like about the current structure is that we have the `Internal` module defining the important surface types and implementing the public API, while most of the interesting logic lives in other modules (like `Lookups`, `MergeSchedule`, `Snapshot`). Many of the API implementations contain relatively little interesting logic except for things like concurrency arbitration. Arguably, we are not following this structure super strictly, because we could still move more interesting logic out of the `Internal` module.  

Maybe you could give some insight into why you'd prefer the new structure? Maybe we can find a middle ground between the current and new structure?

EDIT: also, I just want to clarify I'm willing to merge the PR, I just want to resolve our differing ideas first

mheinzel:
> Thanks for the work! Did anything about the code change other than it being moved?

Just breaking the dependency in the first commit. The second commit is just moving code and adjusting the imports accordingly.

> Also, I hope you don't mind, but before I approve I'd like to start a discussion about the structuring philosophy.

That makes sense and I explicitly wanted feedback, since I'm not super convinced this approach myself.

> What I like about the current structure is that we have the `Internal` module defining the important surface types and implementing the public API, while most of the interesting logic lives in other modules (like `Lookups`, `MergeSchedule`, `Snapshot`). Many of the API implementations contain relatively little interesting logic except for things like concurrency arbitration.

I agree, I like having the API types and signatures in a single module. Gives a nice overview instead of having to puzzle it together across multiple modules, especially when there is currently no distinction between top-level-like modules like `Cursor` and properly internal modules like `Chunk`.

> Arguably, we are not following this structure super strictly, because we could still move more interesting logic out of the `Internal` module.

I think the main offender is currently `Cursor` and for some other parts it would be nice to move a little more code out of the module.
This can be a bit tricky, but I don't hate the current state of things. Having a large module in itself doesn't have to be bad, as long as the overall code structure is clear.

So maybe a good way forward would the be to just close this PR, I open a separate one just moving some Cursor-related logic, and we generally try to keep most of the logic in separate modules?

mheinzel:
@jorisdral See #434. I now think it's overall nicer than what I have here, so I think I'll close this PR?

jorisdral:
> @jorisdral See #434. I now think it's overall nicer than what I have here, so I think I'll close this PR?

If you're happy with what we have now in #434, then we can close this PR. If we do decide later that we want to restructure a little bit more than #434, we can use #432 as inspiration

PR #433, opened 2024-10-14
Fix a race condition in `RWVar`

jorisdral:
`retry` is interruptible, which can leave the `RWVar` in a state that is unrecoverable.

Kudos to @jasagredo for finding this subtle bug!

### DIFF
--- a/src-control/Control/Concurrent/Class/MonadSTM/RWVar.hs
+++ b/src-control/Control/Concurrent/Class/MonadSTM/RWVar.hs
@@ -75,47 +75,85 @@ withReadAccess rwvar k =
       k
 
 {-# SPECIALISE unsafeAcquireWriteAccess :: RWVar IO a -> IO a #-}
-unsafeAcquireWriteAccess :: MonadSTM m => RWVar m a -> m a
-unsafeAcquireWriteAccess rw@(RWVar !var) = do
-    may <- atomically $ readTVar var >>= \case
-      Reading n x
-        | n == 0 -> do
-            writeTVar var Writing
-            pure (Just x)
-        | otherwise -> do
-            writeTVar var (WaitingToWrite n x)
-            pure Nothing
-      WaitingToWrite n x
-        | n == 0 -> do
-            writeTVar var Writing
-            pure (Just x)
-        | otherwise -> retry
-      -- If we see this, someone else has already acquired a write lock, so we retry.
-      Writing -> retry
-    -- When the previous STM transaction returns Nothing, it signals that we set
-    -- the RWState to WaitingToWrite. All that remains is to try acquiring write
-    -- access again.
+-- | Acquire write access. This function assumes that it runs in a masked
+-- context, and that is properly paired with an 'unsafeReleaseWriteAccess'!
+--
+-- If multiple threads try to acquire write access concurrently, then they will
+-- race for access. However, if a thread has set RWState to WaitingToWrite, then
+-- it is guaranteed that the same thread will acquire write access when all
+-- readers have finished. That is, other writes can not "jump the queue". When
+-- the writer finishes, then all other waiting threads will race for write
+-- access again.
+--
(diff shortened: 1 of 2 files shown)


jorisdral:
This is now ready for review. My takeaway is that I should put more thought into interruptibility when writing exception safe code :stuck_out_tongue_closed_eyes:

test-control/Test/Control/Concurrent/Class/MonadSTM/RWVar.hs
```diff
@@ -17,44 +19,71 @@ tests = testGroup "Test.Control.Concurrent.Class.MonadSTM.RWVar" [
...
+
+resultsMonotonic :: [Action Int] -> Bool
+resultsMonotonic = go 0
+  where
+    go _ []               = True
+    go prev (Read x : as) = prev == x     && go x as
+    go prev (Incr x : as) = prev + 1 == x && go x as
+
+actionResults :: SimTrace () -> [Action Int]
+actionResults simTrace = mapMaybe eventSayAction (traceEvents simTrace)
```
jasagredo, line 85: You can use [`selectTraceEventsSay`](https://input-output-hk.github.io/io-sim/io-sim/Control-Monad-IOSim.html#v:selectTraceEventsSay)

test-control/Test/Control/Concurrent/Class/MonadSTM/RWVar.hs
```diff
@@ -17,44 +19,71 @@ tests = testGroup "Test.Control.Concurrent.Class.MonadSTM.RWVar" [
 data Action a = Read a | Incr a
```
jasagredo, line 19: Any particular reason why this is polymorphic on `a` if we use it with `()` later on?

src-control/Control/Concurrent/Class/MonadSTM/RWVar.hs
```diff
@@ -1,3 +1,5 @@
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
```
jasagredo, line 1: This is probably a leftover?

test-control/Test/Control/Concurrent/Class/MonadSTM/RWVar.hs
jorisdral: Done

test-control/Test/Control/Concurrent/Class/MonadSTM/RWVar.hs
jorisdral: It doubles as the result of reading `say` messages as well, in which case it holds `Int`

src-control/Control/Concurrent/Class/MonadSTM/RWVar.hs
jorisdral: Removed

test-control/Test/Control/Concurrent/Class/MonadSTM/RWVar.hs
```diff
@@ -17,44 +19,71 @@ tests = testGroup "Test.Control.Concurrent.Class.MonadSTM.RWVar" [
...
+                let x' = x + 1
+                say (show (Incr x'))
+                pure x'
+      t1 <- async (g a1)
+      t2 <- async (g a2)
+      t3 <- async (g a3)
+      async (cancel t1) >>= wait
+      (_ :: Either AsyncCancelled ()) <- try (wait t1)
+      (_ :: Either AsyncCancelled ()) <- try (wait t2)
+      (_ :: Either AsyncCancelled ()) <- try (wait t3)
```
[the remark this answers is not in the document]
jorisdral: True, but `waitCatch` catches `SomeException`, whereas I'm only interested in `AsyncCancelled` here. I could then match on `SomeException` to see if it is `AsyncCancelled`, but it won't buy us much

dcoutts (approved):
LGTM

PR #434, opened 2024-10-17
move Cursor read logic into separate module

mheinzel:
# Description

Instead of splitting the whole module as proposed in #432, we might want to just move more logic out of the `Internal` module (without API types like `Cursor` and `TableHandle`). I also documented that approach in the module description.

### DIFF
--- a/src/Database/LSMTree/Internal.hs
+++ b/src/Database/LSMTree/Internal.hs
@@ -1,5 +1,16 @@
 {-# LANGUAGE DataKinds #-}
 
+-- | This module brings together the internal parts to provide an API in terms
+-- of untyped serialised keys, values and blobs. It makes no distinction between
+-- normal and monoidal tables, accepting both blobs and mupserts.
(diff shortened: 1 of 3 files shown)


src/Database/LSMTree/Internal.hs
```diff
@@ -1,5 +1,12 @@
 {-# LANGUAGE DataKinds #-}
 
+-- | The internal API here is defined in terms of untyped serialised keys,
+-- values and blobs. It makes no distinction between normal and monoidal tables,
+-- accepting both blobs and mupserts.
+--
+-- This module mainly deals with concurrency- and exception-safe opening and
+-- closing of resources. Any other non-trivial logic should live somewhere else.
+--
```
jorisdral, line 9: Nice, it is indeed helpful if we record the intention of the module here. Maybe we should also note here that this module is, in a way, bringing all internal parts together to implement the public API.

jorisdral (approved):

PR #435, opened 2024-10-22
Do merge work in batches

jorisdral:
This is a follow-up to #426

The best description of the approach can be found in the code, in the GHC-style note `Note [Credits]`

### DIFF
--- a/src/Database/LSMTree/Internal/MergeSchedule.hs
+++ b/src/Database/LSMTree/Internal/MergeSchedule.hs
@@ -921,7 +921,7 @@ supplyMergeCredits (ScaledCredits c) (MergingRun _ _ var) = do
         (stepsDone, stepResult) <- Merge.steps m stepsToDo
         assert (case stepResult of
                   MergeInProgress -> stepsDone >= stepsToDo
-                  MergeComplete   -> True
+                  MergeDone       -> True
                ) $ pure ()
 
         -- This should be the only point at which we write to these variables.
@@ -939,10 +939,10 @@ supplyMergeCredits (ScaledCredits c) (MergingRun _ _ var) = do
         writePrimVar totalCreditsVar $! totalCredits + c
         assert (case stepResult of
                   MergeInProgress -> totalSteps' >= totalCredits'
-                  MergeComplete   -> True
+                  MergeDone       -> True
                ) $ pure ()
 
-        pure $ stepResult == MergeComplete
+        pure $ stepResult == MergeDone
     when mergeIsDone $
       modifyMVarMasked_ var $ \case
         mr@CompletedMerge{} -> pure $! mr
--- a/src/Database/LSMTree/Internal/MergeSchedule.hs
+++ b/src/Database/LSMTree/Internal/MergeSchedule.hs
@@ -1,6 +1,8 @@
 {-# LANGUAGE CPP       #-}
 {-# LANGUAGE DataKinds #-}
 
+{- HLINT ignore "Use when" -}
+
 -- TODO: establish that this implementation matches up with the ScheduledMerges
 -- prototype. See lsm-tree#445.
 module Database.LSMTree.Internal.MergeSchedule (
@@ -19,7 +21,11 @@ module Database.LSMTree.Internal.MergeSchedule (
   , Level (..)
   , MergingRun (..)
   , NumRuns (..)
+  , UnspentCreditsVar (..)
   , MergingRunState (..)
+  , TotalStepsVar (..)
+  , SpentCreditsVar (..)
+  , MergeKnownCompleted (..)
     -- * Flushes and scheduled merges
   , updatesWithInterleavedFlushes
   , flushWriteBuffer
@@ -27,23 +33,27 @@ module Database.LSMTree.Internal.MergeSchedule (
   , MergePolicyForLevel (..)
   , maxRunSize
     -- * Credits
+  , Credit (..)
   , supplyCredits
   , ScaledCredits (..)
   , supplyMergeCredits
+  , CreditThreshold (..)
+  , creditThresholdForLevel
   ) where
 
 import           Control.Concurrent.Class.MonadMVar.Strict
-import           Control.Monad (when)
+import           Control.Monad (void, when)
 import           Control.Monad.Class.MonadST (MonadST)
 import           Control.Monad.Class.MonadSTM (MonadSTM (..))
-import           Control.Monad.Class.MonadThrow (MonadCatch, MonadMask,
-                     MonadThrow (..))
+import           Control.Monad.Class.MonadThrow (MonadCatch (bracketOnError),
+                     MonadMask, MonadThrow (..))
 import           Control.Monad.Fix (MonadFix)
 import           Control.Monad.Primitive
 import           Control.RefCount (RefCount (..))
 import           Control.TempRegistry
 import           Control.Tracer
 import           Data.BloomFilter (Bloom)
+import           Data.Primitive.MutVar
 import           Data.Primitive.PrimVar
 import qualified Data.Vector as V
 import           Database.LSMTree.Internal.Assertions (assert,
@@ -105,8 +115,6 @@ data MergeTrace =
       RunNumber
   | TraceNewMergeSingleRun
       NumEntries -- ^ Size of run
-      RunNumber
-  | TraceExpectCompletedMergeSingleRun
       RunNumber
   deriving stock Show
 
@@ -270,12 +278,28 @@ data Level m h = Level {
 -- | A merging run is either a single run, or some ongoing merge.
 data MergingRun m h =
     -- | A merging of multiple runs.
-    MergingRun !MergePolicyForLevel !NumRuns !(StrictMVar m (MergingRunState m h))
+    MergingRun
+      !MergePolicyForLevel
+      !NumRuns
+      !NumEntries
+      -- ^ Sum of number of entries in the input runs
+      !(UnspentCreditsVar (PrimState m))
+      -- ^ The number of currently /unspent/ credits
+      !(TotalStepsVar (PrimState m))
+      -- ^ The total number of performed merging steps.
+      !(MutVar (PrimState m) MergeKnownCompleted)
+      -- ^ A variable that caches knowledge about whether the merge has been
+      -- completed. If 'MergeKnownCompleted', then we are sure the merge has
+      -- been completed, otherwise if 'MergeMaybeCompleted' we have to check the
+      -- 'MergingRunState'.
+      !(StrictMVar m (MergingRunState m h))
     -- | The result of merging a single run, is a single run.
   | SingleRun !(Run m h)
 
 newtype NumRuns = NumRuns { unNumRuns :: Int }
   deriving stock (Show, Eq)
+
+newtype UnspentCreditsVar s = UnspentCreditsVar { getUnspentCreditsVar :: PrimVar s Int }
 
 data MergingRunState m h =
     CompletedMerge
@@ -284,11 +308,16 @@ data MergingRunState m h =
   | OngoingMerge
       !(V.Vector (Run m h))
       -- ^ Input runs
-      !(PrimVar (PrimState m) Int)
-      -- ^ The total number of performed merging steps.
-      !(PrimVar (PrimState m) Int)
-      -- ^ The total number of supplied credits.
+      !(SpentCreditsVar (PrimState m))
+      -- ^ The total number of spent credits.
       !(Merge m h)
+
+newtype TotalStepsVar s = TotalStepsVar { getTotalStepsVar ::  PrimVar s Int  }
+
+newtype SpentCreditsVar s = SpentCreditsVar { getSpentCreditsVar :: PrimVar s Int }
+
+data MergeKnownCompleted = MergeKnownCompleted | MergeMaybeCompleted
+  deriving stock (Show, Eq, Read)
 
 {-# SPECIALISE addReferenceLevels :: TempRegistry IO -> Levels IO h -> IO () #-}
 addReferenceLevels ::
@@ -326,9 +355,9 @@ forRunAndMergeM_ ::
 forRunAndMergeM_ lvls k1 k2 = V.forM_ lvls $ \(Level mr rs) -> do
     case mr of
       SingleRun r    -> k1 r
-      MergingRun _ _ var -> withMVar var $ \case
+      MergingRun _ _ _ _ _ _ var -> withMVar var $ \case
         CompletedMerge r -> k1 r
-        OngoingMerge irs _ _ m -> V.mapM_ k1 irs >> k2 m
+        OngoingMerge irs _ m -> V.mapM_ k1 irs >> k2 m
     V.mapM_ k1 rs
 
 {-# SPECIALISE foldRunM ::
@@ -345,9 +374,9 @@ foldRunM ::
 foldRunM f x lvls = flip (flip V.foldM x) lvls $ \y (Level mr rs) -> do
     z <- case mr of
       SingleRun r -> f y r
-      MergingRun _ _ var -> withMVar var $ \case
+      MergingRun _ _ _ _ _ _ var -> withMVar var $ \case
         CompletedMerge r -> f y r
-        OngoingMerge irs _ _ _m -> V.foldM f y irs
+        OngoingMerge irs _ _m -> V.foldM f y irs
     V.foldM f z rs
 
 {-# SPECIALISE forRunM ::
@@ -364,6 +393,10 @@ forRunM ::
   -> m (V.Vector a)
 forRunM lvls k = do
     V.reverse . V.fromList <$> foldRunM (\acc r -> k r >>= \x -> pure (x : acc)) [] lvls
+
+{-# SPECIALISE iforLevelM_ :: Levels IO h -> (LevelNo -> Level IO h -> IO ()) -> IO () #-}
+iforLevelM_ :: Monad m => Levels m h -> (LevelNo -> Level m h -> m ()) -> m ()
+iforLevelM_ lvls k = V.iforM_ lvls $ \i lvl -> k (LevelNo (i + 1)) lvl
 
 {-------------------------------------------------------------------------------
   Flushes and scheduled merges
@@ -427,7 +460,7 @@ updatesWithInterleavedFlushes tr conf resolve hfs hbio root uc es reg tc = do
     -- number of supplied credits is based on the size increase of the write
     -- buffer, not the the number of processed entries @length es' - length es@.
     let numAdded = unNumEntries (WB.numEntries wb') - unNumEntries (WB.numEntries wb)
-    supplyCredits numAdded (tableLevels tc)
+    supplyCredits conf (Credit numAdded) (tableLevels tc)
     let tc' = tc { tableWriteBuffer = wb' }
     if WB.numEntries wb' < maxn then do
       pure $! tc'
@@ -672,7 +705,7 @@ addRunToLevels tr conf@TableConfig{..} resolve hfs hbio root uc r0 reg levels = 
         mr <- newMerge policyForLevel Merge.LastLevel ln rs
         return $ V.singleton $ Level mr V.empty
     go !ln rs' (V.uncons -> Just (Level mr rs, ls)) = do
-        r <- expectCompletedMerge ln mr
+        r <- expectCompletedMergeTraced ln mr
         case mergePolicyForLevel confMergePolicy ln ls of
           -- If r is still too small for this level then keep it and merge again
           -- with the incoming runs.
@@ -713,17 +746,9 @@ addRunToLevels tr conf@TableConfig{..} resolve hfs hbio root uc r0 reg levels = 
             mr' <- newMerge LevelLevelling Merge.LastLevel ln (rs' `V.snoc` r)
             pure $! Level mr' V.empty `V.cons` V.empty
 
-    expectCompletedMerge :: LevelNo -> MergingRun m h -> m (Run m h)
-    expectCompletedMerge ln (SingleRun r) = do
-      traceWith tr $ AtLevel ln $ TraceExpectCompletedMergeSingleRun (runNumber $ Run.runRunFsPaths r)
-      pure r
-    expectCompletedMerge ln (MergingRun _ _ var) = do
-      r <- withMVar var $ \case
-        CompletedMerge r -> pure r
-        OngoingMerge{} -> do
-          -- If the algorithm finds an ongoing merge here, then it is a bug in
-          -- our merge sceduling algorithm. As such, we throw a pure error.
-          error "expectCompletedMerge: expected a completed merge, but found an ongoing merge"
+    expectCompletedMergeTraced :: LevelNo -> MergingRun m h -> m (Run m h)
+    expectCompletedMergeTraced ln mr = do
+      r <- expectCompletedMerge mr
       traceWith tr $ AtLevel ln $ TraceExpectCompletedMerge (runNumber $ Run.runRunFsPaths r)
       pure r
 
@@ -751,8 +776,9 @@ addRunToLevels tr conf@TableConfig{..} resolve hfs hbio root uc r0 reg levels = 
                   Run.removeReference
             traceWith tr $ AtLevel ln $ TraceCompletedMerge (Run.runNumEntries r) (runNumber $ Run.runRunFsPaths r)
             V.mapM_ (freeTemp reg . Run.removeReference) rs
-            var <- newMVar $! CompletedMerge r
-            pure $! MergingRun mergepolicy (NumRuns $ V.length rs) var
+            mergeVar <- newMVar $! CompletedMerge r
+
+            mkMergingRun rs MergeKnownCompleted mergeVar
           Incremental -> do
             mergeMaybe <- allocateMaybeTemp reg
               (Merge.new hfs hbio caching alloc mergelast resolve runPaths rs)
@@ -760,10 +786,24 @@ addRunToLevels tr conf@TableConfig{..} resolve hfs hbio root uc r0 reg levels = 
             case mergeMaybe of
               Nothing -> error "newMerge: merges can not be empty"
               Just m -> do
-                totalStepsVar <- newPrimVar $! 0
-                totalCreditsVar <- newPrimVar $! 0
-                var <- newMVar $! OngoingMerge rs totalStepsVar totalCreditsVar m
-                pure $! MergingRun mergepolicy (NumRuns $ V.length rs) var
+                spentCreditsVar <- SpentCreditsVar <$> newPrimVar 0
+                mergeVar <- newMVar $! OngoingMerge rs spentCreditsVar m
+                mkMergingRun rs MergeMaybeCompleted mergeVar
+        where
+          mkMergingRun ::
+               V.Vector (Run m h)
+            -> MergeKnownCompleted
+            -> StrictMVar m (MergingRunState m h)
+            -> m (MergingRun m h)
+          mkMergingRun inputRuns mergeCompleted mergeVar = do
+            unspentCreditsVar <- UnspentCreditsVar <$> newPrimVar 0
+            totalStepsVar <- TotalStepsVar <$> newPrimVar 0
+            mergeKnownCompletedVar <- newMutVar $! mergeCompleted
+            let numInputEntries = NumEntries $
+                  V.sum (V.map (unNumEntries . Run.runNumEntries) inputRuns)
+            pure $! MergingRun
+                      mergepolicy (NumRuns $ V.length rs) numInputEntries
+                      unspentCreditsVar totalStepsVar mergeKnownCompletedVar mergeVar
 
 data MergePolicyForLevel = LevelTiering | LevelLevelling
   deriving stock (Show, Eq)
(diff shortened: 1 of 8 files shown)


src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -101,6 +103,9 @@ data MergeTrace =
       RunNumber
   | TraceExpectCompletedMerge
       RunNumber
+  | TraceExpectCompletedMerge'
+      RunNumber
+      Int
```
jorisdral, line 108: Note to self: should be removed

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -700,24 +713,24 @@ addRunToLevels tr conf@TableConfig{..} resolve hfs hbio root uc r0 reg levels =
...
-          -- a bug in our scheduling, since we take great care to ensure merges
-          -- always finish on time.
-          error "expectCompletedMerge: expected a completed merge, but found an ongoing merge"
+        OngoingMerge rs _ m -> do -- TODO: add warning
+            (_n0, r) <- Merge.stepsToCompletionCounted m 1024
+            RefCount n <- Merge.readRefCount m
+            let !n' = fromIntegralChecked n
+            V.forM_ rs $ \r20 -> Run.removeReferenceN r20 n'
+            Merge.removeReferenceN m n'
+            pure $! (CompletedMerge r, r)
```
jorisdral, line 733: Note to self

This part should be rewritten a little, so that we can disambiguate between a few cases:
* We're doing some remaining work because we didn't go over the credit threshold
* We're doing some remaining work because we accumulated credits but didn't perform the work
* We're expecting the merge to be finished within this number of credits

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -839,59 +854,194 @@ mergeRuns resolve hfs hbio caching alloc runPaths mergeLevel runs = do
   Credits
 -------------------------------------------------------------------------------}
 
-type Credit = Int
+{-
+  Note [Credits]
```
jorisdral, line 858: Should also note that it is okay if a merge finishes earlier than expected, as long as it does not finish *too* early

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -839,59 +854,194 @@ mergeRuns resolve hfs hbio caching alloc runPaths mergeLevel runs = do
...
+    -- Nothing can be returned if the accumulator has already gone below the
+    -- threshold, which may happen if another thread is concurrently doing the
+    -- same loop on the same PrimVar.
+    takeCredits :: Credit -> m (Maybe Credit)
+    takeCredits (Credit prev)
+      | prev < creditsThresh = pure Nothing
+      | otherwise = do
+          -- numThresholds is guaranteed to be >= 1
+          let numThresholds = prev `div` creditsThresh
+          prev' <- casInt creditsAccumVar prev (numThresholds * creditsThresh)
```
jorisdral, line 996: :facepalm: 
```suggestion
          prev' <- casInt creditsAccumVar prev (prev - numThresholds * creditsThresh)
```

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -839,59 +854,194 @@ mergeRuns resolve hfs hbio caching alloc runPaths mergeLevel runs = do
...
+          if prev' == prev then
+            pure (Just (Credit (numThresholds * creditsThresh)))
+          else
+            takeCredits (Credit prev')
```
jorisdral, line 1000: Also: should this reduce the accumulated credits to 0 instead of getting it just under the threshold?

jorisdral: It should not, I think. It makes batches of merge work relatively uniform if we do it like it is written above. I've elaborated on this with a comment in the code.

src/Database/LSMTree/Internal/MergeSchedule.hs
jorisdral: Done. We do work for the remaining unspent credits. If the merge is not finished then, we error.

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -717,12 +729,22 @@ addRunToLevels tr conf@TableConfig{..} resolve hfs hbio root uc r0 reg levels =
...
+    expectCompletedMerge ln (MergingRun _ _ unspentCreditsVar mergeVar) = do
+      -- If there are remaining unspent credits that are below the threshold,
+      -- then we might not have stepped the merge to completion yet. We take the
+      -- remaining unspent credits and we /try/ to finish the merge.
+      isMergeDone <-
+        bracketOnError
+          (takeAllUnspentCredits unspentCreditsVar)
+          (putBackUnspentCredits unspentCreditsVar)
+          (stepMerge mergeVar)
+      when isMergeDone $ completeMerge mergeVar
```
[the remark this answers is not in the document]
mheinzel: (knowing how many total steps a merge requires would also help with making `scaleCreditsForMerge` distribute the work more evenly)

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -888,70 +1003,186 @@ scaleCreditsForMerge SingleRun{} _ = ScaledCredits 0
...
+    unspentCredits <- fetchAddInt unspentCreditsVar c
+    -- Compute our /estimate/ of what the new total of unspent credits is. The
+    -- /actual/ total might have been changed again by a different thread, but
+    -- that is okay because we will find out if\/when we run takeCredits.
+    let unspentCredits' = unspentCredits + c
+
+    -- Try to do some merging work if the new total now goes over the threshold.
+    --
+    -- TODO: when a merge is already completed, then could we skip most of the
+    -- stuff below?
```
mheinzel, line 1040: Finding that out would require reading an `MVar` every time (not just when going over the threshold), which sounds like it might end up being more expensive than credit accounting it might save. And in the common case when the merge is still ongoing it doesn't even save anything.

Still, it might be worth checking the performance impact, because reading the `mergeVar` would also allow us to to move `unspentCreditsVar` to an IMO more appropriate place and avoid the case where `expectCompletedMerge` still needs to do some merging (see other comments).

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -888,70 +1003,186 @@ scaleCreditsForMerge SingleRun{} _ = ScaledCredits 0
...
+  -> Credit
+  -> m (Maybe Credit)
+tryTakeUnspentCredits !unspentCreditsVar thresh@(CreditThreshold !creditsThresh) (Credit !prev)
+  | prev < creditsThresh = pure Nothing
+  | otherwise = do
+      -- numThresholds is guaranteed to be >= 1
+      let !numThresholds = prev `div` creditsThresh
+      prev' <- casInt unspentCreditsVar prev (prev - numThresholds * creditsThresh)
+      if prev' == prev then
+        pure (Just (Credit prev))
```
mheinzel, line 1090: This seems dubious. We reduce the unspent credits by `numThreshold * creditsThresh`, but then return `prev`, which might be larger than that.
So we end up doing `prev` steps, which should bring the unspent credits to 0, but leave some residue in the counter?

Why do we do that whole "multiple of the threshold" thing in the first place? The two options I see are
1. using all accumulated credits
2. doing only a single threshold of work

How do multiples of the threshold give more uniform amounts of merging work than option 1? Is it because of the idea of using co-prime numbers, so it doesn't mess up those?

Option 2 gives a stronger upper bound on work done for a single `supplyMergeCredits`, but due to concurrent use might in rare cases leave more credits around than the threshold. This should usually quickly be fixed by the next updates, but it's something to be careful about.
(edit: It would be problematic with `supplyMergeCredits` being called with a huge number when opening a snapshot, but that could just call `stepMerge` directly as it doesn't have to deal with concurrent management of credits)

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -888,70 +1003,186 @@ scaleCreditsForMerge SingleRun{} _ = ScaledCredits 0
...
+      else
+        tryTakeUnspentCredits unspentCreditsVar thresh (Credit prev')
+
+{-# SPECIALISE putBackUnspentCredits :: PrimVar RealWorld Int -> Credit -> IO () #-}
+putBackUnspentCredits ::
+     PrimMonad m
+  => PrimVar (PrimState m) Int
+  -> Credit
+  -> m ()
+putBackUnspentCredits !unspentCreditsVar (Credit !x) = void $ fetchAddInt unspentCreditsVar x
```
mheinzel, line 1100: So `fetchAddInt` always succeeds and can't spuriously fail like CAS? Seems like it from the documentation I see, but not totally sure.

mheinzel (commented):
Looks good, nice documentation! Just a few points to discuss.

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -270,7 +273,12 @@ data Level m h = Level {
 -- | A merging run is either a single run, or some ongoing merge.
 data MergingRun m h =
     -- | A merging of multiple runs.
-    MergingRun !MergePolicyForLevel !NumRuns !(StrictMVar m (MergingRunState m h))
+    MergingRun
+      !MergePolicyForLevel
+      !NumRuns
+      !(PrimVar (PrimState m) Int)
+      -- ^ The number of currently /unspent/ credits
```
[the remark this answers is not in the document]
jorisdral: > Any option you like? Do you want to change it here or should I do it as part of my PR later?

I'll probably add `Prim` instances even if it is a bit of boilerplate

>  Also, semantically it seems like it doesn't make to track unspent credits when a merge is already completed. This suggests that the counter should live in OngoingMerge.
> 
> Is the reason for it being here simply for performance, avoiding having to look into the MVar on the hot path? Might make sense to add a short comment explaining that.

This is largely explained by `Note [Credits]`: it's outside so that you don't have to acquire the more expensive MVar lock every time you supply credits. Instead, we use atomic operations on the `unspentCredits` variable for maximum concurrency. Putting the counter in `OngoingMerge` is not an option, i think. Semantically it's not too weird: spending unspent credits on a `CompletedMerge` is just a no-op, and spending credits on an `OngoingMerge` does some actual work

src/Database/LSMTree/Internal/MergeSchedule.hs
jorisdral: What do you mean by fail spuriously?

src/Database/LSMTree/Internal/MergeSchedule.hs
jorisdral: > Finding that out would require reading an MVar every time (not just when going over the threshold), which sounds like it might end up being more expensive than credit accounting it might save. And in the common case when the merge is still ongoing it doesn't even save anything.

That's not the only way to do it. You could add an `IORef Bool` or something similar that is `False` initially and only gets updated to `True` once the merge is completed. It would be a single read + a single check every time, instead of doing credit accounting.

> Still, it might be worth checking the performance impact, because reading the mergeVar would also allow us to to move unspentCreditsVar to an IMO more appropriate place and avoid the case where expectCompletedMerge still needs to do some merging (see other comments).

As mentioned before, I don't think that is the way to go. The fact that we have `unspentCreditsVar` outside of the `MVar` is exactly so we can have a (largely) lock-free design

src/Database/LSMTree/Internal/MergeSchedule.hs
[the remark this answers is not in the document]
mheinzel: Both points sound reasonable.

src/Database/LSMTree/Internal/MergeSchedule.hs
mheinzel: Makes sense, although adding redundant state would need to be justified by a noticeable performance improvement, which I don't think we'll see. So I'd probably leave it as a TODO to try later.

src/Database/LSMTree/Internal/MergeSchedule.hs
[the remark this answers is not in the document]
jorisdral: Ah, something I forgot is that `fetchAddInt` and co are defined only for `Int`s. I tried to get around that using `coerce` on for example `PrimVar s TotalCredits`, but it doesn't work very well. So instead I've gone with the slightly less nice thing of having `newtype` around the `PrimVar`s

src/Database/LSMTree/Internal/MergeSchedule.hs
jorisdral: I've elaborated the TODO a bit

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -848,23 +872,114 @@ mergeRuns resolve hfs hbio caching alloc runPaths mergeLevel runs = do
...
+
+  * The total of spent credits + the total of unspent credits is used by the
+  snapshot feature to restore merge work on snapshot load that was lost during
+  snapshot creation.
+
+  * For simplicity, merges are allowed to do more steps than request. However,
+  it does mean that once we do more steps next time a batch of work is done,
+  then we should account for the surplus of steps performed by the previous
+  batch. The total of spent credits + the number of performed merge steps is
+  used to compute this surplus, and adjust for it.
```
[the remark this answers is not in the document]
jorisdral: I'll add a TODO to reconsider at a later point in time whether we want to keep this

src/Database/LSMTree/Internal/MergeSchedule.hs
mheinzel: I don't know all of the details, but CAS instructions can sometimes fail even if the current value matches the expected one (at least on non-x86). There are often different versions exposed (the strong one compiling to a loop on systems where spurious failures can happen), see e.g. https://en.cppreference.com/w/cpp/atomic/atomic_compare_exchange. But I haven't found documentation on the Haskell implementation. With its API it seems like one wouldn't be able to detect a spurious failure, so I hope it uses the strong version.

Either way, nothing similar is mentioned in https://en.cppreference.com/w/cpp/atomic/atomic/fetch_add, so I think `fetchAddInt` does not have to consider spurious failure.

jorisdral: I followed the definitions up to the implementation of the primitive operation underlying `casInt`:

 `casInt -> casIntArray# -> CasByteArrayOp_Int -> __atomic_compare_exchange_n`
 
 And this is the code which seems to implementing the primitive:
 
 https://gitlab.haskell.org/ghc/ghc/-/blob/master/libraries/ghc-prim/cbits/atomic.c?ref_type=heads#L276
 
 `__atomic_compare_exchange_n` is called with `weak=false`, so it is using the strong version. Interesting rabbit hole!

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -839,59 +854,194 @@ mergeRuns resolve hfs hbio caching alloc runPaths mergeLevel runs = do
...
+creditsForMerge (MergingRun LevelTiering (NumRuns n) _ _) = fromIntegral n / 4
+
+newtype CreditThreshold = CreditThreshold Int
+
+-- TODO: the thresholds for doing merge work should be different for each level,
+-- maybe co-prime?
+creditThresholdForLevel :: TableConfig -> LevelNo -> CreditThreshold
+creditThresholdForLevel conf (LevelNo _i) =
+    let AllocNumEntries (NumEntries x) = confWriteBufferAlloc conf
+    in  CreditThreshold x
```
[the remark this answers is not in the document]
jorisdral: Leaving it for a different PR

src/Database/LSMTree/Internal/MergeSchedule.hs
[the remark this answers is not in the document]
jorisdral: Regardless of the question whether `expectCompletedMerge` should do the last bit of merging work, I agree that keeping track of the input sizes would make it possible to better distribute merge work over time

I agree that it would be nice to have the prototype and implementation be in sync, though I'm going to play devil's advocate here: is it not the prototype that should change? If we care about dividing batches of work evenly over time, then it actually might make more sense to have `expectCompletedMerge` do the last bit of work (which is below the threshold) instead of having the last update do more work (above the threshold)

src/Database/LSMTree/Internal/MergeSchedule.hs
mheinzel: > Calling supplyMergeCredits has the advantage of restoring the merge state and all the counters to exactly what it was when the snapshot was taken. We can probably do it with just stepMerge as well

I think `stepMerge` deals with everything but tracking the unspent credits, which arguable could just be set to 0. We are doing a big batch of merging already and don't have latency or concurrency constraints here. So why not spend all available credits? Seems like the simpler solution with less overhead.

src/Database/LSMTree/Internal/MergeSchedule.hs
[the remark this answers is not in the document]
mheinzel: > having the last update do more work (above the threshold)

The last update wouldn't do more work! The proposed case would trigger when we haven't reached the threshold yet, but it's already enough to complete the merge. So it's less merging work than a usual batch.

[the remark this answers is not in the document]
mheinzel: Doing it in `expectCompletedMerge` even might mean this happens on every level (for a cascade of levels getting full), each potentially doing a little less than a batch, which together could also lead to a little spike.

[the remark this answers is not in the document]
jorisdral: Done

src/Database/LSMTree/Internal/MergeSchedule.hs
jorisdral: I think eventually we should store merge progress on disk when snapshotting, such that we don't have to redo the work. I've seen in the WP8 benchmark that it is rather slow to restore progress. The current implementation has close to the same effect as storing merge progress on disk, so I'd rather keep the impl like this

mheinzel: > I've seen in the WP8 benchmark that it is rather slow to restore progress.

I think that's acceptable, but would be nice to avoid that work of course. The downside is that it can significantly increase the size of snapshots.

> The current implementation has close to the same effect as storing merge progress on disk, because the state is restored exactly the same, so I'd rather keep the impl like this.

Okay, makes sense. I can live with that.

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -888,70 +1019,221 @@ scaleCreditsForMerge SingleRun{} _ = ScaledCredits 0
...
+supplyMergeCredits
+  (ScaledCredits c)
+  thresh@(CreditThreshold creditsThresh)
+  (MergingRun _ _ numInputEntries unspentCreditsVar totalStepsVar mergeVar) = do
+    -- Add credits to the total of unspent credits
+    Credit unspentCredits <- addUnspentCredits unspentCreditsVar (Credit c)
+    -- Compute our /estimate/ of what the new total of unspent credits is. The
+    -- /actual/ total might have been changed again by a different thread, but
+    -- that is okay because we will find out if\/when we run takeCredits.
+    let unspentCredits' = unspentCredits + c
```
mheinzel, line 1051: Should we move this addition into `addUnspentCredits`? It seems logical to be part of adding the credits to the counter and we don't need the previous unspent credits for anything else.

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -888,70 +1019,221 @@ scaleCreditsForMerge SingleRun{} _ = ScaledCredits 0
...
+        bracketOnError (takeAllUnspentCredits unspentCreditsVar)
+                       (putBackUnspentCredits unspentCreditsVar)
+                       (stepMerge mergeVar totalStepsVar)
+      when isMergeDone $ completeMerge mergeVar
+    -- We can do some merging work without finishing the merge immediately
+    else do
+      -- Try to do some merging work if the new total now goes over the threshold.
+      --
+      -- TODO: when a merge is already completed, could we skip most of the code
+      -- below? Is it worth it/does it provide a performance benefit?
```
mheinzel, line 1070: I think this has been addressed.

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -888,70 +1019,221 @@ scaleCreditsForMerge SingleRun{} _ = ScaledCredits 0
...
+    -- The merge is already finished
+    if stepsDone >= unNumEntries numInputEntries then
+      pure ()
+    -- We can finish the merge immediately
+    else if stepsDone + unspentCredits' >= unNumEntries numInputEntries then do
+      isMergeDone <-
+        bracketOnError (takeAllUnspentCredits unspentCreditsVar)
+                       (putBackUnspentCredits unspentCreditsVar)
+                       (stepMerge mergeVar totalStepsVar)
+      when isMergeDone $ completeMerge mergeVar
```
mheinzel, line 1064: Also, would it be better switch the two branches, i.e. to first check whether we are over the threshold? If there are more credits than the threshold, we probably only want to do one uniform batch, even if we could complete the merge.

mheinzel: Also, should we `assert isMergeDone` here? Ah no, it could be that some other thread already took the credits (leaving us with 0) and completed (or will complete) the merge. Interesting.

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -888,70 +1019,221 @@ scaleCreditsForMerge SingleRun{} _ = ScaledCredits 0
...
+    -- Compute our /estimate/ of what the new total of unspent credits is. The
+    -- /actual/ total might have been changed again by a different thread, but
+    -- that is okay because we will find out if\/when we run takeCredits.
+    let unspentCredits' = unspentCredits + c
+
+    stepsDone <- readPrimVar (getTotalStepsVar totalStepsVar)
+
+    -- The merge is already finished
+    if stepsDone >= unNumEntries numInputEntries then
+      pure ()
```
mheinzel, line 1057: I'm trying to think about how we can *really* be confident that we can't for some reason have an unfinished merge, despite the number of steps performed matching the amount of work we had to do. :thinking: 

What happens, for example, if an async exception happens between `stepMerge` (returning True) and `completeMerge`? We did all the steps, but the merge will still remain ongoing and never be touched again.

mheinzel (commented):
Looks good, but some additional questions.

src/Database/LSMTree/Internal/MergeSchedule.hs
mheinzel: > won't that mean there is a possibility that we don't finish the merge before the next expectCompletedMerge?

Ah, that's true. Then we probably need to merge the full remainder here even if it's more than a usual batch, which shouldn't happen often.

src/Database/LSMTree/Internal/MergeSchedule.hs
mheinzel: That's all true and the merge does get completed in the sense of doing all merge steps, but it can end up in a state where `Merge.complete` never gets called.

The scenario I am talking about it this:
- `stepMerge` succeeds and returns `True`, so `stepsDone` has been increased and the steps have actually been done, reaching the end of the merge (so `stepsDone` indeed does not exceed the actual amount of work done).
- an exception happens before `completeMerge` runs, so we remain in `OngoingMerge` state.
- `flushWriteBuffer` is not called (as you say)
- But in the *next update*, we enter `supplyMergeCredits` and find that `stepsDone` already must have been enough to complete the merge. So we simply call `pure ()`.
- When later `expectCompleteMerge` runs, it finds `OngoingMerge` (inside of which is a merge that has run to completion, but never been properly `complete`d). So it fails.

jorisdral: Good point, that's indeed a case where it could find an `OngoingMerge`. So, maybe the solution is then to have a single mutable location holding a `False` bool that is set to `True` after `completeMerge`, and use that bool in this if condition

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -270,26 +278,50 @@ data Level m h = Level {
...
+      -- ^ The total number of spent credits.
       !(Merge m h)
 
+newtype TotalStepsVar s = TotalStepsVar { getTotalStepsVar ::  PrimVar s Int  }
+  deriving stock Eq
+
+newtype SpentCreditsVar s = SpentCreditsVar { getSpentCreditsVar :: PrimVar s Int }
+  deriving stock Eq
+
+data MergeCompleted = MergeCompleted | MergeNotCompleted
```
mheinzel, line 322: If we only use this type for this one purpose, maybe we can be more explicit: `MergeKnownCompleted = MergeKnownCompleted | MergePotentiallyNotCompleted` or `MergeMaybeCompleted`.

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -888,70 +1039,224 @@ scaleCreditsForMerge SingleRun{} _ = ScaledCredits 0
...
+supplyMergeCredits
+  (ScaledCredits c)
+  thresh@(CreditThreshold creditsThresh)
+  (MergingRun _ _ numInputEntries unspentCreditsVar totalStepsVar mergeCompletedCacheVar mergeVar) = do
+
+    -- unspentCredits' is our /estimate/ of what the new total of unspent credits is.
+    Credit unspentCredits' <- addUnspentCredits unspentCreditsVar (Credit c)
+
+    totalSteps <- readPrimVar (getTotalStepsVar totalStepsVar)
+    mergeCompleted <- readMutVar mergeCompletedCacheVar
```
mheinzel, line 1071: "Cache" helps a bit, but I think we can make the meaning clearer by calling it something like `mergeKnownCompleted{Var}`, here and in the other places.

mheinzel (commented):
I think the naming can be clarified a bit, but should work!

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -270,26 +278,50 @@ data Level m h = Level {
...
 data MergingRun m h =
     -- | A merging of multiple runs.
-    MergingRun !MergePolicyForLevel !NumRuns !(StrictMVar m (MergingRunState m h))
+    MergingRun
+      !MergePolicyForLevel
+      !NumRuns
+      !NumEntries
+      -- ^ Sum of number of entries in the input runs
+      !(UnspentCreditsVar (PrimState m))
+      -- ^ The number of currently /unspent/ credits
```
mheinzel, line 287: To clarify, these are scaled credits, right?

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -888,70 +1039,224 @@ scaleCreditsForMerge SingleRun{} _ = ScaledCredits 0
...
+    else if totalSteps + unspentCredits' >= unNumEntries numInputEntries then do
+      isMergeDone <-
+        bracketOnError (takeAllUnspentCredits unspentCreditsVar)
+                       (putBackUnspentCredits unspentCreditsVar)
+                       (stepMerge mergeVar totalStepsVar)
+      when isMergeDone $ completeMerge mergeVar mergeKnownCompletedVar
+    -- We can do some merging work without finishing the merge immediately
+    else do
+      -- Try to do some merging work if the new total now goes over the threshold.
+      when (unspentCredits' >= creditsThresh) $ do
```
mheinzel, line 1086: Small style suggestion:

```suggestion
    else if unspentCredits' >= creditsThresh then do
```

And at the end `else pure ()  -- just accumulate credits`. Maybe more consistent with the other branches?

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -888,70 +1039,224 @@ scaleCreditsForMerge SingleRun{} _ = ScaledCredits 0
...
+                        (mapM_ (putBackUnspentCredits unspentCreditsVar)) $ \case
+            Nothing -> pure False
+            Just c' -> stepMerge mergeVar totalStepsVar c'
+
+        -- If we just finished the merge, then we convert the output of the merge
+        -- into a new run. i.e., we complete the merge.
+        --
+        -- If an async exception happens before we get to perform the completion,
+        -- then that is fine. Either the next supplyMergeCredits or
+        -- expectCompletedMerge will complete the merge.
```
mheinzel, line 1104: That's outdated, `expectCompletedMerge` will not complete anything.

mheinzel (approved):
I think this is alright now. Just some small tweaks.

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -848,23 +899,123 @@ mergeRuns resolve hfs hbio caching alloc runPaths mergeLevel runs = do
...
-     Credit
+     TableConfig
+  -> Credit
   -> Levels IO h
   -> IO ()
   #-}
+-- | Supply the given amount of credits to each merge in the levels structure.
+-- This /may/ cause some merges to progress.
+--
+-- See Note [Credits]
```
mheinzel, line 1007: This is not ideal in Haddocks, since the note doesn't appear there. Haddock has "named chunks" which we use in a few places (e.g. `$resource-management`), but we also have a few GHC-style notes.

mheinzel (commented):
I also think the scaled/unscaled credit distinction could be a little clearer in some places, but this will improve as soon as we move `MergingRun` into its own module. In there, we will always consider scaled credits. In `MergeSchedule` we talk about unscaled credits, only converting at the end to call `MergingRun.supplyMergeCredits`.

src/Database/LSMTree/Internal/MergeSchedule.hs
mheinzel: It's fine if the piece of documentation is more for internal consumption, as is the case here. But then we should probably make sure it's not referenced in Haddocks.

src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -182,18 +182,25 @@ data SnapLevel = SnapLevel {
   deriving stock (Show, Eq, Read)
 
 data SnapMergingRun =
-    SnapMergingRun !MergePolicyForLevel !NumRuns !SnapMergingRunState
+    SnapMergingRun !MergePolicyForLevel !NumRuns !NumEntries !UnspentCredits !MergeKnownCompleted !SnapMergingRunState
```
mheinzel, line 185: Do we need to snapshot the `MergeKnownCompleted`? It's like a cache, we can just reconstruct it from the `SnapMergingRunState`.

src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -182,18 +182,25 @@ data SnapLevel = SnapLevel {
   deriving stock (Show, Eq, Read)
```
mheinzel, line 182: Semi-related, but why do these types still have `Read` instances?

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -888,70 +1039,224 @@ scaleCreditsForMerge SingleRun{} _ = ScaledCredits 0
...
+
+    -- unspentCredits' is our /estimate/ of what the new total of unspent credits is.
+    Credit unspentCredits' <- addUnspentCredits unspentCreditsVar (Credit c)
+
+    totalSteps <- readPrimVar (getTotalStepsVar totalStepsVar)
+    mergeKnownCompleted <- readMutVar mergeKnownCompletedVar
+
+    -- The merge is already finished
+    if mergeKnownCompleted == MergeKnownCompleted then
+      pure ()
```
mheinzel, line 1075: Ah, I think we can skip this work too if `MergeKnownCompleted`.

src/Database/LSMTree/Internal/MergeSchedule.hs
jorisdral: Yes, they are

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -270,26 +278,50 @@ data Level m h = Level {
...
-      -- ^ The total number of supplied credits.
+      !(SpentCreditsVar (PrimState m))
+      -- ^ The total number of spent credits.
       !(Merge m h)
 
+newtype TotalStepsVar s = TotalStepsVar { getTotalStepsVar ::  PrimVar s Int  }
+  deriving stock Eq
+
+newtype SpentCreditsVar s = SpentCreditsVar { getSpentCreditsVar :: PrimVar s Int }
+  deriving stock Eq
```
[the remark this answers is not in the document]
jorisdral: I'll remove them

src/Database/LSMTree/Internal/Snapshot.hs
jorisdral: I've thought about this as well, and I think both with and without storing it here works. It just means we can skip a bit of work

src/Database/LSMTree/Internal/MergeSchedule.hs
jorisdral: I've removed the reference to the note. Might be good to at some point to describe in the haddocks just at a high level what supplying a credit means, but the note's contents are mostly implementation details so I think it's okay to keep as a note

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -888,70 +1039,224 @@ scaleCreditsForMerge SingleRun{} _ = ScaledCredits 0
...
+      (OngoingMerge rs _spentCreditsVar m) -> do
+        -- TODO: we'll likely move away from this style of reference counting,
+        -- so this code will change in the future.
+        RefCount n <- Merge.readRefCount m
+        let !n' = fromIntegralChecked n
+        V.forM_ rs $ \r -> Run.removeReferenceN r n'
+        r <- Merge.complete m
+        Merge.removeReferenceN m n'
+        pure $! CompletedMerge r
+    writeMutVar mergeKnownCompletedVar MergeKnownCompleted
```
mheinzel, line 1253: A weird edge case that can happen: If an async exception occurs before writing `MergeKnownCompleted`, next time around, we will not be able to skip the work in `supplyMergeCredits`. Then `stepMerge` returns False, so we don't enter `completeMerge` and we will forever stay at `MergeMaybeCompleted` and keep calling `stepMerge` on a complete merge.

No real problem, as it's unlikely and the only consequence is some performance overhead, but another reminder of how tricky these things are. We should probably change `stepMerge` to return True on `CompletedMerge` and add a comment about what the returned `Bool` really means (not "it just completed", but "it is complete"). Maybe even return `StepResult` instead, which has more meaning than `Bool`.

Also, if we write to `mergeKnownCompletedVar` in `supplyMergeCredits` instead of here, its usage as a cache is localised to that function and more isolated from the rest of the code. Makes it easier to understand what it does.

src/Database/LSMTree/Internal/Snapshot.hs
mheinzel: I think I'd rather keep the snapshot format lean. This would just mean two more lines of code in opening a snapshot and even save one when creating it I think, or are you referring to some other work?

mheinzel (approved):
I think some previous comments are still unresolved, but no blockers.

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -888,70 +1026,244 @@ scaleCreditsForMerge SingleRun{} _ = ScaledCredits 0
...
+
+{-# SPECIALISE expectCompletedMerge :: MergingRun IO h -> IO (Run IO h) #-}
+expectCompletedMerge ::
+     (MonadMVar m, MonadSTM m, MonadST m, MonadMask m, MonadFix m)
+  => MergingRun m h -> m (Run m h)
+expectCompletedMerge (SingleRun r) = pure r
+expectCompletedMerge (MergingRun _ _ numInputEntries _ totalStepsVar@(TotalStepsVar var) mergeKnownCompletedVar mergeVar) = do
+    mergeKnownCompleted <- readMutVar mergeKnownCompletedVar
+    -- The merge is not guaranteed to be complete.
+    when (mergeKnownCompleted == MergeMaybeCompleted) $ do
```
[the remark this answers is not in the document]
jorisdral: I'd argue that it decreases complexity: the cached value has a simple correctness argument, so checking it here means we can skip the more complex bit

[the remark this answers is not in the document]
mheinzel: Not sure, we need to have the branch that finishes an incomplete merge either way. So this just adds an additional branch. But I don't mind much, all good.

PR #437, opened 2024-10-24
Cleanups related to write buffer updates and value resolution in the core library.

jorisdral:
I had realised some of this core library code was unused, in which case I removed the code. In other cases, where the code was only used in tests/benchmarks, I inlined the definitions and did some refactoring.


dcoutts (approved):
LGTM!

PR #439, opened 2024-10-24
Simplify pre-commit scripts.

wenkokke:
Caveats may apply.

Notably, this removes support for the path/command flags and requires that the scripts are run from the repository root.

I'm happy to restore support for the path/command flags and permit the commands to work from any directory, but figured I shouldn't complicate these files if that was unused complexity.


jorisdral (approved):
LGTM! And it's also nice that it now prints the files that are checked/formatted

jorisdral:
FYI, to make CI pass these lines have to be modified to call the renamed formatting scripts:
https://github.com/IntersectMBO/lsm-tree/blob/c361efb779c0dc3c8998f7c0106462353fb012f3/.github/workflows/haskell.yml#L274
https://github.com/IntersectMBO/lsm-tree/blob/c361efb779c0dc3c8998f7c0106462353fb012f3/.github/workflows/haskell.yml#L212

mheinzel (approved):
Nice!

wenkokke:
I've back ported some added robustness from a different project that checks for unstaged files and checks for the formatter versions.

jorisdral (approved):
LGTM again. Let's squash the commits a bit before we merge

mheinzel:
Seems like something went wrong when rebasing, GitHub thinks you're adding 39 new commits.

PR #441, opened 2024-10-27
Add tests for growing vectors

### LINKED ISSUE #420: Add tests for growing vectors
#419 introduces growing vectors. However, it does not include tests for them. We shall add these tests for growing vectors.

jeltsch:
This resolves #420.


test/Test/Database/LSMTree/Internal/Vector/Growing.hs
```diff
@@ -0,0 +1,212 @@
...
+
+-- * Properties to test
+
+prop_finalVectorIsCorrect :: Positive (Small Int)
+                          -> Segments Integer
+                          -> Property
+prop_finalVectorIsCorrect (Positive (Small initialBufferSize))
+                          (Segments segments)
+    = withFinalVector initialBufferSize segments $ \ vec ->
+      toList vec === concatMap segmentToList segments
```
jorisdral, line 106: Or alternatively, compare on vectors. That is, we convert the segments to vector and concat them into a final vector, and then do the `===` check

test/Test/Database/LSMTree/Internal/Vector/Growing.hs
```diff
@@ -0,0 +1,212 @@
...
+import           Test.Tasty.QuickCheck (testProperty)
+
+-- * Tests
+
+tests :: TestTree
+tests = testGroup "Test.Database.LSMTree.Internal.Vector.Growing" $
+        [
+            testProperty "Final vector is correct"
+                         prop_finalVectorIsCorrect
+        ]
```
jorisdral, line 25: Tests tests are not being run yet. You should add include them in the `Main` module

test/Test/Database/LSMTree/Internal/Vector/Growing.hs
```diff
@@ -0,0 +1,212 @@
...
+    mapM_ (appendSegment vec) segments
+    freeze vec
+
+{-
+    Supplies the final contents of a growing vector for constructing a property.
+
+    The resulting property additionally provides statistics about the buffer
+    size scaling exponents, where a buffer size scaling exponent is the binary
+    logarithm of the ratio between the buffer size after and before appending a
+    segment.
```
jorisdral, line 62: I think this could be elaborated a bit. What are these exponent numbers supposed to tell me?

test/Test/Database/LSMTree/Internal/Vector/Growing.hs
```diff
@@ -0,0 +1,212 @@
...
+      [Replicate count  val' | val'   <- shrink val]
+    where
+
+    shrinkNat :: Int -> [Int]
+    shrinkNat = shrinkMap getNonNegative NonNegative
+
+{-
+    A list of segments to be appended to an initially empty vector. 'Segments a'
+    is isomorphic to '[Segment a]' but has a special way of generating test
+    cases, which avoids skewing with respect to buffer size scaling exponents.
```
jorisdral, line 125: You could use `coverTable` or something similar to check that we are indeed growing vectors in the tests. It will also make the tests more future proof, because if anything about the test (distribution) changes, then the coverage might change, and then `coverTable` could catch that

jorisdral (changes requested):

test/Test/Database/LSMTree/Internal/Vector/Growing.hs
jeltsch: What would be the advantage of comparing vectors? The present solution should potentially use less memory at least, since only one vector is constructed and the lists can be processed lazily.

test/Test/Database/LSMTree/Internal/Vector/Growing.hs
jeltsch: Oops, right. 🙂

jeltsch: Fixed in 6f88b3f710bd5ae0b3996d8341435b74f9f7c37d.

test/Test/Database/LSMTree/Internal/Vector/Growing.hs
jeltsch: Hmm, I had hoped that this would be somehow clear in conjunction with the documentation of the `GrowingVector` type. Buffer size scaling exponents tell you if the buffer is grown and, if yes, how much. If you think of growth happening always by doubling the buffer size, a buffer size scaling exponent says how many growth cycles are needed to accommodate the increased vector length. In practice, though, we don’t perform multiple growth cycles for a single append but grow directly to the final buffer size.

What additional explanation would be helpful? Would a reference to the description of the growing approach in the documentation of `GrowingVector` be sufficient perhaps?

test/Test/Database/LSMTree/Internal/Vector/Growing.hs
jeltsch: Well, the tabulation introduced by `withFinalVector` should already achieve this: no growth results in a buffer size scaling exponent of zero, and growth results in a buffer size scaling exponent greater than zero (and in the latter case the precise value even tells you how much the buffer is grown). Am I missing something?

test/Test/Database/LSMTree/Internal/Vector/Growing.hs
jorisdral: Just this intuition

> Buffer size scaling exponents tell you if the buffer is grown and, if yes, how much. If you think of growth happening always by doubling the buffer size, a buffer size scaling exponent says how many growth cycles are needed to accommodate the increased vector length.

would be nice to include.

test/Test/Database/LSMTree/Internal/Vector/Growing.hs
jorisdral: The point of `coverTable` is that you specify how often you expect each label to appear, and fail the test if the probability of a certain label is too small. So it's not just printing the labels so that we can look at them and see indeed that buffer size scaling exponents are non-zero, it's also asserting something about their distribution

jeltsch: I’ve made a corresponding change in a83d9bbbfc0e44175774a27ca49afe4aade996cc.

jorisdral (approved):

PR #442, opened 2024-10-28
Define snapshot versioning and metadata

jorisdral:
If the snapshot versioning scheme is to your liking, then we should probably also make it part of the PR checklist to update version numbers in case there are changes to the snapshot format.

### DIFF
--- a/src/Database/LSMTree/Internal/Snapshot.hs
+++ b/src/Database/LSMTree/Internal/Snapshot.hs
@@ -2,8 +2,9 @@
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 module Database.LSMTree.Internal.Snapshot (
+    SnapshotLabel (..)
     -- * Snapshot format
-    numSnapRuns
+  , numSnapRuns
   , SnapLevels
   , SnapLevel (..)
   , SnapMergingRun (..)
@@ -23,6 +24,8 @@ import           Control.Monad.Primitive (PrimMonad)
 import           Control.TempRegistry
 import           Data.Foldable (forM_)
 import           Data.Primitive.PrimVar
+import           Data.String
+import           Data.Text (Text)
 import qualified Data.Vector as V
 import           Database.LSMTree.Internal.Config
 import           Database.LSMTree.Internal.Entry
@@ -38,6 +41,11 @@ import           Database.LSMTree.Internal.UniqCounter (UniqCounter,
                      incrUniqCounter, uniqueToRunNumber)
 import           System.FS.API (HasFS)
 import           System.FS.BlockIO.API (HasBlockIO)
+
+-- | Custom text to include in a snapshot file
+newtype SnapshotLabel = SnapshotLabel Text
+  deriving stock (Show, Eq, Read)
+  deriving newtype (Semigroup, IsString)
 
 {-------------------------------------------------------------------------------
   Levels snapshot format
--- a/src/Database/LSMTree/Internal/Snapshot.hs
+++ b/src/Database/LSMTree/Internal/Snapshot.hs
@@ -2,7 +2,16 @@
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 module Database.LSMTree.Internal.Snapshot (
-    SnapshotLabel (..)
+    -- * Versioning
+    SnapshotVersion (..)
+  , major
+  , minor
+  , fromMajorMinor
+  , prettySnapshotVersion
+  , currentSnapshotVersion
+    -- * Snapshot metadata
+  , SnapshotMetaData (..)
+  , SnapshotLabel (..)
     -- * Snapshot format
   , numSnapRuns
   , SnapLevels
@@ -24,7 +33,6 @@ import           Control.Monad.Primitive (PrimMonad)
 import           Control.TempRegistry
 import           Data.Foldable (forM_)
 import           Data.Primitive.PrimVar
-import           Data.String
 import           Data.Text (Text)
 import qualified Data.Vector as V
 import           Database.LSMTree.Internal.Config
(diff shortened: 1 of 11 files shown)


src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -34,6 +44,82 @@ import           Database.LSMTree.Internal.RunNumber
...
+-- A version @x.y@ has two components: a major version number @x@ and a minor
+-- version number @y@. @x.y@ is guaranteed to be backwards compatible with
+-- @x'.y'@ as long as @x == x'@ and @y' <= y@. If @x > x'@, then backwards
+-- compatibility *may* be provided, but is not guaranteed.
+--
+-- The version number determines the format of the snapshot metadata file, but
+-- it also doubles as versioning information for the entire snapshot itself. For
+-- example, if a breaking change is made to the merge scheduling algorithm that
+-- would lead to errors when loading an older snapshot, then the major version
+-- should be increased as well.
```
mheinzel, line 73: So if I understand correctly:
- the major version is necessary because we want to be able to make breaking changes
- the minor version is useful because we can make changes that are backwards compatible, but not forward compatible, plus it might help with debugging things to know exactly which version something is.

Is that correct?

src/Database/LSMTree/Normal.hs
```diff
@@ -1,3 +1,5 @@
+{-# LANGUAGE OverloadedStrings #-}
```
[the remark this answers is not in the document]
jorisdral: What aspect doesn't appeal to you? I don't mind changing this bit of code, but I'm not sure what your preferred alternative would be

src/Database/LSMTree/Internal/Snapshot.hs
jorisdral: Yup, that's about it. Should I clarify more in the comment?

mheinzel: That would be great.

src/Database/LSMTree/Normal.hs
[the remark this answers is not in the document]
mheinzel: I mulled this over for a while now.

A snapshot label is a static thing for each table (they could basically be a type level `Symbol`) and their structure is very clearly defined, so being able to manipulate them freely seems like a mis-feature.
It's a bit contrived, but for example it makes it easier to modify a label in one place (e.g. creating the snapshot), but not another (e.g. opening the snapshot), leading to a mismatch.

I think both use cases seen in this PR can be addressed differently:

Firstly, this uses the `IsString` instance when giving instances of `Labellable` to create a `SnapshotLabel`. But I don't think the newtype provides much value here. The method could simply be `makeSnapshotLabel :: Proxy a -> String/Text` and then internally we wrap that, possibly using a utility wrapper function that produces a `SnapshotLabel` (similar to `Class.serialiseValue` returning `RawBytes`, not `SerialisedValue`).

Secondly, we want to ensure normal snapshots are not open as monoidal tables and vice versa, so we use the `Semigroup` instance internally to append `(normal)`. That feels a little hacky. Why not add a `Normal | Monoidal` field to `SnapshotMetaData`? Then we could also give a more descriptive error message when we encounter a mismatch.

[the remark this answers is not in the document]
jorisdral: > I mulled this over for a while now.
> 
> A snapshot label is a static thing for each table (they could basically be a type level `Symbol`) and their structure is very clearly defined, so being able to manipulate them freely seems like a mis-feature. 

Sure, labels are be static in the sense that they "type" a snapshot, but the structure is not at all clearly defined to me. How do you encode the type information?

We were actually thinking of removing the `Labellable` class altogether, and letting the user themselves provide a `SnapshotLabel` explicitly. The user can then decide what they want to put in the label to "type" their snapshots

> It's a bit contrived, but for example it makes it easier to modify a label in one place (e.g. creating the snapshot), but not another (e.g. opening the snapshot), leading to a mismatch.

We only create labels in two places, so the lockstep tests are always going to catch that mismatch

> I think both use cases seen in this PR can be addressed differently:
> 
> Firstly, this uses the `IsString` instance when giving instances of `Labellable` to create a `SnapshotLabel`. But I don't think the newtype provides much value here. The method could simply be `makeSnapshotLabel :: Proxy a -> String/Text` and then internally we wrap that, possibly using a utility wrapper function that produces a `SnapshotLabel` (similar to `Class.serialiseValue` returning `RawBytes`, not `SerialisedValue`).

I agree that maybe the newtype is currently not too helpful, but 'm not sure the `IsString` instance is a problem -- without `OverloadedStrings`, you'd use `Text.pack` instead of string literals. Both are valid ways to construct `Text`

> Secondly, we want to ensure normal snapshots are not open as monoidal tables and vice versa, so we use the `Semigroup` instance internally to append `(normal)`. That feels a little hacky. Why not add a `Normal | Monoidal` field to `SnapshotMetaData`? Then we could also give a more descriptive error message when we encounter a mismatch.

True it's a bit hacky, but that could go away once `SnapshotLabel`s are passed in explicitly by the user

I'm still not convinced that there is a need to hard-code this distinction between normal and monoidal tables: a normal table is "just" a special case of a monoidal table. The fact that our code relies nowhere on the distinction makes me think that it is purely a surface distinction for the user.

Overall, I think for now at least it is okay to have the added flexibility of text while losing some of the benefits of static checks.

src/Database/LSMTree/Internal/Snapshot.hs
jorisdral: Updated. Let me know what you think

src/Database/LSMTree/Normal.hs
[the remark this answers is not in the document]
mheinzel: > I'm not sure the IsString instance is a problem

I don't think it's a problem. I don't see a need for it either, and it potentially adds some complexity for users (having to understand that the required `SnapshotLabel` can be produced by enabling a language extension and using a type class, vs. just passing a string), but that can be addressed by documentation. So I'm okay with keeping it.

> I'm still not convinced that there is a need to hard-code this distinction between normal and monoidal tables: a normal table is "just" a special case of a monoidal table. The fact that our code relies nowhere on the distinction makes me think that it is purely a surface distinction for the user.

I feel similarly, but we currently have the divide in the API, so the question is, do we want users to be able to snapshot a normal table and open it as a monoidal one? If yes, no need to encode it at all. If not, we should prevent it properly.

it feels wrong to do it via stuffing `" (normal)"` into the user-provided label. Why stringly typed tags instead of a proper type that's part of the metadata? And currently, if there is an error message because the user mixed up their tags, it will leak that internal glued-on part of the label, which is confusing.

Either way, I don't see the advantage of providing a `Semigroup` API. It's just a (unnecessarily) larger API surface.

src/Database/LSMTree/Internal/Snapshot.hs
mheinzel: :+1:

src/Database/LSMTree/Normal.hs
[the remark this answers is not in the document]
jorisdral: Changed the code to not expose `SnapshotLabel`, and I've added a table type field to the snapshot metadata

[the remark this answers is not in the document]
jorisdral: And I've started using the new `SnapshotTableType` field in https://github.com/IntersectMBO/lsm-tree/pull/443

src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -39,6 +42,11 @@ import           Database.LSMTree.Internal.UniqCounter (UniqCounter,
 import           System.FS.API (HasFS)
 import           System.FS.BlockIO.API (HasBlockIO)
 
+-- | Custom text to include in a snapshot file
+newtype SnapshotLabel = SnapshotLabel Text
```
dcoutts, line 128: A newtype is good but do we need `Text` here? This is the only thing that pulls in the extra `text` package dep.

dcoutts (commented):
LGTM

src/Database/LSMTree/Internal/Snapshot.hs
dcoutts: Never mind, adding the dep on `cborg` in #443 adds `text` anyway.

PR #446, opened 2024-10-31
Measure latency for batches of updates and lookups in the WP8 benchmark

jorisdral:
We add a new `measure-batch-latency` flag that, if enabled, produces a file with datapoints about the that can be rendered, for example using `gnuplot`. In sequential mode, the datapoints consist of latency for lookups, and latency for updates separately. In pipelined mode, there are two additional datapoints for the sync latency between threads.

When the flag is disabled, all functions related to latency measurements default to no-ops. These functions are always inlined, and should be optimised away nicely.


mheinzel (approved):
I think the overhead for creating a `Handle` would have been fine, but works either way.

jorisdral:
> I think the overhead for creating a `Handle` would have been fine, but works either way.

Me too, but now it won't create an empty file in your directory, which is nicer

PR #448, opened 2024-11-01
Rename `TableHandle` to `Table`

### LINKED ISSUE #369: Rename `TableHandle` to `Table`
It's more concise, and I don't think the `Handle` part of the name is too important. We have other types that are handle-like, without reflecting that explicitly in their name.

jorisdral:
Resolves #369


test/Test/Database/LSMTree/Internal/Monkey.hs
```diff
@@ -274,7 +274,7 @@ mkBloomST requestedFPR xs = runST $ do
 -- * The FPR (and bits/hash functions) calculations are approximations.
 -- * Rounding errors in the Haskell implementation of FPR calculations
 -- * The Monkey tuning is incompatible with @bloomfilter@'s /next power of 2/
---   rounding of th ebits.
+--   rounding of t ebits.
```
mheinzel, line 277:
```suggestion
--   rounding of the bits.
```

mheinzel (approved):
LGTM

PR #452, opened 2024-11-06
Rename table merge to table union in the public API

jorisdral:
"Table union" seems to be more appropriate than "table merge", since "merge" is often used in the LSM-Tree literature to mean "run merges". The table union semantics are also similar to `Map.union` from the well known `Data.Map` module, so just from the name it should be more obvious at a glance what the semantics of "table union" are.

The semantics of table union are made a little more specific in the haddocks.


src/Database/LSMTree/Monoidal.hs
```diff
@@ -652,30 +652,34 @@ duplicate :: forall m k v.
...
      ResolveValue v
   => Table IO k v
   -> Table IO k v
   -> IO (Table IO k v) #-}
--- | Merge full tables, creating a new table.
+-- | Union two full tables, creating a new table.
+--
+-- Unions are left-biased, meaning inserts and deletes in the first table take
+-- precedence over inserts and deletes in the second table. Monoidal upserts
+-- will be resolved monoidally, but also from left to right.
```
mheinzel, line 667: It's good to add more explanation here, but this one doesn't fully convey the semantics. Deletes from the first table do not affect the second table, so it's misleading to say they take precedence. And if a key is present in both tables, the value is always resolved monoidally, even it just plain inserts have been used. Only if also two blobs were present, the first one take precedence (but that can't happen in monoidal tables).

Generally, tables really behave like maps here. It doesn't make a difference whether a value was never inserted or deleted later. Similarly, whether it was inserted as a single value or using multiple mupserts. These details don't leak out.

So it's probably easiest to explain in terms of `Data.Map.unionWith (<>)`.

mheinzel (commented):
The renaming makes sense, but the explanation of the semantics is off.

src/Database/LSMTree/Monoidal.hs
jorisdral: > It's good to add more explanation here, but this one doesn't fully convey the semantics. Deletes from the first table do not affect the second table, so it's misleading to say they take precedence. 

You mean that a delete on the left table and insert on the right table might resolve to an insert? If that is the case, should we forget about deletes at all?

> And if a key is present in both tables, the value is always resolved monoidally, even if just plain inserts have been used.  Only if also two blobs were present, the first one take precedence (but that can't happen in monoidal tables).

And here do you mean, for example, that a mupsert on the left would be monoidally combined with an insert on the right? Indeed, I realise my explanation is here incorrect. Though, if it's two plain inserts, the left value does take precedence, doesn't it? It's a `First` monoid in that scenario -- isn't that the same as a left-biased union?

> Generally, tables really behave like maps here. It doesn't make a difference whether a value was never inserted or deleted later. Similarly, whether it was inserted as a single value or using multiple mupserts. These details don't leak out. So it's probably easiest to explain in terms of `Data.Map.unionWith (<>)`.

I think I had misremembered the union semantics. Makes sense to explain it in terms of using `Data.Map.unionWith`, but I think that could also be slightly tricky still. Should we use a mental model of `Map k v` or `Map k (Entry v)` or `Map k (InsertOrMuspert v)`? My understanding right now is that we should use the last one as our mental model. Is that correct?

mheinzel: > You mean that a delete on the left table and insert on the right table might resolve to an insert? If that is the case, should we forget about deletes at all?

Exactly. And we do forget about Deletes once they are on the last level (what we call a last level merge). If they could still matter (e.g. through union with another table), we couldn't do that.

> Though, if it's two plain inserts, the left value does take precedence, doesn't it?

Even then, they are merged.

> Makes sense to explain it in terms of using Data.Map.unionWith, but I think that could also be slightly tricky still. Should we use a mental model of Map k v or Map k (Entry v) or Map k (InsertOrMuspert v)?

Just `Map k v`. Look at the implementation in `Model.Table`. It is literally a key-value map with `Map.unionWith resolveSerialised` (with a little noise because of handling blobs).

jorisdral: Okay, updated the documentation. Turns out I was just confusing myself

mheinzel (approved):

PR #453, opened 2024-11-06
Restore uniqueness of GHA cache names per configuration

jorisdral:
And make sure to hash all configuration files (like `*.cabal*` and `cabal.project*`) correctly, and include those hahses in the GHA cache names too. We no longer hash put `**` in front of the globs, because that matches on files in `dist-newstyle` as well.

This is a follow-up to #450, which broke uniqueness 

Note that changes to *any* `cabal.project*` file will lead to a changed hash name, but I'd prefer that over the more involved approach of determining exactly which config files are in use by a job, and only hash those. It's also not really a problem if the hash name changes: it just means we store a new cache and the old one at some point gets evicted when it hasn't been used in a while.

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -93,10 +93,10 @@ jobs:
         cache-name: cache-cabal-build-130824
       with:
         path: ${{ steps.setup-haskell.outputs.cabal-store }}
-        key: ${{ runner.os }}-${{ matrix.ghc }}-${{ env.cache-name }}-${{ hashFiles('**/*.cabal') }}-${{ hashFiles('**/cabal.project') }}-${{ hashFiles('**/cabal.project.debug') }}
+        key: ${{ runner.os }}-${{ matrix.ghc }}-${{ env.cache-name }}-${{ hashFiles('*.cabal') }}-${{ hashFiles('cabal.project*') }}-${{ matrix.cabal-project-file }}
         restore-keys: |
-          ${{ runner.os }}-${{ matrix.ghc }}-${{ env.cache-name }}-${{ hashFiles('**/*.cabal') }}-${{ hashFiles('**/cabal.project') }}-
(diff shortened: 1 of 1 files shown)


.github/workflows/haskell.yml
```diff
@@ -93,10 +93,10 @@ jobs:
         cache-name: cache-cabal-build-130824
       with:
         path: ${{ steps.setup-haskell.outputs.cabal-store }}
-        key: ${{ runner.os }}-${{ matrix.ghc }}-${{ env.cache-name }}-${{ hashFiles('**/*.cabal') }}-${{ hashFiles('**/cabal.project') }}-${{ hashFiles('**/cabal.project.debug') }}
+        key: ${{ runner.os }}-${{ matrix.ghc }}-${{ env.cache-name }}-${{ hashFiles('*.cabal') }}-${{ hashFiles('cabal.project*') }}-${{ matrix.cabal-project-file }}
```
mheinzel, line 96: So this includes the `cabal.project` that is just a symlink? Does it matter, or do we consistently use `matrix.cabal-project-file` to pick the one to use?

jorisdral: [`hashFiles`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#hashfiles) requires a file pattern that allows globs. Presumably the glob results in the same files as listed here:

```
❯ echo cabal.project*
cabal.project cabal.project.blockio-uring cabal.project.debug cabal.project.local cabal.project.release
```

Just to err on the safe side, I thought it would be good to hash all configuration files, i.e., `*.cabal` and `cabal.project.*`. Omitting or including `cabal.project` in this fine would have both been fine, because we use `matrix.cabal-project-file` consistely to pick the actual configuration for builds/tests

mheinzel: Okay, then it seems good. Was just wondering whether `cabal.project.*` would be better.

mheinzel (approved):
I am not too familiar with our CI setup, but this looks good to me.

PR #455, opened 2024-11-07
Separate utilities related to pages

jeltsch:
This pull request moves `PageNo`, `NumPages`, and `PageSpan` along with the operations associated with them out of `Database.LSMTree.Internal.IndexCompact` and into a new, dedicated module. The reason for this change is that those things are not specific to the compact index, not even to indexes at all. This pull request is not least part of the preparations for making index types selectable.

### DIFF
--- a/src/Database/LSMTree/Internal/Page.hs
+++ b/src/Database/LSMTree/Internal/Page.hs
@@ -1,0 +1,64 @@
+-- | Utilities related to pages.
+module Database.LSMTree.Internal.Page (
+    PageNo (..)
+  , nextPageNo
+  , NumPages (..)
+  , getNumPages
+  , PageSpan (..)
+  , singlePage
+  , multiPage
+  , pageSpanSize
+  ) where
+
+import           Control.DeepSeq (NFData (..))
+
+-- | A 0-based number identifying a disk page.
+newtype PageNo = PageNo { unPageNo :: Int }
+  deriving stock (Show, Eq, Ord)
+  deriving newtype NFData
+
+-- | Increment the page number.
+--
+-- Note: This does not ensure that the incremented page number exists within a given page span.
+{-# INLINE nextPageNo #-}
+nextPageNo :: PageNo -> PageNo
+nextPageNo = PageNo . succ . unPageNo
+
+-- | The number of pages contained by an index or other paging data-structure.
+--
+-- Note: This is a 0-based number; take care to ensure arithmetic underflow
+-- does not occur during subtraction operations!
+newtype NumPages = NumPages Word
+  deriving stock (Eq, Ord, Show)
+  deriving newtype (NFData)
+
+-- | A type-safe "unwrapper" for 'NumPages'. Use this accessor whenever you want
+-- to convert 'NumPages' to a more versatile number type.
+{-# INLINE getNumPages #-}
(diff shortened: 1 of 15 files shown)


src/Database/LSMTree/Internal/Page.hs
```diff
@@ -0,0 +1,63 @@
...
+import           Control.DeepSeq (NFData (..))
+
+-- | A 0-based number identifying a disk page.
+newtype PageNo = PageNo { unPageNo :: Int }
+  deriving stock (Show, Eq, Ord)
+  deriving newtype NFData
+
+-- | Increment the page number.
+--
+-- Note: This does not encure that the incremented page number exists within a given page span.
```
jorisdral, line 22:
```suggestion
-- Note: This does not ensure that the incremented page number exists within a given page span.
```

src/Database/LSMTree/Internal/Page.hs
```diff
@@ -0,0 +1,63 @@
...
+
+{-# INLINE singlePage #-}
+singlePage :: PageNo -> PageSpan
+singlePage i = PageSpan i i
+
+{-# INLINE multiPage #-}
+multiPage :: PageNo -> PageNo -> PageSpan
+multiPage i j = PageSpan i j
+
+pageSpanSize :: PageSpan -> NumPages
```
jorisdral, line 61:
```suggestion
{-# INLINE pageSpanSize #-}
pageSpanSize :: PageSpan -> NumPages
```

jorisdral: It's probably inlined even without the pragma, but let's add it anyway

jorisdral (approved):
LGTM!

> ... The reason for this change is that those things are not specific to the compact index, **not even to indexes** at all. ...

I'd argue that the second argument disputable and that it depends on your point of view, but that's beside the point

jeltsch:
> > The reason for this change is that those things are not specific to the compact index, not even to indexes at all.
>
> I’d argue that the second argument is disputable and that it depends on your point of view, but that’s beside the point.

Well, the things that are now in this new, dedicated module are only used in conjunction with indexes, from what I know; so in a way you’re completely right. However, semantically those things are not specific to index at all, as they are only about referencing and counting pages and pages not only show up in conjunction with indexes. Actually, the close relationship between those page-related things and indexes seems to be partly an implementation detail to me, considering, for example, the arrangement of storing the total number of pages along with the index.

src/Database/LSMTree/Internal/Page.hs
jeltsch: That’s funny: Everything after the import list in this new module was taken verbatim from the compact-index module; so also the “encure” had been there before. I think at some point I saw this “encure” and just thought that it was some high-level word, which I just didn’t happen to know. 😅 I just had a web search performed for “encure” and found out that apparently “Encure” is just the name of some band. 😁

jeltsch: Fixed in 8e5793e5ee0a8fb322b821722ea9493b938b5bae.

src/Database/LSMTree/Internal/Page.hs
jeltsch: Done in 10dcc12bb7e4c17093cab3dc4eac766b259f7839.

PR #456, opened 2024-11-08
Create a proper directory for each snapshot 

jorisdral:
And put the snapshot metadata file in there. In addition, compute the checksum for the metadata file and put it in a separate file in the same directory.

### DIFF
--- a/src/Database/LSMTree/Internal.hs
+++ b/src/Database/LSMTree/Internal.hs
@@ -72,7 +72,7 @@ import           Control.Concurrent.Class.MonadSTM (MonadSTM (..))
 import           Control.Concurrent.Class.MonadSTM.RWVar (RWVar)
 import qualified Control.Concurrent.Class.MonadSTM.RWVar as RW
 import           Control.DeepSeq
-import           Control.Monad (unless, void, when)
+import           Control.Monad (unless)
 import           Control.Monad.Class.MonadST (MonadST (..))
 import           Control.Monad.Class.MonadThrow
 import           Control.Monad.Fix (MonadFix)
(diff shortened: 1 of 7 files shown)


wenkokke:
LGTM

src/Database/LSMTree/Internal.hs
```diff
@@ -1105,9 +1105,17 @@ snapshot resolve snap label tableType t = do
...
-              when b $ throwIO (ErrSnapshotExists snap)
+
+      -- Guard that the snapshot does not exist already
+      let snapDir = Paths.namedSnapshotDir (tableSessionRoot thEnv) snap
+      doesSnapshotExist <-
+        FS.doesDirectoryExist (tableHasFS thEnv) (Paths.getNamedSnapshotDir snapDir)
+      if doesSnapshotExist then
+        throwIO (ErrSnapshotExists snap)
+      else
+        -- we assume the snapshots directory already exists, so we just have to
```
wenkokke, line 1116: Where is the snapshots directory created? Should we handle the error when it does not exist?

src/Database/LSMTree/Internal.hs
```diff
@@ -1105,9 +1105,17 @@ snapshot resolve snap label tableType t = do
...
-      FS.doesFileExist (tableHasFS thEnv) snapPath >>= \b ->
-              when b $ throwIO (ErrSnapshotExists snap)
+
+      -- Guard that the snapshot does not exist already
+      let snapDir = Paths.namedSnapshotDir (tableSessionRoot thEnv) snap
+      doesSnapshotExist <-
+        FS.doesDirectoryExist (tableHasFS thEnv) (Paths.getNamedSnapshotDir snapDir)
+      if doesSnapshotExist then
+        throwIO (ErrSnapshotExists snap)
+      else
```
wenkokke, line 1115: Should we extract this into a helper function? It feels like it's repeated quite a bit throughout.

jorisdral: Hmm, I tried to provide helper functions but it's not simplifying the code much, so I'm keeping it as is

wenkokke: Seems fine. It is quite simple.

wenkokke (approved):

PR #460, opened 2024-11-11
Make the different index types use a common chunk type

jeltsch:
This pull request makes the compact index use the `Chunk` type defined in `Database.LSMTree.Internal.Chunk`, which the ordinary index is already using. This paves the way to a common interface for both the ordinary and the compact index.

### DIFF
--- a/src/Database/LSMTree/Internal/Chunk.hs
+++ b/src/Database/LSMTree/Internal/Chunk.hs
@@ -5,7 +5,8 @@ module Database.LSMTree.Internal.Chunk
 (
     -- * Chunks
     Chunk (Chunk),
-    fromChunk,
+    toByteVector,
+    toByteString,
 
     -- * Balers
     Baler,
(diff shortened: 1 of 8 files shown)


src/Database/LSMTree/Internal/Chunk.hs
```diff
@@ -19,23 +20,31 @@ import           Prelude hiding (length)
...
-newtype Chunk = Chunk (Vector Word8)
+newtype Chunk = Chunk (Vector Word8) deriving stock (Eq, Show)
 
-fromChunk :: Chunk -> Vector Word8
-fromChunk (Chunk content) = content
+-- | Yields the contents of a chunk as a byte vector.
+toByteVector :: Chunk -> Vector Word8
+toByteVector (Chunk byteVector) = byteVector
+
+-- | Yields the contents of a chunk as a byte string.
```
jorisdral, line 44:
```suggestion
-- | Yields the contents of a chunk as a strict byte string.
```

jorisdral (approved):

src/Database/LSMTree/Internal/Chunk.hs
jeltsch: I made a corresponding change in 9b1599ee1eaf72ce29a4bbe7f84d191b7f47fb2f. I put the “strict” in parentheses, since the naming within the `bytestring` package and also the naming of the above function is centered around the idea that strict byte strings are the default.

PR #462, opened 2024-11-11
Adding support for GHC-9.12 based on the alpha release

recursion-ninja:
# Description

I added support for the new GHC-9.12 based on the current alpha release version. The only changes required were the bumping of version bounds and relaxing upper bound constrains on the `base` package in the project's direct and transitive dependencies.

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -22,7 +22,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["8.10.7", "9.2.8", "9.4.8", "9.6.4", "9.8.2", "9.10.1"]
+        # TODO: Change "9.12.0.20241031" --> "9.12.1" after official release
+        ghc: ["8.10.7", "9.2.8", "9.4.8", "9.6.4", "9.8.2", "9.10.1", "9.12.0.20241031"]
         cabal: ["3.10.2.1"]
         os: [ubuntu-latest, windows-latest, macOS-latest]
         cabal-flags: [""]
@@ -67,6 +68,8 @@ jobs:
         ghc-version: ${{ matrix.ghc }}
         cabal-version: ${{ matrix.cabal }}
         cabal-update: true
+        # TODO: remove once 9.12.1 is officially released
+        ghcup-release-channel: https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-prereleases-0.0.8.yaml
 
     - name: Install liburing (on Linux)
       id: setup-liburing
--- a/cabal.project.debug
+++ b/cabal.project.debug
@@ -26,3 +26,4 @@ if impl(ghc >=9.4.6 && <9.5 || >=9.6.3)
 
   package fs-sim
     ghc-options: -fcheck-prim-bounds
+
--- a/test/kmerge-test.hs
+++ b/test/kmerge-test.hs
@@ -29,6 +29,7 @@ import           Test.Tasty.QuickCheck (testProperty, (===))
 import qualified KMerge.Heap as K.Heap
 import qualified KMerge.LoserTree as K.Tree
 
+
 -- tests and benchmarks for various k-way merge implementations.
 -- in short: loser tree is optimal in comparison counts performed,
 -- but mutable heap implementation has lower constant factors.
@@ -57,6 +58,7 @@ main = do
     _ <- evaluate $ force input7
     _ <- evaluate $ force input5
 
+
     defaultMainWithIngredients B.benchIngredients $ testGroup "kmerge"
         [ testGroup "tests"
             [ testGroup "merge"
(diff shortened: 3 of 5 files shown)


jorisdral (dismissed):
If we start supporting `9.12`, then we should also add GHA jobs for that compiler version to our CI. It should be sufficient to add the version to the following line:

https://github.com/IntersectMBO/lsm-tree/blob/2044d0002a1a1a74c928238f5247318351a4fa5a/.github/workflows/haskell.yml#L25

cabal.project.debug
```diff
@@ -25,4 +25,7 @@ if impl(ghc >=9.4.6 && <9.5 || >=9.6.3)
     ghc-options: -fcheck-prim-bounds
 
   package fs-sim
-    ghc-options: -fcheck-prim-bounds
\ No newline at end of file
+    ghc-options: -fcheck-prim-bounds
+
+if impl(ghc >=9.12)
+  allow-newer: base
```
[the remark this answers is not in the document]
recursion-ninja: Removed!

recursion-ninja:
> If we start supporting `9.12`, then we should also add GHA jobs for that compiler version to our CI. 

That seems obvious. Slightly embarrassed I missed this.

recursion-ninja:
*Four* of the `kmerge-test` test cases involving `sortConcat` are failing in CI for GHC 9.12.1:
  1. `kmerge/tests/count/eight/sortConcat`
  2. `kmerge/tests/count/five/sortConcat`
  3. `kmerge/tests/count/levelling-min/sortConcat`
  4. `kmerge/tests/count/levelling-max/sortConcat`

Curiously, these same tests all pass on my MacOS machine. @jorisdral Not sure how to proceed.

jorisdral:
> _Four_ of the `kmerge-test` test cases involving `sortConcat` are failing in CI for GHC 9.12.1:
> 
>     1. `kmerge/tests/count/eight/sortConcat`
> 
>     2. `kmerge/tests/count/five/sortConcat`
> 
>     3. `kmerge/tests/count/levelling-min/sortConcat`
> 
>     4. `kmerge/tests/count/levelling-max/sortConcat`

I've searched the GHC issue tracker a bit and it seems the implementation of `sort` has changed, which could be the reason for the changed count of comparisons. Ref: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/11870. It would be good to verify whether that change is the cause

recursion-ninja:
> It would be good to verify whether that change is the cause.

This is indeed the cause. I copied the `sort` and `sortBy` implementations from `base-4.19.1` into the `kmerge-test.hs` test-suite as `sortOriginal` and `sortOriginalBy`, respectively. Subsequently, I replaced all calls to `L.sort` with the new `sortOriginal` function. When rerunning the test-suite using `ghc-9.12.0`, the test suite ***passes*** when invoking the old sorting routine rather than the new implementation in `base4.21`.

Now the question is how to proceed...

Perhaps, use CPP based on the base version? Would like your insight on this matter.

wenkokke:
I'm not sure what the exact test is doing, but it might be good to relax it to rely less on the implementation details of sort, e.g., use an upper bound rather than an exact number of comparisons, or verify that the result is ordered rather than an exact order.

recursion-ninja:
@wenkokke That is a good idea. I'm not really satisfied with the CPP solution I quickly slapped together, so I'll look into the intent behind the test case.

jorisdral:
> @wenkokke That is a good idea. I'm not really satisfied with the CPP solution I quickly slapped together, so I'll look into the intent behind the test case.

From what I recall, the intent here is to count comparisons for different heap-like implementations so that we can compare the implementations. I don't think we have to preserve the exact count for the `sort` test across GHC versions, as it's mainly a baseline. We can change the expected count based on the GHC version with a CPP pragma -- that seems to be the simplest solution.

.github/workflows/haskell.yml
```diff
@@ -67,6 +68,7 @@ jobs:
         ghc-version: ${{ matrix.ghc }}
         cabal-version: ${{ matrix.cabal }}
         cabal-update: true
+        ghcup-release-channel: https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-prereleases-0.0.8.yaml
```
jorisdral, line 71:
```suggestion
        TODO: remove once 9.12.1 is officially released
        ghcup-release-channel: https://raw.githubusercontent.com/haskell/ghcup-metadata/master/ghcup-prereleases-0.0.8.yaml
```

.github/workflows/haskell.yml
```diff
@@ -67,6 +68,8 @@ jobs:
         ghc-version: ${{ matrix.ghc }}
         cabal-version: ${{ matrix.cabal }}
         cabal-update: true
+        TODO: remove once 9.12.1 is officially released
```
jorisdral, line 71:
```suggestion
        # TODO: remove once 9.12.1 is officially released
```

test/kmerge-test.hs
```diff
@@ -57,6 +60,8 @@ main = do
     _ <- evaluate $ force input7
     _ <- evaluate $ force input5
 
+    traverse_ print input8
```
jorisdral, line 63:
```suggestion
```

jorisdral (approved):

PR #463, opened 2024-11-12
Make MergingRun reference-counted

mheinzel:
Previously, the `Merge` datatype contained the `RefCounter`, but its input runs needed to be managed separately. This PR moves it up in the hierarchy, into the `MergingRun` type, which is also responsible for managing the input runs.

There are some TODOs I want to address in further PRs, once this one is merged. In particular, I want to create a separate module for the ` MergingRun`, as it is one of the LSM's fundamental abstractions. This also includes making other code interactwith the `MergingRun` through a clearly defined interface, as opposed to interacting with the internal state by directly.


src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -271,30 +284,112 @@ type Levels m h = V.Vector (Level m h)
...
+  -> NumRuns
+  -> NumEntries
+  -> MergeKnownCompleted
+  -> MergingRunState m h
+  -> m (MergingRun m h)
+newMergingRun mergePolicy mergeNumRuns mergeNumEntries knownCompleted state = do
+    mergeUnspentCredits <- UnspentCreditsVar <$> newPrimVar 0
+    mergeStepsPerformed <- TotalStepsVar <$> newPrimVar 0
+    case state of
+      OngoingMerge{}   -> assert (knownCompleted == MergeMaybeCompleted) (pure ())
```
jorisdral, line 342: `newMergingRun` is used rarely enough that it can be a proper error instead of an assertion, IMO

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -271,30 +284,112 @@ type Levels m h = V.Vector (Level m h)
...
+mergingRunGetRuns reg mr =
+    -- We take the references while holding the MVar to make sure the MergingRun
+    -- does not get completed concurrently before we are done.
+    withMVar (mergeState mr) $ \case
+      CompletedMerge r -> V.singleton <$> allocRun r
+      OngoingMerge rs _ _ -> rs <$ V.forM_ rs allocRun
+  where
+    allocRun r = do
+        allocateTemp reg (Run.addReference r) (\_ -> Run.removeReference r)
+        pure r
```
jorisdral, line 381: I think that the name is not clear about the fact that it adds references. Maybe we should call it `tempAddReferenceInputRuns` or something along those lines?

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -1014,78 +1054,80 @@ newtype ScaledCredits = ScaledCredits Int
...
+      LevelTiering ->
+        -- A tiering merge has 5 runs at most (one could be held back to merged
+        -- again) and must be completed before the level is full (once 4 more
+        -- runs come in).
+        ScaledCredits (c * (1 + 4))
+      LevelLevelling ->
+        -- A levelling merge has 1 input run and one resident run, which is (up
+        -- to) 4x bigger than the others. It needs to be completed before
+        -- another run comes in.
+        -- TODO: this is currently assuming a naive worst case, where the
```
jorisdral, line 1071: Empty newline

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -1226,40 +1268,41 @@ completeMerge ::
...
     modifyMVarMasked_ mergeVar $ \case
-      mr@CompletedMerge{} -> pure $! mr
+      mrs@CompletedMerge{} -> pure $! mrs
       (OngoingMerge rs _spentCreditsVar m) -> do
-        -- TODO: we'll likely move away from this style of reference counting,
-        -- so this code will change in the future.
-        RefCount n <- Merge.readRefCount m
-        let !n' = fromIntegralChecked n
-        V.forM_ rs $ \r -> Run.removeReferenceN r n'
+        -- first try to complete the merge, in case it fails
```
jorisdral, line 1273:
```suggestion
        -- first try to complete the merge before performing other side effects, in case the completion fails
```

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -1226,40 +1268,41 @@ completeMerge ::
...
-    withMVar mergeVar $ \case
+    r <- withMVar mergeState $ \case
       CompletedMerge r -> pure r
       OngoingMerge{} -> do
-        -- If the algorithm /still/ finds an ongoing merge here, then it is a bug in
+        -- If the algorithm finds an ongoing merge here, then it is a bug in
         -- our merge sceduling algorithm. As such, we throw a pure error.
         error "expectCompletedMerge: expected a completed merge, but found an ongoing merge"
+    -- return a fresh reference to the run
+    -- TODO TempRegistry?
```
jorisdral, line 1300: Empty newline

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -1226,40 +1268,41 @@ completeMerge ::
...
+    Run.addReference r
+    removeReferenceMergingRun mr
```
jorisdral, line 1302: Use `allocTemp` and `freeTemp`

src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -218,12 +218,12 @@ numSnapRuns sl = V.sum $ V.map go1 sl
...
+    snapIncoming     :: !SnapIncomingRun
   , snapResidentRuns :: !(V.Vector RunNumber)
   }
   deriving stock (Show, Eq)
 
-data SnapMergingRun =
+data SnapIncomingRun =
     SnapMergingRun !MergePolicyForLevel !NumRuns !NumEntries !UnspentCredits !MergeKnownCompleted !SnapMergingRunState
   | SnapSingleRun !RunNumber
   deriving stock (Show, Eq)
```
jorisdral, line 229: Is it intentional that `SnapIncomingRun` represents the snapshot format of both `IncomingRun` and `MergingRun`?

jorisdral (changes requested):
Looks good! Just some small suggestions and then we can merge this

prototypes/ScheduledMerges.hs
```diff
@@ -82,25 +82,26 @@ data Level s =
...
 data MergeLastLevel = MergeMidLevel | MergeLastLevel
   deriving stock (Eq, Show)
 
--- | A \"merging run\" is the representation of an ongoing incremental merge,
--- and in mutable. It is also a unit of sharing between duplicated LSM handles.
+-- | We represent single runs specially, rather than putting them in as a
+-- 'CompletedMerge'.
+-- This is for two reasons: to see statically that it's a single run without
+-- having to read the 'STRef', and secondly to make it easier to avoid supplying
+-- merge credits. It's not essential, but simplifies things somewhat.
```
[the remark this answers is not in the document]
mheinzel: No real reason, I think I just separated things into blocks and moved them around without joining them again in the end.

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -184,22 +186,33 @@ data LevelsCache m h = LevelsCache_ {
...
-    , cachedIndexes   = mapStrict Run.runIndex rs
-    , cachedKOpsFiles = mapStrict Run.runKOpsFile rs
-    }
+    -- TODO: this is not terribly performant, but it is also not sure if we are
+    -- going to need this in the end. We might get rid of the LevelsCache.
+    rs <- fmap fold $ V.forM lvls $ \(Level ir rs) -> do
+      incoming <- case ir of
+        Single r   -> V.singleton <$> allocRun r
+        Merging mr -> mergingRunGetRuns reg mr
+      (incoming <>) <$> V.forM rs allocRun
```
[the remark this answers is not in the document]
mheinzel: done

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -271,30 +284,112 @@ type Levels m h = V.Vector (Level m h)
...
+  -> MergeKnownCompleted
+  -> MergingRunState IO h
+  -> IO (MergingRun IO h)
+  #-}
+-- This allows constructing ill-formed MergingRuns, but the flexibility is
+-- needed for creating a merging run that is already Completed, as well as
+-- opening a merging run from a snapshot.
+-- TODO: instead create a Single run when OneShot merging?
+-- TODO: do not store MergeKnownCompleted in snapshot? It's redundant.
+-- TODO: slightly different API for opening from snapshot?
```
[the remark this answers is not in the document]
mheinzel: Hmm, I don't like that it takes up this much space, and I don't really need the TODOs to show up in Haddock in the first place. But I just added newlines for now.

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -271,30 +284,112 @@ type Levels m h = V.Vector (Level m h)
...
+      -- | A variable that caches knowledge about whether the merge has been
+      -- completed. If 'MergeKnownCompleted', then we are sure the merge has been
+      -- completed, otherwise if 'MergeMaybeCompleted' we have to check the
       -- 'MergingRunState'.
-      !(StrictMVar m (MergingRunState m h))
-    -- | The result of merging a single run, is a single run.
-  | SingleRun !(Run m h)
+    , mergeKnownCompleted :: !(MutVar (PrimState m) MergeKnownCompleted)
+    , mergeState          :: !(StrictMVar m (MergingRunState m h))
+    , mergeRefCounter     :: !(RC.RefCounter m)
```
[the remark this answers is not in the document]
mheinzel: I think for `Merge` and (ultimately probably also `MergingRun`) the idea is to use a qualified import. It also contains identifiers like `new`.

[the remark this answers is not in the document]
mheinzel: We can pick a different prefix, but I don't think it makes much of a difference. Also, IMO `mergingRun` is quite long and I don't particularly like abbreviations like `mr`.

src/Database/LSMTree/Internal/MergeSchedule.hs
mheinzel: I hope to remove this code soon either way, as I don't think one should be able to pass in an invalid `MergeKnownCompleted` in the first place.

src/Database/LSMTree/Internal/MergeSchedule.hs
mheinzel: Naming is difficult here. It also doesn't just add references, and it can also return the resulting run (instead of input runs). 

Once it's in a `MergingRun` module and returns `Vector (Ref (Run m h))` I would call it something like `{get,make,create,clone}RunRefs`, depending on terminology used in the `Ref` API. Maybe `mergingRunCloneRunRefs` is already a decent choice now?

src/Database/LSMTree/Internal/MergeSchedule.hs
mheinzel: Ah, maybe I still want to address this TODO in this PR.

src/Database/LSMTree/Internal/Snapshot.hs
mheinzel: Well, `IncomingRun` also represents/includes `MergingRun`. `IncomingRun` and `SnapIncomingRun` are isomorphic. I just don't think adding another layer (factoring out into a `SnapMergingRun` type) is worth it.

mheinzel:
I addressed most feedback, but still want to change the code following https://github.com/IntersectMBO/lsm-tree/pull/463#discussion_r1838173878 and https://github.com/IntersectMBO/lsm-tree/pull/463#discussion_r1838186877, plus more, depending on your opinion on some of my replies.

src/Database/LSMTree/Internal/MergeSchedule.hs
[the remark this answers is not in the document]
jorisdral: The advantage of using something like `mr` is that you can often create unique identifiers without sacrificing much of the identifier length. With qualified imports you might even choose to get rid of prefixes

I don't feel strongly about the names, I just wanted to bring up that `merge` is an overloaded name. It can refer to a `Merge` field, a `MergingRun` field, or a function (from the verb  "to merge")

jorisdral (approved):

PR #464, opened 2024-11-13
Prototype improvements, support mupsert

mheinzel:
Just a few small refactorings and tweaks, partly in preparation of table union support in the prototype. Probably best reviewed by looking at the individual commits.

### DIFF
--- a/prototypes/ScheduledMerges.hs
+++ b/prototypes/ScheduledMerges.hs
@@ -19,7 +19,7 @@
 module ScheduledMerges (
     -- * Main API
     LSM,
-    Key, Value, Blob,
+    Key (K), Value (V), resolveValue, Blob (B),
     new,
     LookupResult (..),
     lookup, lookups,
@@ -122,12 +122,20 @@ runSize = Map.size
 bufferSize :: Buffer -> Int
 bufferSize = Map.size
 
-type Op     = Update Value Blob
+type Op = Update Value Blob
 
-type Key    = Int
-type Value  = Int
-type Blob   = Int
+newtype Key = K Int
+  deriving stock (Eq, Ord, Show)
+  deriving newtype Enum
 
+newtype Value  = V Int
+  deriving stock (Eq, Show)
+
+resolveValue :: Value -> Value -> Value
+resolveValue (V x) (V y) = V (x + y)
+
+newtype Blob = B Int
+  deriving stock (Eq, Show)
 
 -- | The size of the 4 tiering runs at each level are allowed to be:
 -- @4^(level-1) < size <= 4^level@
@@ -425,6 +433,7 @@ new = do
 inserts :: Tracer (ST s) Event -> LSM s -> [(Key, Value)] -> ST s ()
 inserts tr lsm kvs = updates tr lsm [ (k, Insert v Nothing) | (k,v) <- kvs ]
 
+-- TODO: support (and test!) blobs
 insert :: Tracer (ST s) Event -> LSM s -> Key -> Value -> ST s ()
 insert tr lsm k v = update tr lsm k (Insert v Nothing)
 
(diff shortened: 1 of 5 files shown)


prototypes/ScheduledMerges.hs
jorisdral, line 1: It's outside of the scope of this PR and it does not have to be resolved here, but: looking at this module and how it depends on `lsm-tree` but only for the `LookupResult` and `Update` types makes me wonder whether we should decouple the prototypes from the core library

prototypes/ScheduledMergesTestQLS.hs
```diff
@@ -131,14 +140,14 @@ instance InLockstep Model where
   observeModel (MDump   x) = OId x
 
   usedVars  ANew             = []
```
jorisdral, line 142: Alignment

prototypes/ScheduledMergesTestQLS.hs
```diff
@@ -35,17 +35,17 @@ prop_LSM = Lockstep.runActions (Proxy :: Proxy Model)
 
 type ModelLSM = Int
 
-newtype Model = Model { mlsms :: Map ModelLSM (Map Key Value) }
+newtype Model = Model { mlsms :: Map ModelLSM (Map Key (Value, Maybe Blob)) }
   deriving stock (Show)
```
jorisdral, line 39: I'm okay with including blobs, though I'd like to get some clarification here why we're adding blob support. Is it an aspect we need to prototype?

For scheduling, it does not matter whether blobs are supported, because a value + blob is isomorphic to just a value (that includes the blob as part of the value).

For unions, blobs are combined differently than values, so in that sense it may be useful. Though you could also argue that the important aspect of unions to prototype is proper bracketing, and I suspect that does not rely much on whether updates include mupserts and/or blobs.

Does that make sense?

prototypes/ScheduledMergesTest.hs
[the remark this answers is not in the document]
mheinzel: It was only moved.

prototypes/ScheduledMerges.hs
mheinzel: I thought the same and removed the dependency when adding Mupsert support. It even became necessary for `Update`, since it needs to support both blobs and mupserts, similar to the one in `Model.Table`.

prototypes/ScheduledMergesTestQLS.hs
mheinzel: I agree that blobs aren't the most interesting things to model here and I wouldn't have added them from scratch. However, they were already supported in the prototype, but only through `update`, not `insert`. Also, they weren't tested (and not part of the QLS model). So I just wanted to make it more consistent.

prototypes/ScheduledMerges.hs
```diff
@@ -468,21 +488,32 @@ supply (LSMHandle scr lsmr) credits = do
...
+data LookupResult v b =
+    NotFound
+  | Found !v !(Maybe b)
+  deriving stock (Eq, Show)
+
 lookups :: LSM s -> [Key] -> ST s [(Key, LookupResult Value Blob)]
-lookups lsm = mapM (\k -> (k,) <$> lookup lsm k)
+lookups lsm ks = do
+    runs <- concat <$> allLayers lsm
+    return $ map (\k -> (k, doLookup k runs)) ks
```
jorisdral, line 499: Isn't the code before equivalent to the code after?

jorisdral (approved):

prototypes/ScheduledMerges.hs
mheinzel: Yes, I just wanted to extract a pure function in preparation for lookups in table unions.

PR #466, opened 2024-11-13
Proper snapshot implementation

jorisdral:
This supersedes #457. The aim of this PR is to switch to properly implementing snapshots using hard links, although we're using a placeholder `hardLink` function for now. The PR can best be read patch by patch.


src/Database/LSMTree/Internal.hs
```diff
@@ -424,29 +423,20 @@ openSession tr hfs hbio dir = do
...
+        mkSession sessionFileLock
 
     restoreSession sessionFileLock = do
         traceWith tr TraceRestoreSession
         -- If the layouts are wrong, we throw an exception, and the lock file
         -- is automatically released by bracketOnError.
         checkTopLevelDirLayout
+
+        FS.removeDirectoryRecursive hfs activeDirPath -- TODO: exceptions safety
+        FS.createDirectory hfs activeDirPath
```
jorisdral, line 435: I should address this in this PR. Might be as simple as 

```suggestion
        FS.removeDirectoryRecursive hfs activeDirPath 
          `finally` FS.createDirectoryIfMissing False hfs activeDirPath
```

src/Database/LSMTree/Internal/Run.hs
```diff
@@ -135,25 +136,17 @@ mkWeakBlobRef Run{runBlobFile} blobspan =
...
-
     FS.hClose runHasFS runKOpsFile
-      `finally`
-         BlobFile.removeReference runBlobFile
+    BlobFile.removeReference runBlobFile
+
+    FS.removeFile runHasFS (runKOpsPath runRunFsPaths)
+    FS.removeFile runHasFS (runFilterPath runRunFsPaths)
+    FS.removeFile runHasFS (runIndexPath runRunFsPaths)
+    FS.removeFile runHasFS (runChecksumsPath runRunFsPaths)
```
mheinzel, line 149: There is no effort any more to handle exceptions here. Do you still want to do that, or at least add a TODO?

test/Test/Database/LSMTree/Internal/Run.hs
```diff
@@ -183,7 +186,10 @@ prop_WriteAndOpen ::
   -> IO Property
 prop_WriteAndOpen fs hbio wb =
     withRun fs hbio (simplePath 1337) (serialiseRunData wb) $ \written -> do
-      loaded <- openFromDisk fs hbio CacheRunData (simplePath 1337)
+      let paths = Run.runRunFsPaths written
+          paths' = paths { runNumber = RunNumber 17}
+      hardLinkRunFiles fs hbio paths paths'
+      loaded <- openFromDisk fs hbio CacheRunData (simplePath 17)
```
mheinzel, line 192: Oh yeah, this test needed fixing. :+1:

src/Database/LSMTree/Internal.hs
mheinzel: Yes, otherwise if we fail in the middle, where no active dir exists, we can currently never open that session again.

Other possible approaches would be to only remove the content of the active dir or make `checkTopLevelDirLayout` more lenient (creating the the directory if it doesn't exist).

test/Test/Database/LSMTree/Internal.hs
```diff
@@ -49,11 +37,6 @@ tests = testGroup "Test.Database.LSMTree.Internal" [
         , testProperty "twiceOpenSession" twiceOpenSession
         , testCase "sessionDirLayoutMismatch" sessionDirLayoutMismatch
         , testCase "sessionDirDoesNotExist" sessionDirDoesNotExist
-        , testProperty "prop_interimRestoreSessionUniqueRunNames"
-            prop_interimRestoreSessionUniqueRunNames
-        ]
-    , testGroup "Table" [
-          testProperty "prop_interimOpenSnapshot" prop_interimOpenSnapshot
```
mheinzel, line 56: Do we want to add new tests in exchange? We still test snapshotting through the public API in the state machine and class-based tests, but we usually also have tests for the internals to catch issues more directly.

src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -287,23 +168,86 @@ snapIncomingRun (Merging MergingRun {..}) = do
...
-    pure (SnapOngoingMerge (V.map runNumber rs) (SpentCredits spentCredits) (Merge.mergeLevel m))
+    pure (SnapOngoingMerge rs (SpentCredits spentCredits) (Merge.mergeLevel m))
+
+{-------------------------------------------------------------------------------
+  Runs
+-------------------------------------------------------------------------------}
 
-runNumber :: Run m h -> RunNumber
-runNumber r = Paths.runNumber (Run.runRunFsPaths r)
+-- TODO: temp registry
```
mheinzel, line 189: Do you still want to tackle this here or later?

src/Database/LSMTree/Internal.hs
jorisdral: Both alternatives are valid, though removing the dir contents is a bit more involved.  `listDirectory` just provides us with directory entry names and not the entry types (file/directory), which is something we'll have to figure out on the fly ourselves. I've gone with the `finally` approach for now.

src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -287,23 +168,86 @@ snapIncomingRun (Merging MergingRun {..}) = do
...
+snapshotRuns ::
+     MonadThrow m
+  => NamedSnapshotDir
+  -> SnapLevels (Run m h)
+  -> m (SnapLevels RunNumber)
+snapshotRuns (NamedSnapshotDir targetDir) levels = for levels $ \run -> do
+    let sourcePaths = Run.runRunFsPaths run
+    let targetPaths = sourcePaths { runDir = targetDir }
+    hardLinkRunFiles (Run.runHasFS run) (Run.runHasBlockIO run) sourcePaths targetPaths
+    pure (runNumber targetPaths)
```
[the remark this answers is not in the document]
jorisdral: Yup, it just means the function is slightly simpler

src/Database/LSMTree/Internal/Run.hs
jorisdral: I've added a TODO

src/Database/LSMTree/Internal/Snapshot.hs
jorisdral: Done!

test/Test/Database/LSMTree/Internal.hs
jorisdral: `interimOpenSnapshot` does not have much value over the class-based tests and state machine tests we already have, but since we don't actually test the session that much I've re-introduced some of the testing stuff into the `newSession` and `restoreSession` properties

src/Database/LSMTree/Internal.hs
```diff
@@ -1177,7 +1166,11 @@ openSnapshot sesh label tableType override snap resolve = do
           <- allocateTemp reg
                (WBB.new hfs blobpath)
                WBB.removeReference
-        tableLevels <- openLevels reg hfs hbio conf (sessionUniqCounter seshEnv) (sessionRoot seshEnv) resolve snappedLevels
+
+        let actDir = Paths.activeDir (sessionRoot seshEnv)
+        snappedLevels' <- openRuns hfs hbio conf (sessionUniqCounter seshEnv) snapDir actDir snappedLevels
+        tableLevels <- openLevels reg hfs hbio conf (sessionUniqCounter seshEnv) resolve actDir snappedLevels'
```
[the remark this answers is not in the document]
jorisdral: Renamed them a bit. Let me know what you think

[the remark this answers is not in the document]
mheinzel: Looks good!

mheinzel (approved):
Nice!

PR #467, opened 2024-11-13
Move general append utilities into general index module

jeltsch:
This pull request moves `Append` and `append` out of `Database.LSMTree.Internal.IndexCompact`. The reason is that `Append` is not specific to the compact index and `append` will not be either once the class of index types is introduced. The target of this move is a new module for indexes in general, situated in the `extras` package, as only tests and benchmarks are using `Append` and `append`.

### DIFF
--- a/src-extras/Database/LSMTree/Extras/Index.hs
+++ b/src-extras/Database/LSMTree/Extras/Index.hs
@@ -1,0 +1,72 @@
+module Database.LSMTree.Extras.Index
+(
+    Append (AppendSinglePage, AppendMultiPage),
+    append,
+    append'
+)
+where
+
+import           Control.DeepSeq (NFData (rnf))
+import           Control.Monad.ST.Strict (ST)
+import           Data.Foldable (toList)
+import           Data.Word (Word32)
+import           Database.LSMTree.Internal.Chunk (Chunk)
+import           Database.LSMTree.Internal.IndexCompactAcc (IndexCompactAcc)
+import qualified Database.LSMTree.Internal.IndexCompactAcc as IndexCompact
+                     (appendMulti, appendSingle)
+import           Database.LSMTree.Internal.IndexOrdinaryAcc (IndexOrdinaryAcc)
+import qualified Database.LSMTree.Internal.IndexOrdinaryAcc as IndexOrdinary
+                     (appendMulti, appendSingle)
+import           Database.LSMTree.Internal.Serialise (SerialisedKey)
+
+-- | Instruction for appending pages, to be used in conjunction with indexes.
+data Append
+    = {-|
+          Append a single page that fully comprises one or more key–value pairs.
+      -}
+      AppendSinglePage
+          SerialisedKey -- ^ Minimum key
+          SerialisedKey -- ^ Maximum key
+    | {-|
+          Append multiple pages that together comprise a single key–value pair.
+      -}
+      AppendMultiPage
+          SerialisedKey -- ^ Sole key
+          Word32        -- ^ Number of overflow pages
+
+instance NFData Append where
+
+    rnf (AppendSinglePage minKey maxKey)
+        = rnf minKey `seq` rnf maxKey
+    rnf (AppendMultiPage key overflowPageCount)
+        = rnf key `seq` rnf overflowPageCount
+
+{-|
+    Add information about appended pages to an index under incremental
+    construction.
+
+    Internally, 'append' uses 'IndexCompact.appendSingle' and
+    'IndexCompact.appendMulti', and the usage restrictions of those functions
+    apply also here.
+-}
+append :: Append -> IndexCompactAcc s -> ST s [Chunk]
+append instruction indexAcc = case instruction of
+    AppendSinglePage minKey maxKey
+        -> toList <$> IndexCompact.appendSingle (minKey, maxKey) indexAcc
+    AppendMultiPage key overflowPageCount
+        -> IndexCompact.appendMulti (key, overflowPageCount) indexAcc
+
+{-|
+    A variant of 'append' for ordinary indexes, which is only used temporarily
+    until there is a type class of index types.
+
+    Internally, 'append'' uses 'IndexOrdinary.appendSingle' and
+    'IndexOrdinary.appendMulti', and the usage restrictions of those functions
+    apply also here.
+-}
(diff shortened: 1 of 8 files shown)


src/Database/LSMTree/Internal/IndexCompactAcc.hs
```diff
@@ -126,26 +122,6 @@ newPinnedMVec64 lenWords = do
...
-  rnf (AppendSinglePage kmin kmax)  = rnf kmin `seq` rnf kmax
-  rnf (AppendMultiPage k nOverflow) = rnf k `seq` rnf nOverflow
-
--- | Append a new page entry to a mutable compact index.
---
--- INVARIANTS: see [construction invariants](#construction-invariants).
-append :: Append -> IndexCompactAcc s -> ST s [Chunk]
-append (AppendSinglePage kmin kmax) ica = toList <$> appendSingle (kmin, kmax) ica
-append (AppendMultiPage k n) ica        = appendMulti (k, n) ica
-
```
jorisdral, line 148: Just a small suggestion for PR review ergonomics: when code is moved to a different module and subsequently changed as well, it's hard to see from the diff what the interesting code changes are. My usual approach is to move the code in one commit, and do the interesting changes in a different commit.

Right now we're only talking about a small bit of code, so it's fine and doesn't have to change in this PR, but it might be good to keep in mind for future PRs

src/Database/LSMTree/Internal/IndexCompactAcc.hs
```diff
@@ -126,26 +122,6 @@ newPinnedMVec64 lenWords = do
...
-    -- a single page.
-  | AppendMultiPage SerialisedKey Word32 -- ^ Number of overflow pages
-
-instance NFData Append where
-  rnf (AppendSinglePage kmin kmax)  = rnf kmin `seq` rnf kmax
-  rnf (AppendMultiPage k nOverflow) = rnf k `seq` rnf nOverflow
-
--- | Append a new page entry to a mutable compact index.
---
--- INVARIANTS: see [construction invariants](#construction-invariants).
```
jorisdral, line 144: This invariant goes missing. Are we going to have preconditions/invariants on the index class that capture the requirements for both the compact and ordinary index?

src-extras/Database/LSMTree/Extras/Index.hs
```diff
@@ -0,0 +1,64 @@
...
+          SerialisedKey -- ^ Maximum key
+      {-^
+          Append a single page that fully comprises one or more key–value pairs.
+      -}
+    | AppendMultiPage
+          SerialisedKey -- ^ Sole key
+          Word32        -- ^ Number of overflow pages
+      {-^
+          Append multiple pages that together comprise a single key–value pair.
+      -}
```
jorisdral, line 35: The suffix documentation for constructors isn't proper haddocks. Put the docs in front instead

```suggestion
-- | Instruction for appending pages, to be used in conjunction with indexes.
data Append
    = {- |
          Append a single page that fully comprises one or more key–value pairs.
      -}
      AppendSinglePage
          SerialisedKey -- ^ Minimum key
          SerialisedKey -- ^ Maximum key
    | {- |
          Append multiple pages that together comprise a single key–value pair.
      -}
      AppendMultiPage
          SerialisedKey -- ^ Sole key
          Word32        -- ^ Number of overflow pages
```

jorisdral (changes requested):

src/Database/LSMTree/Internal/IndexCompactAcc.hs
jeltsch: > My usual approach is to move the code in one commit, and do the interesting changes in a different commit.

Sounds very reasonable to me. I’ll try to remember this approach and use it in future pull requests of mine. Sadly, Git essentially doesn’t track file movement. Well, its manual page calls it right at the beginning “the stupid content tracker”. 😉

src-extras/Database/LSMTree/Extras/Index.hs
jeltsch: Actually, [suffix documentation of data constructors *is* possible](https://haskell-haddock.readthedocs.io/latest/markup.html#constructors-and-record-fields), but the documentation must come right after the constructors instead of the whole data type alternatives. Looking at this now, it is quite obvious to me that what I did can’t work, as it is not clear whether documentation following a data type alternative with at least one field refers to the data constructor or the last field.

jeltsch: Fixed in 9e8ec9e333c246f162ca7f2a73e385c2077ec09a.

src/Database/LSMTree/Internal/IndexCompactAcc.hs
jeltsch: Yes, it is the intention to have the sortedness invariant, which is the one invariant currently stated for the compact index, as a general invariant for all indexes and to have corresponding documentation along with the index type class declaration. That said, it’s surely a good idea to refer to the usage restrictions for appending also here. I’ve added a corresponding statement that works in the present situation and will continue to work when the index type class is introduced. The commit introducing this change is 0bf6c30ec10fa89c0fcc0489f2b75dba6a4e4edd.

src/Database/LSMTree/Internal/IndexCompactAcc.hs
jorisdral: One scenario where Git is smart enough is when a file is first moved and then its contents are changed. The diff will show that the file is renamed, instead of showing that one file was deleted and one new file was created. And the diff of the renamed file just shows the changed lines, not the unchanged lines. But that's about as smart as it gets in my experience

jorisdral (approved):
LGTM! Let's squash and then press merge

src/Database/LSMTree/Internal/IndexCompactAcc.hs
jeltsch: Hmm, now I wonder whether this mishap might have been because of me not actually moving the file but only moving its contents. Does Git perhaps track inode numbers? 🤔

jeltsch: I found the answer: Git does not track inode numbers. It detects renames based on content only. See the documentation of the `-M` option in the `git-diff(1)` man page.

PR #468, opened 2024-11-14
Remove redundant constraints in the public API

jorisdral:


src/Database/LSMTree/Monoidal.hs
```diff
@@ -681,7 +672,7 @@ union :: forall m k v.
   => Table m k v
   -> Table m k v
   -> m (Table m k v)
-union = undefined
+union = error "union: not yet implemented" $ union @m @k @v
```
mheinzel, line 675: So the recursive call avoids redundant constraint warnings?

mheinzel (approved):
LGTM

src/Database/LSMTree/Monoidal.hs
jorisdral: Yes, exactly

PR #469, opened 2024-11-14
Introduce a unified `LSMTree` API 

jorisdral:


src/Database/LSMTree.hs
```diff
@@ -0,0 +1,551 @@
...
+{-------------------------------------------------------------------------------
+  Tables
+-------------------------------------------------------------------------------}
+
+type Table = Internal.Table'
+
+{-# SPECIALISE withTable ::
+     Session IO
+  -> Common.TableConfig
+  -> (Table IO k v b -> IO a)
```
mheinzel, line 128: In the other modules we use `blob` for that type parameter. While I generally think it makes sense to use only `b`, only renaming it here, but not in the 90% identical other modules doesn't seem ideal. For example, I looked at the difference of this module to the `Normal` one, and might again in the future to make sure the modules stay in sync. But now this b/blob difference shows up as noise in the diff.

src/Database/LSMTree.hs
```diff
@@ -0,0 +1,551 @@
...
+
+{-------------------------------------------------------------------------------
+  Monoidal value resolution
+-------------------------------------------------------------------------------}
+
+resolve ::  forall v. ResolveValue v => Proxy v -> Internal.ResolveSerialisedValue
+resolve = coerce . resolveValue
+
+-- | Newtype wrapper for use with deriving via, which resolves values as a
+-- 'Data.Semigroup.First' semigroup.
```
mheinzel, line 537: Another way than DerivingVia would be to use `Table k (ResolveAsFirst v) b`, which we could also provide a type synonym for. Something like this?

```haskell
type SimpleTable k v = Table k (ResolveAsFirst v) Void
type BlobTable k v b = Table k (ResolveAsFirst v) b
type MonoidalTable k v = Table k v Void
```

But not too important for now, I think, as the main purpose of this module is for testing. Later on, if we really want this module to be the entry point for users, we can still work on making that a nicer experience.

src/Database/LSMTree.hs
```diff
@@ -0,0 +1,551 @@
...
+--
+-- The instance makes 'Insert's isomorphic to 'Mupsert's. This can be a sensible
+-- default if there is no intent to use 'Mupsert's.
```
mheinzel, line 540: I'd rather say that the instance makes mupserts behave like inserts. So, the other way around, and without talking about isomorphisms, which we don't define in this context.

And I would explicitly mention that this avoids the need of providing a custom instance.

src/Database/LSMTree.hs
```diff
@@ -0,0 +1,551 @@
+-- TODO: document this module
```
mheinzel, line 1: What's your plan for this?

For the individual function Haddocks, they can go out of sync, but we already have to consider keeping `Normal`/`Monoidal` in sync, so maybe having a third copy is not too bad?

And I would definitely add a comment to the module explaining that it's experimental and mainly for testing, as well as referring to the `Normal`/`Monoidal` version for full documentation. Also, a TODO to decide which public API(s) to provide in the end.

src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -61,7 +61,7 @@ import           System.FS.BlockIO.API (HasBlockIO)
 newtype SnapshotLabel = SnapshotLabel Text
   deriving stock (Show, Eq)
 
-data SnapshotTableType = SnapNormalTable | SnapMonoidalTable
+data SnapshotTableType = SnapNormalTable | SnapMonoidalTable | SnapFullTable
```
mheinzel, line 64: image

Long term, we should probably either have one or two types of tables, but it kind of makes sense while we're undecided on the best way. Maybe add a TODO?

src/Database/LSMTree/Internal.hs
```diff
@@ -129,6 +131,18 @@ data Session' m = forall h. Typeable h => Session' !(Session m h)
 instance NFData (Session' m) where
   rnf (Session' s) = rnf s
 
+type Table' :: (Type -> Type) -> Type -> Type -> Type -> Type
+data Table' m k v b = forall h. Typeable h => Table' (Table m h)
+
+instance NFData (Table' m k v b) where
+  rnf (Table' t) = rnf t
```
mheinzel, line 138: I'm not a big fan of the pattern of defining these types here and using synonyms in the API modules, as synonyms can leak into error messages and it also makes the diff between them larger vs having three `data Table` there.

But the pattern already exists, so it makes sense to follow it here. :+1:

src/Database/LSMTree.hs
jorisdral: Makes sense. Maybe I'll change it back to `blob` here, and make the renaming change in a separate PR

src/Database/LSMTree.hs
jorisdral: We discussed this 1-on-1 already, but for now I think we should omit the documentation until we'

* decide how to deduplicate documentation between modules, or
* decide that we want to duplicate haddocks across different modules

I'll be more verbose in the TODO here

src/Database/LSMTree/Internal.hs
jorisdral: In our current setup, we need to expose the constructors for our `NoThunks` assertions, but @dcoutts suggested we do not expose the constructors from the public API. That's why I moved them here and added type synonyms in the public APIs

jorisdral: IMO it would have been fine to expose the constructors in the public API, since if you want to do anything with the internal `Table` then you'll have to import `Internal` modules anyway

src/Database/LSMTree.hs
mheinzel: Makes sense not to duplicate, but yes, in that case a short header explaining why this module exists (and doesn't have docs) is still good to have.

src/Database/LSMTree.hs
jorisdral: Did the renaming in https://github.com/IntersectMBO/lsm-tree/pull/480

mheinzel (approved):

PR #471, opened 2024-11-14
Moves `ChecksumHandle` to its own module

wenkokke:
This PR:

- Moves `ChecksumHandle` and its core functions to their own module (`Database.LSMTree.Internal.ChecksumHandle`).
- Generalises the helper functions in `RunBuilder` (`writeRawPage`, `writeRawOverflowPages`, `writeBlob`, `copyBlob`, `writeFilter`, `writeIndexHeader`, `writeIndexChunk`, `writeIndexFinal`) to only take the arguments they need, rather than the entire `RunBuilder`.
- Moves those helper functions to the new `ChecksumHandle` module. This enables them to be reused by the `WriteBufferWriter`. (Technically, the latter four are not needed, but it seemed better to generalise and move them for consistency.)

### DIFF
--- a/.stylish-haskell.yaml
+++ b/.stylish-haskell.yaml
@@ -511,3 +511,6 @@ language_extensions:
 # We run stylish from the root of the repo which is a project so this flag
 # doesn't apply
 cabal: false
+
+# Return an exit failure on format.
+exit_code: error_on_format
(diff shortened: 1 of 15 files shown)


wenkokke:
The contributed commits look off, so I'm reopening with only [51b1041](https://github.com/IntersectMBO/lsm-tree/pull/471/commits/51b10412de9a3c9e16f648e816597fda68365b9f) cherry-picked.

PR #473, opened 2024-11-15
Revisit snapshot labels

jorisdral:
Require the user to pass in snapshot labels explicitly. The responsibility of managing label falls on the user, just like the user has the reponsibility of managing snapshot names.


jorisdral:
@wenkokke thanks for the review. I probably won't merge the PR yet until #417 has landed on `main`

PR #475, opened 2024-11-18
Add mupserts and table unions to the state machine tests

jorisdral:
This is a follow-up to #469. This PR is put in draft mode until we've come to a conclusion about the proposal in #469.

In this PR, we first create a class and reference implementation for the "full" API that is introduced in #469. Then, we switch the state machine tests to use this class instead of the `Normal` class. Finally, we can add infrastructure for both mupserts and table unions to the state machine tests. We start testing mupserts immediately, but table union operations are not yet generated because table union does not have an implementation yet.

### DIFF
--- a/lsm-tree.cabal
+++ b/lsm-tree.cabal
@@ -348,6 +348,7 @@ test-suite lsm-tree-test
     Database.LSMTree.Class.Monoidal
     Database.LSMTree.Class.Normal
     Database.LSMTree.Model
+    Database.LSMTree.Model.IO
     Database.LSMTree.Model.IO.Monoidal
     Database.LSMTree.Model.IO.Normal
     Database.LSMTree.Model.Session
--- a/lsm-tree.cabal
+++ b/lsm-tree.cabal
@@ -343,6 +343,7 @@ test-suite lsm-tree-test
   hs-source-dirs: test
   main-is:        Main.hs
   other-modules:
+    Database.LSMTree.Class
     Database.LSMTree.Class.Common
     Database.LSMTree.Class.Monoidal
     Database.LSMTree.Class.Normal
--- a/lsm-tree.cabal
+++ b/lsm-tree.cabal
@@ -343,6 +343,7 @@ test-suite lsm-tree-test
   hs-source-dirs: test
   main-is:        Main.hs
   other-modules:
+    Database.LSMTree.Class.Common
     Database.LSMTree.Class.Monoidal
     Database.LSMTree.Class.Normal
     Database.LSMTree.Model
--- a/lsm-tree.cabal
+++ b/lsm-tree.cabal
@@ -383,10 +383,10 @@ test-suite lsm-tree-test
     Test.Database.LSMTree.Internal.Vector.Growing
     Test.Database.LSMTree.Model.Table
     Test.Database.LSMTree.Monoidal
-    Test.Database.LSMTree.Normal.StateMachine
-    Test.Database.LSMTree.Normal.StateMachine.DL
-    Test.Database.LSMTree.Normal.StateMachine.Op
-    Test.Database.LSMTree.Normal.UnitTests
+    Test.Database.LSMTree.StateMachine
+    Test.Database.LSMTree.StateMachine.DL
+    Test.Database.LSMTree.StateMachine.Op
+    Test.Database.LSMTree.UnitTests
     Test.System.Posix.Fcntl.NoCache
     Test.Util.FS
     Test.Util.Orphans
--- a/src-extras/Database/LSMTree/Extras/Generators.hs
+++ b/src-extras/Database/LSMTree/Extras/Generators.hs
@@ -46,6 +46,7 @@ import           Data.List (sort)
 import qualified Data.Primitive.ByteArray as BA
 import qualified Data.Vector.Primitive as VP
 import           Data.Word
+import qualified Database.LSMTree as Unified
 import           Database.LSMTree.Common (Range (..))
 import           Database.LSMTree.Extras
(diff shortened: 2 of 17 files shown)


jorisdral:
#469 was merged, so I'll update this PR before undrafting it

test/Test/Database/LSMTree/Normal/StateMachine.hs
```diff
@@ -1183,11 +1224,12 @@ arbitraryActionWithVars _ ctx (ModelState st _stats) =
...
       | otherwise      =
         [ (1,  fmap Some $ Close <$> genTableVar)
-        , (10, fmap Some $ Lookups <$> genLookupKeys <*> genTableVar)
-        , (5,  fmap Some $ RangeLookup <$> genRange <*> genTableVar)
-        , (10, fmap Some $ Updates <$> genUpdates <*> genTableVar)
-        , (10, fmap Some $ Inserts <$> genInserts <*> genTableVar)
-        , (10, fmap Some $ Deletes <$> genDeletes <*> genTableVar)
+        , (8, fmap Some $ Lookups <$> genLookupKeys <*> genTableVar)
+        , (4,  fmap Some $ RangeLookup <$> genRange <*> genTableVar)
+        , (8, fmap Some $ Updates <$> genUpdates <*> genTableVar)
```
jorisdral, line 1229:
```suggestion

        , (4, fmap Some $ RangeLookup <$> genRange <*> genTableVar)
```

test/Test/Database/LSMTree/Normal/StateMachine.hs
```diff
@@ -1183,11 +1224,12 @@ arbitraryActionWithVars _ ctx (ModelState st _stats) =
       | null tableVars = []
       | otherwise      =
         [ (1,  fmap Some $ Close <$> genTableVar)
-        , (10, fmap Some $ Lookups <$> genLookupKeys <*> genTableVar)
-        , (5,  fmap Some $ RangeLookup <$> genRange <*> genTableVar)
-        , (10, fmap Some $ Updates <$> genUpdates <*> genTableVar)
-        , (10, fmap Some $ Inserts <$> genInserts <*> genTableVar)
-        , (10, fmap Some $ Deletes <$> genDeletes <*> genTableVar)
+        , (8, fmap Some $ Lookups <$> genLookupKeys <*> genTableVar)
+        , (4,  fmap Some $ RangeLookup <$> genRange <*> genTableVar)
```
jorisdral, line 1228:
```suggestion

        , (4, fmap Some $ RangeLookup <$> genRange <*> genTableVar)
```

lsm-tree.cabal
```diff
@@ -342,9 +342,12 @@ test-suite lsm-tree-test
   hs-source-dirs: test
   main-is:        Main.hs
   other-modules:
+    Database.LSMTree.Class
+    Database.LSMTree.Class.Common
     Database.LSMTree.Class.Monoidal
     Database.LSMTree.Class.Normal
```
mheinzel, line 348: So we added the unified API in #469 to allow writing tests against it. Here we add a *third* version of `IsTable`, the IO model etc. for the the unified API. But why not unify the model/tests as well, resulting in a single IO model and single `IsTable` class? That would make the tests much leaner.

The only argument against it that I can think of is that we're not yet sure which API(s) we want to provide in the end. If we later remove the unified API again, we might have to re-duplicate the model/tests. :shrug:

jorisdral: Sure, it would be nice if we could eventually deduplicate the test stuff too, but it's a lot more work, and not strictly necessary for the statemachine tests

mheinzel: Okay, but would be nice to discuss and tackle soonish then.

src-extras/Database/LSMTree/Extras/Generators.hs
```diff
@@ -46,6 +46,7 @@ import           Data.List (sort)
 import qualified Data.Primitive.ByteArray as BA
 import qualified Data.Vector.Primitive as VP
 import           Data.Word
+import qualified Database.LSMTree as LSMTree
```
mheinzel, line 49: To fit the pattern of `Normal`/`Monoidal`, would it make sense to import it as `Unified` here and in other places?

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -1412,15 +1472,16 @@ updateStats action lookUp modelBefore _modelAfter result =
...
-              count (i, iwb, d) (_, upd) = case upd  of
-                R.Insert _ Nothing -> (i+1, iwb  , d  )
-                R.Insert _ Just{}  -> (i  , iwb+1, d  )
-                R.Delete{}         -> (i  , iwb  , d+1)
+                    -> (Int, Int, Int, Int)
+              count (i, iwb, d, m) (_, upd) = case upd  of
+                R.Insert _ Nothing -> (i+1, iwb  , d  , m    )
+                R.Insert _ Just{}  -> (i  , iwb+1, d  , m    )
+                R.Delete{}         -> (i  , iwb  , d+1, m    )
+                R.Mupsert{}        -> (i  , iwb  , d  , m + 1)
```
mheinzel, line 1484:
```suggestion
                R.Insert _ Nothing -> (i+1, iwb  , d  , m  )
                R.Insert _ Just{}  -> (i  , iwb+1, d  , m  )
                R.Delete{}         -> (i  , iwb  , d+1, m  )
                R.Mupsert{}        -> (i  , iwb  , d  , m+1)
```

mheinzel (approved):

lsm-tree.cabal
jorisdral: Working on it in #486

PR #476, opened 2024-11-18
Introduce classes for indexes and their accumulators

jeltsch:
This pull request introduces type classes that provide a common interface to different types of indexes and their accumulators and instantiates these type classes for the compact and the ordinary index.

### DIFF
--- a/lsm-tree.cabal
+++ b/lsm-tree.cabal
@@ -130,10 +130,10 @@ library
     Database.LSMTree.Internal.CRC32C
     Database.LSMTree.Internal.Cursor
     Database.LSMTree.Internal.Entry
-    Database.LSMTree.Internal.IndexCompact
-    Database.LSMTree.Internal.IndexCompactAcc
-    Database.LSMTree.Internal.IndexOrdinary
-    Database.LSMTree.Internal.IndexOrdinaryAcc
+    Database.LSMTree.Internal.Index.Compact
+    Database.LSMTree.Internal.Index.CompactAcc
+    Database.LSMTree.Internal.Index.Ordinary
+    Database.LSMTree.Internal.Index.OrdinaryAcc
     Database.LSMTree.Internal.Lookup
     Database.LSMTree.Internal.Merge
     Database.LSMTree.Internal.MergeSchedule
@@ -356,8 +356,8 @@ test-suite lsm-tree-test
     Test.Database.LSMTree.Internal.Chunk
     Test.Database.LSMTree.Internal.CRC32C
     Test.Database.LSMTree.Internal.Entry
-    Test.Database.LSMTree.Internal.IndexCompact
-    Test.Database.LSMTree.Internal.IndexOrdinary
+    Test.Database.LSMTree.Internal.Index.Compact
+    Test.Database.LSMTree.Internal.Index.Ordinary
     Test.Database.LSMTree.Internal.Lookup
     Test.Database.LSMTree.Internal.Merge
     Test.Database.LSMTree.Internal.Monkey
@@ -452,7 +452,7 @@ benchmark lsm-tree-micro-bench
   main-is:        Main.hs
   other-modules:
     Bench.Database.LSMTree.Internal.BloomFilter
-    Bench.Database.LSMTree.Internal.IndexCompact
+    Bench.Database.LSMTree.Internal.Index.Compact
     Bench.Database.LSMTree.Internal.Lookup
     Bench.Database.LSMTree.Internal.Merge
     Bench.Database.LSMTree.Internal.RawPage
--- a/lsm-tree.cabal
+++ b/lsm-tree.cabal
@@ -130,6 +130,7 @@ library
     Database.LSMTree.Internal.CRC32C
     Database.LSMTree.Internal.Cursor
     Database.LSMTree.Internal.Entry
+    Database.LSMTree.Internal.Index
     Database.LSMTree.Internal.Index.Compact
     Database.LSMTree.Internal.Index.CompactAcc
     Database.LSMTree.Internal.Index.Ordinary
--- a/src-extras/Database/LSMTree/Extras/Index.hs
+++ b/src-extras/Database/LSMTree/Extras/Index.hs
@@ -1,8 +1,11 @@
+{-|
+    Provides additional support for working with fence pointer indexes and their
+    accumulators.
+-}
 module Database.LSMTree.Extras.Index
 (
     Append (AppendSinglePage, AppendMultiPage),
-    append,
-    append'
+    append
 )
 where
 
@@ -11,12 +14,8 @@ import           Control.Monad.ST.Strict (ST)
 import           Data.Foldable (toList)
 import           Data.Word (Word32)
 import           Database.LSMTree.Internal.Chunk (Chunk)
-import           Database.LSMTree.Internal.Index.CompactAcc (IndexCompactAcc)
-import qualified Database.LSMTree.Internal.Index.CompactAcc as IndexCompact
-                     (appendMulti, appendSingle)
-import           Database.LSMTree.Internal.Index.OrdinaryAcc (IndexOrdinaryAcc)
-import qualified Database.LSMTree.Internal.Index.OrdinaryAcc as IndexOrdinary
-                     (appendMulti, appendSingle)
+import           Database.LSMTree.Internal.Index (IndexAcc, appendMulti,
+                     appendSingle)
 import           Database.LSMTree.Internal.Serialise (SerialisedKey)
 
 -- | Instruction for appending pages, to be used in conjunction with indexes.
@@ -42,31 +41,15 @@ instance NFData Append where
         = rnf key `seq` rnf overflowPageCount
 
 {-|
-    Add information about appended pages to an index under incremental
-    construction.
+    Adds information about appended pages to an index and outputs newly
+    available chunks.
 
-    Internally, 'append' uses 'IndexCompact.appendSingle' and
-    'IndexCompact.appendMulti', and the usage restrictions of those functions
-    apply also here.
+    See the documentation of the 'IndexAcc' class for constraints to adhere to.
 -}
-append :: Append -> IndexCompactAcc s -> ST s [Chunk]
+append :: IndexAcc j => Append -> j s -> ST s [Chunk]
 append instruction indexAcc = case instruction of
     AppendSinglePage minKey maxKey
-        -> toList <$> IndexCompact.appendSingle (minKey, maxKey) indexAcc
+        -> toList <$> appendSingle (minKey, maxKey) indexAcc
     AppendMultiPage key overflowPageCount
-        -> IndexCompact.appendMulti (key, overflowPageCount) indexAcc
-
-{-|
-    A variant of 'append' for ordinary indexes, which is only used temporarily
-    until there is a type class of index types.
-
-    Internally, 'append'' uses 'IndexOrdinary.appendSingle' and
-    'IndexOrdinary.appendMulti', and the usage restrictions of those functions
-    apply also here.
--}
-append' :: Append -> IndexOrdinaryAcc s -> ST s [Chunk]
-append' instruction indexAcc = case instruction of
-    AppendSinglePage minKey maxKey
-        -> toList <$> IndexOrdinary.appendSingle (minKey, maxKey) indexAcc
-    AppendMultiPage key overflowPageCount
-        -> IndexOrdinary.appendMulti (key, overflowPageCount) indexAcc
+        -> appendMulti (key, overflowPageCount) indexAcc
+{-# INLINABLE append #-}
--- a/src-extras/Database/LSMTree/Extras/Index.hs
+++ b/src-extras/Database/LSMTree/Extras/Index.hs
@@ -11,11 +11,11 @@ import           Control.Monad.ST.Strict (ST)
 import           Data.Foldable (toList)
 import           Data.Word (Word32)
 import           Database.LSMTree.Internal.Chunk (Chunk)
-import           Database.LSMTree.Internal.IndexCompactAcc (IndexCompactAcc)
-import qualified Database.LSMTree.Internal.IndexCompactAcc as IndexCompact
+import           Database.LSMTree.Internal.Index.CompactAcc (IndexCompactAcc)
+import qualified Database.LSMTree.Internal.Index.CompactAcc as IndexCompact
                      (appendMulti, appendSingle)
-import           Database.LSMTree.Internal.IndexOrdinaryAcc (IndexOrdinaryAcc)
-import qualified Database.LSMTree.Internal.IndexOrdinaryAcc as IndexOrdinary
+import           Database.LSMTree.Internal.Index.OrdinaryAcc (IndexOrdinaryAcc)
+import qualified Database.LSMTree.Internal.Index.OrdinaryAcc as IndexOrdinary
                      (appendMulti, appendSingle)
 import           Database.LSMTree.Internal.Serialise (SerialisedKey)
 
--- a/src/Database/LSMTree/Internal/Index.hs
+++ b/src/Database/LSMTree/Internal/Index.hs
@@ -1,0 +1,100 @@
+{-# LANGUAGE TypeFamilies #-}
+
+{-|
+    Provides a common interface to different types of fence pointer indexes and
+    their accumulators.
+-}
+module Database.LSMTree.Internal.Index
+(
+    Index (search, sizeInPages, fromSBS),
+    IndexAcc (ResultingIndex, appendSingle, appendMulti, unsafeEnd)
+)
+where
+
+import           Control.Monad.ST.Strict (ST)
+import           Data.ByteString.Short (ShortByteString)
+import           Data.Word (Word32)
+import           Database.LSMTree.Internal.Chunk (Chunk)
+import           Database.LSMTree.Internal.Entry (NumEntries)
+import           Database.LSMTree.Internal.Page (NumPages, PageSpan)
+import           Database.LSMTree.Internal.Serialise (SerialisedKey)
+
+-- | The class of index types.
+class Index i where
+
+    {-|
+        Searches for a page span that contains a key–value pair with the given
+        key. If there is indeed such a pair, the result is the corresponding
+        page span; if there is no such pair, the result is an arbitrary but
+        valid page span.
+    -}
+    search :: SerialisedKey -> i -> PageSpan
+
+    -- | Yields the number of pages covered by an index.
+    sizeInPages :: i -> NumPages
+
+    {-|
+        Reads an index along with the number of entries of the respective run
+        from a byte string.
+
+        The byte string must contain the serialised index exactly, with no
+        leading or trailing space. Furthermore, its contents must be stored
+        64-bit-aligned.
+
+        The contents of the byte string may be directly used as the backing
+        memory for the constructed index. Currently, this is done for compact
+        indexes.
+
+        For deserialising numbers, the endianness of the host system is used. If
+        serialisation has been done with a different endianness, this mismatch
+        is detected by looking at the type–version indicator.
+    -}
+    fromSBS :: ShortByteString -> Either String (NumEntries, i)
+
+{-|
+    The class of index accumulator types, where an index accumulator denotes an
+    index under incremental construction.
+
+    Incremental index construction is only guaranteed to work correctly when the
+    following conditions are met:
+
+      * The supplied key ranges do not overlap and are given in ascending order.
+
+      * Each supplied key is at least 8 and at most 65535 bytes long.
+        (Currently, construction of compact indexes needs the former and
+        construction of ordinary indexes needs the latter bound.)
+-}
+class Index (ResultingIndex j) => IndexAcc j where
+
+    -- | The type of indexes constructed by accumulators of a certain type
+    type ResultingIndex j
+
+    {-|
+        Adds information about a single page that fully comprises one or more
+        key–value pairs to an index and outputs newly available chunks.
+
+        See the documentation of the 'IndexAcc' class for constraints to adhere
+        to.
+    -}
+    appendSingle :: (SerialisedKey, SerialisedKey) -> j s -> ST s (Maybe Chunk)
+
+    {-|
+        Adds information about multiple pages that together comprise a single
+        key–value pair to an index and outputs newly available chunks.
+
+        The provided 'Word32' value denotes the number of /overflow/ pages, so
+        that the number of pages that comprise the key–value pair is the
+        successor of that number.
+
+        See the documentation of the 'IndexAcc' class for constraints to adhere
+        to.
+    -}
+    appendMulti :: (SerialisedKey, Word32) -> j s -> ST s [Chunk]
+
+    {-|
+        Returns the constructed index, along with a final chunk in case the
+        serialised key list has not been fully output yet, thereby invalidating
+        the index under construction. Executing @unsafeEnd index@ is only safe
+        when @index@ is not used afterwards.
+    -}
+    unsafeEnd :: j s -> ST s (Maybe Chunk, ResultingIndex j)
--- a/src/Database/LSMTree/Internal/Index/Compact.hs
+++ b/src/Database/LSMTree/Internal/Index/Compact.hs
@@ -2,7 +2,7 @@
 --
 -- TODO: add utility functions for clash probability calculations
 --
-module Database.LSMTree.Internal.IndexCompact (
+module Database.LSMTree.Internal.Index.Compact (
     -- $compact
     IndexCompact (..)
     -- * Queries
@@ -59,7 +59,7 @@ import           Database.LSMTree.Internal.Vector
   @i@, to min-max information for keys on that page.
 
   Fence-pointer indexes can be constructed and serialised incrementally, see
-  module "Database.LSMTree.Internal.IndexCompactAcc".
+  module "Database.LSMTree.Internal.Index.CompactAcc".
 
   Given a serialised target key @k@, an index can be 'search'ed to find a disk
   page @i@ that /might/ contain @k@. Fence-pointer indices offer no guarantee of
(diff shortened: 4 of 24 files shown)


src/Database/LSMTree/Internal/Index/CompactAcc.hs
```diff
@@ -9,14 +10,10 @@
...
 -}
-module Database.LSMTree.Internal.IndexCompactAcc (
+module Database.LSMTree.Internal.Index.CompactAcc (
     -- * Construction
-    -- $construction-invariants
     IndexCompactAcc (..)
   , new
-  , appendSingle
-  , appendMulti
-  , unsafeEnd
```
jorisdral, line 19: I think there is value in keeping these exports

src/Database/LSMTree/Internal/Index/Compact.hs
```diff
@@ -378,11 +376,6 @@ instance NFData IndexCompact where
   Queries
 -------------------------------------------------------------------------------}
 
--- |  Searches for a page span that contains a key–value pair with the given key.
---
--- If there is indeed such a pair, the result is the corresponding page span; if
--- there is no such pair, the result is an arbitrary but valid page span.
---
```
jorisdral, line 385: If the idea here is that we can remove the comment because the documentation is now elsewhere (i.e., the `Index` class), then we should link to that documentation here

src/Database/LSMTree/Internal/Index/Compact.hs
```diff
@@ -548,13 +540,6 @@ putPaddingTo64 written
   Deserialisation
 -------------------------------------------------------------------------------}
 
--- | The input bytestring must be 64 bit aligned and exactly contain the
--- serialised compact index, with no leading or trailing space.
--- It is directly used as the backing memory for the compact index.
---
--- Also note that the implementation reads values in little-endian byte order.
--- If the file has been serialised in big-endian order, the mismatch will be
--- detected by looking at the type–version indicator.
```
jorisdral, line 557: I don't know how to better solve this, but it seems unfortunate that we lose documentation on these functions, even if the documentation is moved there. Maybe link to where the documentation is now found?

src/Database/LSMTree/Internal/Index/CompactAcc.hs
jorisdral, line 1: Same comments about removing documentation and linking to replacement documentation apply to this module

src/Database/LSMTree/Internal/Index.hs
jorisdral, line 1: Have you considered defining the instances for `IndexCompact` and `IndexOrdinary` in this module?

src/Database/LSMTree/Internal/Index.hs
```diff
@@ -0,0 +1,90 @@
...
+
+    Incremental index construction is only guaranteed to work correctly when the
+    following conditions are met:
+
+      * The provided information about pages is not in conflict with the
+        requirement that the keys in a run are ascending.
+
+      * The length of each key can be represented by a 16-bit word.
+-}
+class Index (ResultingIndex j) => IndexAcc j where
```
jorisdral, line 57: Shouldn't `new` also be part of the class?

test/Test/Database/LSMTree/Internal/Index/Ordinary.hs
```diff
@@ -362,7 +362,8 @@ prop_numberOfEntriesFromSerialisedIndexWorks entryCount lastKeys
     where
 
     errorMsgOrEntryCount :: Either String NumEntries
-    errorMsgOrEntryCount = fst <$> fromSBS (serialisedIndex entryCount lastKeys)
+    errorMsgOrEntryCount
+        = fst <$> fromSBS @IndexOrdinary (serialisedIndex entryCount lastKeys)
```
jorisdral, line 366: If we kept the exports in the `IndexOrdinary` module, then we wouldn't have to apply a type here

src/Database/LSMTree/Internal/Index/Ordinary.hs
jorisdral, line 1: Same comments about removing documentation and linking to replacement documentation apply to this module

src/Database/LSMTree/Internal/Index/Ordinary.hs
```diff
@@ -4,13 +4,7 @@
 module Database.LSMTree.Internal.Index.Ordinary
 (
     IndexOrdinary (IndexOrdinary),
-    toLastKeys,
-
-    -- * Search
-    search,
-
-    -- * Deserialisation
-    fromSBS
```
jorisdral: Same comment about keeping the exports

src/Database/LSMTree/Internal/Index/OrdinaryAcc.hs
```diff
@@ -7,10 +9,7 @@
 module Database.LSMTree.Internal.Index.OrdinaryAcc
 (
     IndexOrdinaryAcc,
-    new,
-    appendSingle,
-    appendMulti,
-    unsafeEnd
```
jorisdral, line 13: Same comment about keeping the exports

src/Database/LSMTree/Internal/Index/OrdinaryAcc.hs
jorisdral, line 1: Same comments about removing documentation and linking to replacement documentation apply to this module

jorisdral (changes requested):
Looks promising! 

I left some suggestions. There are two types of suggestions I'd like to elaborate on here:

* Keep the implementations for *concrete*  implementations of the index class and keep exporting them from the modules. We don't need the generality of the class in case we are testing/benchmarking concrete index types and their functions. Also, until we make the index selectable, using the concrete functions reduces the risk of missed specialisations.

* We seem to lose some documentation on concrete implementations for index class functions. Either we should keep some of that documentation, or explicitly link to where the new documentation can be found. And we should double check that we're not losing any information in the process

lsm-tree.cabal
```diff
@@ -130,10 +130,10 @@ library
...
     Database.LSMTree.Internal.Cursor
     Database.LSMTree.Internal.Entry
-    Database.LSMTree.Internal.IndexCompact
-    Database.LSMTree.Internal.IndexCompactAcc
-    Database.LSMTree.Internal.IndexOrdinary
-    Database.LSMTree.Internal.IndexOrdinaryAcc
+    Database.LSMTree.Internal.Index.Compact
+    Database.LSMTree.Internal.Index.CompactAcc
+    Database.LSMTree.Internal.Index.Ordinary
+    Database.LSMTree.Internal.Index.OrdinaryAcc
```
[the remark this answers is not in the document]
jeltsch: I think that expressing inherent structure is generally a good idea, which is why I introduced this additional nesting of module names. Of course, this should be weighed against difficulties this might cause.

As far as the source code itself is concerned, the only consequence of this change is that a dot has to be put here and there. Regarding the organization of the source code in the file system, there is the need for an additional subdirectory, which may or may not be considered burdensome. I myself tend to refer to individual source files via patterns like `src/**/Compact.hs`, which are supported by Vim, by Bash with the `globstar` option, and by the Z Shell (where they come from, as far as I know). With this approach, things even become simpler with this deeper hierarchy; for example, I don’t have to write the longer `src/**/IndexCompact.hs` anymore. The situation may be the other way round with different tools and different tool usage.

All in all, I prefer to have this change. You and @mheinzel said outside of GitHub that you don’t have a strong bias towards one or the other solution. I’d say that, if @dcoutts doesn’t have any strong opinion on this, we should go for this change.

src/Database/LSMTree/Internal/Index.hs
jeltsch: I think this wouldn’t be such a good idea. After all, type classes are open, and, while we currently have these two particular index types, we or someone else may want to add more in the future. When instantiating the index-related classes in `Database.LSMTree.Internal.Index`, one would have to change this supposedly general module every time a new index implementation is added. On the other hand, would there be any *benefit* from moving the class instantiations to `Database.LSMTree.Internal.Index`?

src/Database/LSMTree/Internal/Index.hs
```diff
@@ -0,0 +1,90 @@
...
+
+{-|
+    The class of index accumulator types, where an index accumulator denotes an
+    index under incremental construction.
+
+    Incremental index construction is only guaranteed to work correctly when the
+    following conditions are met:
+
+      * The provided information about pages is not in conflict with the
+        requirement that the keys in a run are ascending.
```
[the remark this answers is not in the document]
jorisdral: > I deliberately avoided saying that keys should be appended in ascending order, because we actually don’t append keys but key ranges, where a key range is typically specified by a minimum and a maximum key.

Sorted ranges then?

> Yes, in our concrete index implementations, we store only the minimum or only the maximum keys, but the interface does not mandate this. Incidentally, I think it would be better for the interface to do so, but I fear it’s too late now to change this.

I'm not sure I understand what you mean here. For `search` to work properly, the stored key ranges have to be sorted. The implementation does not check this, but we've been mentioning it explicitly as a requirement for correctness

src/Database/LSMTree/Internal/Index.hs
jorisdral: FYI I just wondered if you had an opinion on the two ways of structuring it, I'm fine with the way it is done now. All the stuff I'm writing below is just for the sake of having an interesting discussion :P

I'd say both approaches are nearly equivalent. The choice is mainly about code structure. The module dependency graph changes based on the approach, but I feel that for the most part it's up to preference which one is picked. 

> After all, type classes are open, and, while we currently have these two particular index types, we or someone else may want to add more in the future. When instantiating the index-related classes in Database.LSMTree.Internal.Index, one would have to change this supposedly general module every time a new index implementation is added. 

You can mix and match: have some instances with the class, some instances with the concrete types that are made an instance of the class. All approaches are fine IMO, the result is the same.

> On the other hand, would there be any benefit from moving the class instantiations to `Database.LSMTree.Internal.Index`?

Probably not, but I also don't see a clear benefit from doing it the other way

lsm-tree.cabal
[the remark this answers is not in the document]
jeltsch: > Does `src/**/*Compact.hs` work in your case?

Yes, it does, in both Vim and Bash (with `globstar`). I don’t have the Z Shell installed, but I’m pretty sure that it works in it as well.

src/Database/LSMTree/Internal/Index.hs
jeltsch: > You can mix and match: have some instances with the class, some instances with the datatype definitions. All approaches are fine IMO, the result is the same.

I like to clearly distinguish between an interface and its implementation. When defining a particular datatype like `Set` along with basic operations on it, the way to go is to put both the interface and the implementation into the same module. With type classes and their instantiations it’s different. Here you separately declare an interface, which you can implement multiple times, even in other packages. In this situation it seems to be good to have one module for the interface (the class declaration) and one module for each implementation (the type declaration and the corresponding instance declaration).

Of course, there is also the possibility of putting an instance declaration into the same module as the class declaration. However, I’d use this approach in a different situation. Say you introduce a new type class that has natural instantiations for existing, basic types, like `Integer` or `[]`. Then you obviously can’t have the corresponding instance declarations in the modules where the types are declared. In this case, it seems natural to me to place these instance declarations into the module where the class declaration is.

jeltsch:
> Keep the implementations for _concrete_  implementations of the index class and keep exporting them from the modules. We don't need the generality of the class in case we are testing/benchmarking concrete index types and their functions. Also, until we make the index selectable, using the concrete functions reduces the risk of missed specialisations.

Would you be okay with relying on specialization instead? I think that from a design perspective it’s better to not repeat APIs (and thus code, albeit only little of it).

Of course, we should have a clear approach for guaranteeing that specialization opportunities aren’t missed. From what I understand, an instance declaration introduces specializations for its method implementations. As a result, we should only have to make sure that the modules that use class methods for specific index or index accumulator types import the modules that declare these types (as these contain the corresponding instance declarations). However, this usually happens anyway. The current situation in this regard is as follows:

* The benchmark and most of the tests that are about concrete index or index accumulator types are implemented in modules of their own. These modules necessarily import the modules with the corresponding type declarations, as they have to construct concrete indexes or index accumulators, either manually or via QuickCheck generators (in which case the type of the values to be generated must be explicitly specified).

* The module `Test.Database.LSMTree.Internal.Lookup`, while not directly about indexes, imports the module `Database.LSMTree.Internal.Index.Compact`, because it has to explicitly specify the index type to be used for test runs.

* The module `Test.Database.LSMTree.Internal.RunAcc` currently doesn’t import a module that declares the index and the index accumulator type it uses. However, this is just because the underlying `Database.LSMTree.Internal.RunAcc` module still uses `IndexCompact` and `IndexCompactAcc` unconditionally. When we change this module to use arbitrary index and index accumulator types, we will have to modify its `new` function such that it allows for the specification of an index accumulator type. The test module for `RunAcc` will then have to explicitly pick an index accumulator type and therefore pull in the module where this type is defined (which in turn pulls in the module that defines the underlying index type).

jeltsch:
> We seem to lose some documentation on concrete implementations for index class functions. Either we should keep some of that documentation, or explicitly link to where the new documentation can be found. And we should double check that we’re not losing any information in the process.

I think that I was very careful to not *lose* any information. My goal was to document things that are common to all types of indexes with the classes only. I can add references to this documentation from the modules about the concrete types wherever it is sensible.

src/Database/LSMTree/Internal/Index.hs
```diff
@@ -0,0 +1,90 @@
...
+        leading or trailing space.
+    -}
+    fromSBS :: ShortByteString -> Either String (NumEntries, i)
+
+{-|
+    The class of index accumulator types, where an index accumulator denotes an
+    index under incremental construction.
+
+    Incremental index construction is only guaranteed to work correctly when the
+    following conditions are met:
```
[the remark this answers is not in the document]
jeltsch: Right. I didn’t move the corresponding documentation to the type class, because this requirement is specific to compact indexes. However, this restriction for compact indexes cannot be ignored when specifying the general interface, because this interface is also implemented for compact indexes. In the case of the requirement that the lengths of key sizes must fit into 16 bits each, I also moved the corresponding documentation to the type class. So, yes, I should do the same also with the requirement you mention here.

src/Database/LSMTree/Internal/Index/Compact.hs
```diff
@@ -675,6 +660,27 @@ checkedBitVec off len ba
...
+instance Index IndexCompact where
+
+    search :: SerialisedKey -> IndexCompact -> PageSpan
+    search = search
+
+    sizeInPages :: IndexCompact -> NumPages
+    sizeInPages = sizeInPages
+
+    -- | The input bytestring must be 64 bit aligned. It is directly used as the
+    -- backing memory for the compact index.
```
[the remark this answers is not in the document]
jeltsch: I’m not even sure what this constraint means. The first component of a serialized index is at index 0 and thus 64-bit-aligned, and where the other components are depends on the format of the index. Or does the above requirement refer to the memory address at which the content of the byte string starts? If yes, why is this 64 bit alignment required? Is it only for performance reasons? In this case, it wouldn’t be necessary to have it as a true requirement for all index types, documented with the type class, but could be mentioned as a recommendation here, in the module about *compact* indexes.

src/Database/LSMTree/Internal/Index/Compact.hs
```diff
@@ -2,23 +2,19 @@
...
-  , sizeInPages
     -- * Non-incremental serialisation
   , toLBS
     -- * Incremental serialisation
     -- $incremental-serialisation
   , headerLBS
   , finalLBS
   , word64VectorToChunk
-    -- * Deserialisation
-  , fromSBS
```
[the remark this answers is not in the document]
jeltsch: Please see [my below general comment on specialization](https://github.com/IntersectMBO/lsm-tree/pull/476#issuecomment-2491953851).

src/Database/LSMTree/Internal/Index/CompactAcc.hs
jeltsch: Please see [my below general comment on specialization](https://github.com/IntersectMBO/lsm-tree/pull/476#issuecomment-2491953851).

test/Test/Database/LSMTree/Internal/Index/Ordinary.hs
jeltsch: That’s right, but see [my below general comment on specialization](https://github.com/IntersectMBO/lsm-tree/pull/476#issuecomment-2491953851).

src/Database/LSMTree/Internal/Index/Ordinary.hs
jeltsch: Please see [my below general comment on specialization](https://github.com/IntersectMBO/lsm-tree/pull/476#issuecomment-2491953851).

src/Database/LSMTree/Internal/Index/OrdinaryAcc.hs
jeltsch: Please see [my below general comment on specialization](https://github.com/IntersectMBO/lsm-tree/pull/476#issuecomment-2491953851).

src-extras/Database/LSMTree/Extras/Index.hs
```diff
@@ -42,31 +41,14 @@ instance NFData Append where
...
+    Adds information about appended pages to an index and outputs newly
+    available chunks.
 
-    Internally, 'append' uses 'IndexCompact.appendSingle' and
-    'IndexCompact.appendMulti', and the usage restrictions of those functions
-    apply also here.
+    See the documentation of the 'IndexAcc' class for constraints to adhere to.
 -}
-append :: Append -> IndexCompactAcc s -> ST s [Chunk]
+append :: IndexAcc j => Append -> j s -> ST s [Chunk]
```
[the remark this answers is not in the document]
jeltsch: Yes, makes sense. Done in 67bc0ca46776d6906caf76e7ed8f9135e2a33827.

src/Database/LSMTree/Internal/Index.hs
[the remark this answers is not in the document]
jeltsch: > I’ll try to change the description to something direct.

Done in e7ef851269b3e37e4d89d8a21217ddb51f05fa26.

src/Database/LSMTree/Internal/Index.hs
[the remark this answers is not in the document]
jeltsch: Hmm, it actually seems that this requirement hadn’t even been documented before. I’ll add it then.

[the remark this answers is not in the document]
jeltsch: The corresponding change is in eeec9da67dd9bbbdad34cb8d01213893becf3ffe.

src/Database/LSMTree/Internal/Index/Compact.hs
jeltsch: I’ve moved this documentation to the class now and improved it along the way. The corresponding changes have been introduced by 559cc181bf7fd7114aba70d3c1604ab9000b8a52. Since 3141593ede80ab3ad5e649bdf0ece9286cd936ae has added a reference to the documentation in the class, the problem you’re mentioning here should be solved.

src/Database/LSMTree/Internal/Index/Compact.hs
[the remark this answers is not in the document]
jeltsch: I’ve dealt with this documentation as part of 559cc181bf7fd7114aba70d3c1604ab9000b8a52 now. I’ve settled on the interpretation that the whole contents of the byte string must be 64-bit-aligned and just specified this alignment as a general requirement, independent of the index type. Is this all okay?

src/Database/LSMTree/Internal/Index/CompactAcc.hs
jeltsch: Done in 3141593ede80ab3ad5e649bdf0ece9286cd936ae.

src/Database/LSMTree/Internal/Index/Ordinary.hs
jeltsch: Done in 3141593ede80ab3ad5e649bdf0ece9286cd936ae.

jeltsch: Not really, since the operations of ordinary indexes are directly implemented in the instance declaration, so that it’s somehow clear that any documentation is to be expected in the class declaration.

src/Database/LSMTree/Internal/Index/OrdinaryAcc.hs
jeltsch: Not really, since the operations of ordinary index accumulators are directly implemented in the instance declaration, so that it’s somehow clear that any documentation is to be expected in the class declaration.

jeltsch:
> I can add references to this documentation from the modules about the concrete types wherever it is sensible.

Done in 3141593ede80ab3ad5e649bdf0ece9286cd936ae.

src/Database/LSMTree/Internal/Index.hs
jeltsch: Originally, I had intended to have `new` as a class method, but then I dropped this idea.

An issue with `new` is that its different implementations do not take the same kind of data via arguments. Yes, the versions for the compact and the ordinary index both have an `Int` argument. However, this argument means different things depending on the index type: for compact indexes it specifies the *maximum* chunk size, for ordinary indexes it specifies the *minimum* chunk size. It is conceivable that in the future another index type is added whose implementation of `new` takes data of an entirely different kind.

Of course, it would be possible to use an associated type to specify what the type of the input to `new` should be. However, this would be unnecessarily complicated, as a polymorphic `new` isn’t really needed. I envision that the type of index and perhaps concrete data to feed to `new` is received as a configuration and then a specific implementation of `new` is invoked depending on this configuration. After an index accumulator has been produced this way, constructing and querying the index will be performed using the class methods.

jeltsch: Note that there seems to be an analogy to object-oriented programming in C++ here: in C++, ordinary methods and destructors can be virtual, allowing dynamic binding, but constructors can’t.

lsm-tree.cabal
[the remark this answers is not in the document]
jorisdral: I think it's okay to assume we can include the reorganisation.

src/Database/LSMTree/Internal/Index/Compact.hs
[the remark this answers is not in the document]
jorisdral: It means that the underlying byte array must be 64-bit aligned, which means the start address must be 64-bit aligned. It's important for performance (so we don't have to do unaligned memory operations). Maybe it's also important for correctness, I'm not sure if tests would fail currently if you use an unaligned bytestring.

In the current class setup, you wouldn't be able to figure out whether to pass an aligned or unaligned bytestring to the `fromSBS` function, without some additional information. It's probably simplest to require that a 64-bit aligned bytestring is passed to `fromSBS` regardless of the index type. The change you made in 559cc18 looks good to me.

Another approach is to add some extra info to the class, either through an additional class function or an associated type class, but I think the current approach of just requiring 64-bit alignment is sufficient.

src/Database/LSMTree/Internal/Index.hs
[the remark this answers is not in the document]
jorisdral: > Done in https://github.com/IntersectMBO/lsm-tree/commit/e7ef851269b3e37e4d89d8a21217ddb51f05fa26.

LGTM

> What I meant is that it would not be necessary for appendSingle to receive a minimum and a maximum key of a page. Instead, it would be enough for it to only receive either a minimum or a maximum key and have indexes of all types store only minimum keys or only maximum keys. Actually, this would also better resemble the idea of a fence pointer index: the index should store fences, which are situated between pages, and a close approximation of a fence is either the maximum key of the preceding page or the minimum key of the succeeding page.

I understand what you mean, though it's not so straightforward for the compact index. The compact index needs to know about the both the minimum and maximum key on each page so that it can decide whether the 64-byte prefix of the minimum key is a proper fence. If the maximum key on the previous page and the minimum key on the next page have the same 64-byte prefix, then the minimum key is not a proper fence. This is also documented in the `IndexCompact`:

https://github.com/IntersectMBO/lsm-tree/blob/11738bdd7b352d32e43cd8e45d93fde391c3b91d/src/Database/LSMTree/Internal/IndexCompact.hs#L152-L179

src/Database/LSMTree/Internal/Index.hs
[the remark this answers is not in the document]
jorisdral: It was documented in `format-run.md`. https://github.com/IntersectMBO/lsm-tree/blob/11738bdd7b352d32e43cd8e45d93fde391c3b91d/doc/format-run.md?plain=1#L201

It's not explicitly mentioned in the code though, so it's good that we're mentioning it in the class documentation now

src/Database/LSMTree/Internal/Index.hs
jorisdral: All right, I suppose we'll see in a later PR how `new` is used

src/Database/LSMTree/Internal/Index/Ordinary.hs
jorisdral: If the module were to expose the concrete implementations, then the comments do apply

src/Database/LSMTree/Internal/Index/OrdinaryAcc.hs
jorisdral: https://github.com/IntersectMBO/lsm-tree/pull/476#discussion_r1853735978

jeltsch:
> Maybe as a middle ground, we could expose the class or the class functions from the concrete implementation modules. It also means you get the the full API if you import a concrete implementation module, and then you don’t have to import the general `Index` module as well.

Interesting idea.

Let me summarize the possibilities of what a module providing a concrete index type can export in terms of the general index operations:

1. Versions of the class methods that work for the concrete index type
2. The (polymorphic) class methods themselves (via re-export)
3. Nothing at all

Option 1 increases the risk of name clashes, which have to be resolved then, because there is a single name for a class method and its concrete versions. Option 2 makes the interfaces of the modules providing concrete index types self-contained, so that the general index module doesn’t need to be imported when working with just a concrete index type. Option 3 clearly separates the general interface from its concrete implementations.

I’m generally not a big fan of publicly visible re-exports. They mean that a single item is available from different modules, which not only makes the structure of the overall library interface more complicated but introduces the necessity for the library user to decide from which module to import a particular item.

That said, maybe option 2 is still an acceptable compromise in this particular situation. I think that I’ll go for it.

jeltsch:
> That said, maybe option 2 is still an acceptable compromise in this particular situation. I think that I’ll go for it.

I’ve made corresponding changes in 5b463fc18919b52db1a55884e86659de3b61fee9. I’ve also adapted imports of the modules that provide index support, but only imports into tests and benchmarks, not into other parts of the core library, because the latter are going to be made independent of index types anyhow.

PR #478, opened 2024-11-22
Persist write buffer when creating a snapshot

wenkokke:
This is a WIP PR that serialises the write buffer when creating a snapshot.

I'd love some feedback :)

### DIFF
--- a/src/Database/LSMTree/Internal.hs
+++ b/src/Database/LSMTree/Internal.hs
@@ -102,7 +102,7 @@ import           Database.LSMTree.Internal.BlobRef (WeakBlobRef (..))
 import qualified Database.LSMTree.Internal.BlobRef as BlobRef
 import           Database.LSMTree.Internal.Config
 import qualified Database.LSMTree.Internal.Cursor as Cursor
-import           Database.LSMTree.Internal.Entry (Entry, unNumEntries)
+import           Database.LSMTree.Internal.Entry (Entry)
 import           Database.LSMTree.Internal.Lookup (ByteCountDiscrepancy,
                      ResolveSerialisedValue, lookupsIO)
 import           Database.LSMTree.Internal.MergeSchedule
@@ -126,6 +126,7 @@ import qualified System.FS.API as FS
 import           System.FS.API (FsError, FsErrorPath (..), FsPath, HasFS)
 import qualified System.FS.BlockIO.API as FS
 import           System.FS.BlockIO.API (HasBlockIO)
+import Database.LSMTree.Internal.WriteBufferReader (readWriteBuffer)
 
 {-------------------------------------------------------------------------------
   Existentials
@@ -1144,7 +1145,6 @@ createSnapshot ::
   -> m ()
 createSnapshot resolve snap label tableType t = do
     traceWith (tableTracer t) $ TraceSnapshot snap
-    let conf = tableConfig t
     withOpenTable t $ \thEnv ->
       withTempRegistry $ \reg -> do -- TODO: use the temp registry for all side effects
         let hfs = tableHasFS thEnv
@@ -1163,50 +1163,21 @@ createSnapshot resolve snap label tableType t = do
             (FS.createDirectory hfs (Paths.getNamedSnapshotDir snapDir))
             (\_ -> FS.removeDirectoryRecursive hfs (Paths.getNamedSnapshotDir snapDir))
 
-        -- Write the write buffer.
-        --
-        -- TODO: Seeing as the write buffer is pure, we probably don't need to keep the read lock lock while we're writing to disk.
-        RW.withReadAccess (tableContent thEnv) $ \content -> do
-          writeBufferRunNumber <- uniqueToRunNumber <$> incrUniqCounter (tableSessionUniqCounter thEnv)
-          let fsPaths = Paths.WriteBufferFsPaths (Paths.getNamedSnapshotDir snapDir) writeBufferRunNumber
-          WBW.writeWriteBuffer hfs hbio fsPaths (tableWriteBuffer content) (tableWriteBufferBlobs content)
+        -- Get the table content.
+        content <- RW.withReadAccess (tableContent thEnv) pure
 
-        -- For the temporary implementation it is okay to just flush the buffer
-        -- before taking the snapshot.
-        content <- modifyWithTempRegistry
-                      (RW.unsafeAcquireWriteAccess (tableContent thEnv))
-                      (atomically . RW.unsafeReleaseWriteAccess (tableContent thEnv))
-                      $ \innerReg content -> do
-          -- TODO: When we flush the buffer here, it might be underfull, which
-          -- could mess up the scheduling. The conservative approach is to supply
-          -- credits as if the buffer was full, and then flush the (possibly)
-          -- underfull buffer. However, note that this bit of code
-          -- here is probably going to change anyway because of #392
-          let credits = case confWriteBufferAlloc conf of
-                AllocNumEntries n -> Credits (unNumEntries n)
-          supplyCredits conf credits (tableLevels content)
-          content' <- flushWriteBuffer
-                (TraceMerge `contramap` tableTracer t)
-                conf
-                resolve
-                hfs
-                (tableHasBlockIO thEnv)
-                (tableSessionRoot thEnv)
-                (tableSessionUniqCounter thEnv)
-                innerReg
-                content
-          pure (content', content')
-        -- At this point, we've flushed the write buffer but we haven't created the
-        -- snapshot file yet. If an asynchronous exception happens beyond this
-        -- point, we'll take that loss, as the inner state of the table is still
-        -- consistent.
+        -- Write the write buffer.
+        snapWriteBuffer <- uniqueToRunNumber <$> incrUniqCounter (tableSessionUniqCounter thEnv)
+        let wbPaths = Paths.WriteBufferFsPaths (Paths.getNamedSnapshotDir snapDir) snapWriteBuffer
+        WBW.writeWriteBuffer hfs hbio wbPaths (tableWriteBuffer content) (tableWriteBufferBlobs content)
 
         -- Convert to snapshot format
         snapLevels <- toSnapLevels (tableLevels content)
+
         -- Hard link runs into the named snapshot directory
         snapLevels' <- snapshotRuns reg hbio snapDir snapLevels
 
-        let snapMetaData = SnapshotMetaData label tableType (tableConfig t) snapLevels'
+        let snapMetaData = SnapshotMetaData label tableType (tableConfig t) snapWriteBuffer snapLevels'
             SnapshotMetaDataFile contentPath = Paths.snapshotMetaDataFile snapDir
             SnapshotMetaDataChecksumFile checksumPath = Paths.snapshotMetaDataChecksumFile snapDir
         writeFileSnapshotMetaData hfs contentPath checksumPath snapMetaData
@@ -1247,7 +1218,7 @@ openSnapshot sesh label tableType override snap resolve = do
           Left e  -> throwIO (ErrSnapshotDeserialiseFailure e snap)
           Right x -> pure x
 
-        let SnapshotMetaData label' tableType' conf snapLevels = snapMetaData
+        let SnapshotMetaData label' tableType' conf snapWriteBuffer snapLevels = snapMetaData
 
         unless (tableType == tableType') $
           throwIO (ErrSnapshotWrongTableType snap tableType tableType')
@@ -1264,15 +1235,20 @@ openSnapshot sesh label tableType override snap resolve = do
 
         let actDir = Paths.activeDir (sessionRoot seshEnv)
 
+        -- Read write buffer
+        let wbPaths = Paths.WriteBufferFsPaths (Paths.getNamedSnapshotDir snapDir) snapWriteBuffer
+        tableWriteBuffer <- readWriteBuffer hfs hbio resolve tableWriteBufferBlobs wbPaths
+
         -- Hard link runs into the active directory,
         snapLevels' <- openRuns reg hfs hbio conf (sessionUniqCounter seshEnv) snapDir actDir snapLevels
+        
         -- Convert from the snapshot format, restoring merge progress in the process
         tableLevels <- fromSnapLevels reg hfs hbio conf (sessionUniqCounter seshEnv) resolve actDir snapLevels'
         releaseRuns reg snapLevels'
 
         tableCache <- mkLevelsCache reg tableLevels
         newWith reg sesh seshEnv conf' am $! TableContent {
-            tableWriteBuffer = WB.empty
+            tableWriteBuffer
           , tableWriteBufferBlobs
           , tableLevels
           , tableCache
--- a/src/Database/LSMTree/Internal.hs
+++ b/src/Database/LSMTree/Internal.hs
@@ -1164,6 +1164,8 @@ createSnapshot resolve snap label tableType t = do
             (\_ -> FS.removeDirectoryRecursive hfs (Paths.getNamedSnapshotDir snapDir))
 
         -- Write the write buffer.
+        --
+        -- TODO: Seeing as the write buffer is pure, we probably don't need to keep the read lock lock while we're writing to disk.
         RW.withReadAccess (tableContent thEnv) $ \content -> do
           writeBufferRunNumber <- uniqueToRunNumber <$> incrUniqCounter (tableSessionUniqCounter thEnv)
           let fsPaths = Paths.WriteBufferFsPaths (Paths.getNamedSnapshotDir snapDir) writeBufferRunNumber
--- a/src/Database/LSMTree/Internal.hs
+++ b/src/Database/LSMTree/Internal.hs
@@ -121,6 +121,7 @@ import           Database.LSMTree.Internal.Snapshot.Codec
 import           Database.LSMTree.Internal.UniqCounter
 import qualified Database.LSMTree.Internal.WriteBuffer as WB
 import qualified Database.LSMTree.Internal.WriteBufferBlobs as WBB
+import qualified Database.LSMTree.Internal.WriteBufferWriter as WBW
 import qualified System.FS.API as FS
 import           System.FS.API (FsError, FsErrorPath (..), FsPath, HasFS)
 import qualified System.FS.BlockIO.API as FS
@@ -1161,6 +1162,12 @@ createSnapshot resolve snap label tableType t = do
           allocateTemp reg
             (FS.createDirectory hfs (Paths.getNamedSnapshotDir snapDir))
             (\_ -> FS.removeDirectoryRecursive hfs (Paths.getNamedSnapshotDir snapDir))
+
+        -- Write the write buffer.
+        RW.withReadAccess (tableContent thEnv) $ \content -> do
+          writeBufferRunNumber <- uniqueToRunNumber <$> incrUniqCounter (tableSessionUniqCounter thEnv)
+          let fsPaths = Paths.WriteBufferFsPaths (Paths.getNamedSnapshotDir snapDir) writeBufferRunNumber
+          WBW.writeWriteBuffer hfs hbio fsPaths (tableWriteBuffer content) (tableWriteBufferBlobs content)
 
         -- For the temporary implementation it is okay to just flush the buffer
         -- before taking the snapshot.
--- a/src/Database/LSMTree/Internal.hs
+++ b/src/Database/LSMTree/Internal.hs
@@ -1160,8 +1160,10 @@ createSnapshot snap label tableType t = do
             (FS.createDirectory hfs (Paths.getNamedSnapshotDir snapDir))
             (\_ -> FS.removeDirectoryRecursive hfs (Paths.getNamedSnapshotDir snapDir))
 
-        -- Get the table content.
-        content <- RW.withReadAccess (tableContent thEnv) pure
+        -- Duplicate references to the table content, so that resources do not disappear
+        -- from under our feet while taking a snapshot. These references are released
+        -- again after the snapshot files/directories are written.
+        content <- RW.withReadAccess (tableContent thEnv) (duplicateTableContent reg)
 
         -- Snapshot the write buffer.
         let activeDir = Paths.activeDir (tableSessionRoot thEnv)
@@ -1179,6 +1181,9 @@ createSnapshot snap label tableType t = do
             SnapshotMetaDataFile contentPath = Paths.snapshotMetaDataFile snapDir
             SnapshotMetaDataChecksumFile checksumPath = Paths.snapshotMetaDataChecksumFile snapDir
         writeFileSnapshotMetaData hfs contentPath checksumPath snapMetaData
+
+        -- Release the table content
+        releaseTableContent reg content
 
 {-# SPECIALISE openSnapshot ::
      Session IO h
@@ -1231,10 +1236,8 @@ openSnapshot sesh label tableType override snap resolve = do
         let activeDir = Paths.activeDir (sessionRoot seshEnv)
 
         -- Read write buffer
-        activeWriteBufferNumber <- uniqueToRunNumber <$> incrUniqCounter uc
-        let activeWriteBufferPaths = Paths.WriteBufferFsPaths (Paths.getActiveDir activeDir) activeWriteBufferNumber
         let snapWriteBufferPaths = Paths.WriteBufferFsPaths (Paths.getNamedSnapshotDir snapDir) snapWriteBuffer
-        (tableWriteBuffer, tableWriteBufferBlobs) <- openWriteBuffer reg resolve hfs hbio snapWriteBufferPaths activeWriteBufferPaths
+        (tableWriteBuffer, tableWriteBufferBlobs) <- openWriteBuffer reg resolve hfs hbio uc activeDir snapWriteBufferPaths
 
         -- Hard link runs into the active directory,
         snapLevels' <- openRuns reg hfs hbio conf (sessionUniqCounter seshEnv) snapDir activeDir snapLevels
--- a/src/Database/LSMTree/Internal.hs
+++ b/src/Database/LSMTree/Internal.hs
@@ -121,12 +121,10 @@ import           Database.LSMTree.Internal.Snapshot.Codec
 import           Database.LSMTree.Internal.UniqCounter
 import qualified Database.LSMTree.Internal.WriteBuffer as WB
 import qualified Database.LSMTree.Internal.WriteBufferBlobs as WBB
-import qualified Database.LSMTree.Internal.WriteBufferWriter as WBW
 import qualified System.FS.API as FS
 import           System.FS.API (FsError, FsErrorPath (..), FsPath, HasFS)
 import qualified System.FS.BlockIO.API as FS
 import           System.FS.BlockIO.API (HasBlockIO)
-import Database.LSMTree.Internal.WriteBufferReader (readWriteBuffer)
 
 {-------------------------------------------------------------------------------
   Existentials
@@ -1128,8 +1126,7 @@ readCursorWhile resolve keyIsWanted n Cursor {..} fromEntry = do
 -------------------------------------------------------------------------------}
 
 {-# SPECIALISE createSnapshot ::
-     ResolveSerialisedValue
-  -> SnapshotName
+     SnapshotName
   -> SnapshotLabel
   -> SnapshotTableType
   -> Table IO h
@@ -1137,18 +1134,18 @@ readCursorWhile resolve keyIsWanted n Cursor {..} fromEntry = do
 -- |  See 'Database.LSMTree.Normal.createSnapshot''.
 createSnapshot ::
      (MonadMask m, MonadMVar m, MonadST m, MonadSTM m)
-  => ResolveSerialisedValue
-  -> SnapshotName
+  => SnapshotName
   -> SnapshotLabel
   -> SnapshotTableType
   -> Table m h
   -> m ()
-createSnapshot resolve snap label tableType t = do
+createSnapshot snap label tableType t = do
     traceWith (tableTracer t) $ TraceSnapshot snap
     withOpenTable t $ \thEnv ->
       withTempRegistry $ \reg -> do -- TODO: use the temp registry for all side effects
-        let hfs = tableHasFS thEnv
+        let hfs  = tableHasFS thEnv
             hbio = tableHasBlockIO thEnv
+            uc   = tableSessionUniqCounter thEnv
 
         -- Guard that the snapshot does not exist already
         let snapDir = Paths.namedSnapshotDir (tableSessionRoot thEnv) snap
@@ -1166,10 +1163,11 @@ createSnapshot resolve snap label tableType t = do
         -- Get the table content.
         content <- RW.withReadAccess (tableContent thEnv) pure
 
-        -- Write the write buffer.
-        snapWriteBuffer <- uniqueToRunNumber <$> incrUniqCounter (tableSessionUniqCounter thEnv)
-        let wbPaths = Paths.WriteBufferFsPaths (Paths.getNamedSnapshotDir snapDir) snapWriteBuffer
-        WBW.writeWriteBuffer hfs hbio wbPaths (tableWriteBuffer content) (tableWriteBufferBlobs content)
+        -- Snapshot the write buffer.
+        let activeDir = Paths.activeDir (tableSessionRoot thEnv)
+        let wb = tableWriteBuffer content
+        let wbb = tableWriteBufferBlobs content
+        snapWriteBufferNumber <- Paths.writeBufferNumber <$> snapshotWriteBuffer reg hfs hbio uc activeDir snapDir wb wbb
 
         -- Convert to snapshot format
         snapLevels <- toSnapLevels (tableLevels content)
@@ -1177,7 +1175,7 @@ createSnapshot resolve snap label tableType t = do
         -- Hard link runs into the named snapshot directory
         snapLevels' <- snapshotRuns reg hbio snapDir snapLevels
 
-        let snapMetaData = SnapshotMetaData label tableType (tableConfig t) snapWriteBuffer snapLevels'
+        let snapMetaData = SnapshotMetaData label tableType (tableConfig t) snapWriteBufferNumber snapLevels'
             SnapshotMetaDataFile contentPath = Paths.snapshotMetaDataFile snapDir
             SnapshotMetaDataChecksumFile checksumPath = Paths.snapshotMetaDataChecksumFile snapDir
         writeFileSnapshotMetaData hfs contentPath checksumPath snapMetaData
@@ -1206,6 +1204,7 @@ openSnapshot sesh label tableType override snap resolve = do
       withTempRegistry $ \reg -> do
         let hfs     = sessionHasFS seshEnv
             hbio    = sessionHasBlockIO seshEnv
+            uc      = sessionUniqCounter seshEnv
 
         -- Guard that the snapshot exists
         let snapDir = Paths.namedSnapshotDir (sessionRoot seshEnv) snap
@@ -1228,22 +1227,20 @@ openSnapshot sesh label tableType override snap resolve = do
 
         let conf' = applyOverride override conf
         am <- newArenaManager
-        blobpath <- Paths.tableBlobPath (sessionRoot seshEnv) <$>
-                      incrUniqCounter (sessionUniqCounter seshEnv)
-        tableWriteBufferBlobs <- allocateTemp reg (WBB.new hfs blobpath)
-                                                  releaseRef
 
-        let actDir = Paths.activeDir (sessionRoot seshEnv)
+        let activeDir = Paths.activeDir (sessionRoot seshEnv)
 
         -- Read write buffer
-        let wbPaths = Paths.WriteBufferFsPaths (Paths.getNamedSnapshotDir snapDir) snapWriteBuffer
-        tableWriteBuffer <- readWriteBuffer hfs hbio resolve tableWriteBufferBlobs wbPaths
+        activeWriteBufferNumber <- uniqueToRunNumber <$> incrUniqCounter uc
+        let activeWriteBufferPaths = Paths.WriteBufferFsPaths (Paths.getActiveDir activeDir) activeWriteBufferNumber
+        let snapWriteBufferPaths = Paths.WriteBufferFsPaths (Paths.getNamedSnapshotDir snapDir) snapWriteBuffer
+        (tableWriteBuffer, tableWriteBufferBlobs) <- openWriteBuffer reg resolve hfs hbio snapWriteBufferPaths activeWriteBufferPaths
 
         -- Hard link runs into the active directory,
-        snapLevels' <- openRuns reg hfs hbio conf (sessionUniqCounter seshEnv) snapDir actDir snapLevels
-        
+        snapLevels' <- openRuns reg hfs hbio conf (sessionUniqCounter seshEnv) snapDir activeDir snapLevels
+
         -- Convert from the snapshot format, restoring merge progress in the process
-        tableLevels <- fromSnapLevels reg hfs hbio conf (sessionUniqCounter seshEnv) resolve actDir snapLevels'
+        tableLevels <- fromSnapLevels reg hfs hbio conf (sessionUniqCounter seshEnv) resolve activeDir snapLevels'
         releaseRuns reg snapLevels'
 
         tableCache <- mkLevelsCache reg tableLevels
--- a/src/Database/LSMTree/Internal/Entry.hs
+++ b/src/Database/LSMTree/Internal/Entry.hs
@@ -3,6 +3,7 @@ module Database.LSMTree.Internal.Entry (
   , hasBlob
   , onValue
   , onBlobRef
+  , traverseBlobRef
   , NumEntries (..)
   , unNumEntries
     -- * Value resolution/merging
@@ -46,6 +47,17 @@ onBlobRef def g = \case
     InsertWithBlob _ br -> g br
     Mupdate{}           -> def
     Delete              -> def
+
+traverseBlobRef ::
+     Applicative t
+  => (blobref -> t blobref')
+  -> Entry v blobref
+  -> t (Entry v blobref')
+traverseBlobRef f = \case
+  Insert v -> pure (Insert v)
+  InsertWithBlob v blobref -> InsertWithBlob v <$> f blobref
+  Mupdate v -> pure (Mupdate v)
+  Delete -> pure Delete
 
 instance Bifunctor Entry where
   first f = \case
--- a/src/Database/LSMTree/Internal/Entry.hs
+++ b/src/Database/LSMTree/Internal/Entry.hs
@@ -3,7 +3,6 @@ module Database.LSMTree.Internal.Entry (
   , hasBlob
   , onValue
   , onBlobRef
-  , traverseBlobRef
   , NumEntries (..)
   , unNumEntries
     -- * Value resolution/merging
@@ -47,17 +46,6 @@ onBlobRef def g = \case
     InsertWithBlob _ br -> g br
     Mupdate{}           -> def
     Delete              -> def
-
-traverseBlobRef ::
-     Applicative t
-  => (blobref -> t blobref')
-  -> Entry v blobref
-  -> t (Entry v blobref')
-traverseBlobRef f = \case
-  Insert v -> pure (Insert v)
-  InsertWithBlob v blobref -> InsertWithBlob v <$> f blobref
-  Mupdate v -> pure (Mupdate v)
-  Delete -> pure Delete
 
 instance Bifunctor Entry where
   first f = \case
--- a/src/Database/LSMTree/Internal/Paths.hs
+++ b/src/Database/LSMTree/Internal/Paths.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE PatternSynonyms #-}
 module Database.LSMTree.Internal.Paths (
     SessionRoot (..)
   , lockFile
@@ -24,10 +25,14 @@ module Database.LSMTree.Internal.Paths (
   , runFilterPath
   , runIndexPath
   , runChecksumsPath
-    -- * Checksums
+    -- * Checksums for Run files
   , checksumFileNamesForRunFiles
   , toChecksumsFile
   , fromChecksumsFile
+    -- * Checksums for WriteBuffer files
+  , checksumFileNamesForWriteBufferFiles
+  , toChecksumsFileForWriteBufferFiles
+  , fromChecksumsFileForWriteBufferFiles
     -- * ForRunFiles abstraction
   , ForKOps (..)
   , ForBlob (..)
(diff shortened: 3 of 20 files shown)


src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,227 @@
...
+  -> WriteBufferFsPaths
+  -> WriteBuffer
+  -> WriteBufferBlobs m h
+  -> m ()
+writeWriteBuffer hfs hbio fsPaths buffer blobs = do
+  writer <- new hfs hbio fsPaths
+  for_ (WB.toList buffer) $ \(key, op) ->
+    -- TODO: The fmap entry here reallocates even when there are no blobs.
+    addKeyOp writer key (WBB.mkRawBlobRef blobs <$> op)
+  void $ unsafeFinalise True writer
```
jorisdral, line 60: This should get a `SPECIALISE` pragma for `IO`

src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,227 @@
...
+
+-- data SerialisedWriteBuffer m h = SerialisedWriteBuffer
+--     { -- | The file system paths for all the files used by the serialised write buffer.
+--       serialisedWriteBufferFsPaths    :: !WriteBufferFsPaths
+--       -- | The (read mode) file handles.
+--     , serialisedWriteBufferHandles    :: !(ForWriteBufferFiles (FS.Handle h))
+--     , serialisedWriteBufferHasFS      :: !(HasFS m h)
+--     , serialisedWriteBufferHasBlockIO :: !(HasBlockIO m h)
+--     }
```
jorisdral, line 69:
```suggestion
```

src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,227 @@
...
+
+data WriteBufferWriter m h = WriteBufferWriter
```
jorisdral, line 71: Give this some haddocks. Maybe also hint at the fact that the `WriteBufferWriter` is largely based on `RunBuilder`, and that is why we link its documenation some places.

src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,227 @@
...
+    { writeBufferWriterFsPaths    = fsPaths,
+      writeBufferWriterHasFS      = hfs,
+      writeBufferWriterHasBlockIO = hbio,
+      ..
+    }
+
+-- | See 'RunBuilder.unsafeFinalise'.
+-- Finalise an incremental 'WriteBufferWriter'.
+-- Do /not/ use a 'WriteBufferWriter' after finalising it.
+unsafeFinalise ::
```
jorisdral, line 111: Put newlines in between if the intent is to separate the text into paragraphs

src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,227 @@
...
+  FS.withFile writeBufferWriterHasFS (writeBufferChecksumsPath writeBufferWriterFsPaths) (FS.WriteMode FS.MustBeNew) $ \h -> do
+    CRC.writeChecksumsFile' writeBufferWriterHasFS h checksums
+    FS.hDropCacheAll writeBufferWriterHasBlockIO h
+  -- drop the KOps and blobs files from the cache if asked for
+  when dropCaches $ do
+    dropCache writeBufferWriterHasBlockIO (forWriteBufferKOpsRaw writeBufferWriterHandles)
+    dropCache writeBufferWriterHasBlockIO (forWriteBufferBlobRaw writeBufferWriterHandles)
+  for_ writeBufferWriterHandles $ closeHandle writeBufferWriterHasFS
+  return (writeBufferWriterHasFS, writeBufferWriterHasBlockIO, writeBufferWriterFsPaths)
+
```
jorisdral, line 130:
```suggestion
```

src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,227 @@
...
+--   -> [RawOverflowPage]
+--   -> m ()
+-- addLargeSerialisedKeyOp WriteBufferWriter{..} page overflowPages =
+--   assert (RawPage.rawPageNumKeys page == 1) $
+--   assert (RawPage.rawPageHasBlobSpanAt page 0 == 0) $
+--   assert (RawPage.rawPageOverflowPages page > 0) $
+--   assert (RawPage.rawPageOverflowPages page == length overflowPages) $ do
+--     !pages <- ST.stToIO $ selectPages <$> flushPageIfNonEmpty writeBufferWriterPageAcc <*> pure page
+--     for_ pages $ writeRawPage writeBufferWriterHasFS (forWriteBufferKOps writeBufferWriterHandles)
+--     writeRawOverflowPages writeBufferWriterHasFS (forWriteBufferKOps writeBufferWriterHandles) overflowPages
```
jorisdral, line 164: If it is not used, we can remove the code

```suggestion
```

src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,227 @@
...
+  writeBufferWriterHandles <-
+    traverse (makeHandle hfs) (pathsForWriteBufferFiles fsPaths)
+  return WriteBufferWriter
+    { writeBufferWriterFsPaths    = fsPaths,
+      writeBufferWriterHasFS      = hfs,
+      writeBufferWriterHasBlockIO = hbio,
+      ..
+    }
+
+-- | See 'RunBuilder.unsafeFinalise'.
```
jorisdral, line 108: `RunBuilder` is not imported, so you'll have to use fully qualified names, like `Database.LSMTree.Internal.RunBuilder.unsafeFinalise`

src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,227 @@
...
+    FS.hDropCacheAll writeBufferWriterHasBlockIO h
+  -- drop the KOps and blobs files from the cache if asked for
+  when dropCaches $ do
+    dropCache writeBufferWriterHasBlockIO (forWriteBufferKOpsRaw writeBufferWriterHandles)
+    dropCache writeBufferWriterHasBlockIO (forWriteBufferBlobRaw writeBufferWriterHandles)
+  for_ writeBufferWriterHandles $ closeHandle writeBufferWriterHasFS
+  return (writeBufferWriterHasFS, writeBufferWriterHasBlockIO, writeBufferWriterFsPaths)
+
+
+-- | See 'RunBuilder.addKeyOp'.
```
jorisdral, line 132: Use fully a fully qualified name for `addKeyOp`

src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,227 @@
...
+-- addLargeSerialisedKeyOp WriteBufferWriter{..} page overflowPages =
+--   assert (RawPage.rawPageNumKeys page == 1) $
+--   assert (RawPage.rawPageHasBlobSpanAt page 0 == 0) $
+--   assert (RawPage.rawPageOverflowPages page > 0) $
+--   assert (RawPage.rawPageOverflowPages page == length overflowPages) $ do
+--     !pages <- ST.stToIO $ selectPages <$> flushPageIfNonEmpty writeBufferWriterPageAcc <*> pure page
+--     for_ pages $ writeRawPage writeBufferWriterHasFS (forWriteBufferKOps writeBufferWriterHandles)
+--     writeRawOverflowPages writeBufferWriterHasFS (forWriteBufferKOps writeBufferWriterHandles) overflowPages
+
+-- | See 'RunAcc.addSmallKeyOp'.
```
jorisdral, line 166: Use fully a fully qualified name for `addSmallKeyOp`

src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,227 @@
...
+        not <$> PageAcc.pageAccAddElem pageAcc key op
+    if pageBoundaryNeeded
+      then do
+        mPage <- flushPageIfNonEmpty pageAcc
+        added <- PageAcc.pageAccAddElem pageAcc key op
+        assert added $ pure mPage
+      else do
+        pure Nothing
+
+-- | See 'RunAcc.addLargeKeyOp'.
```
jorisdral, line 186: Use fully a fully qualified name for `addLargeKeyOp`

src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,227 @@
...
+    -- new large key/op will need more than one page to itself.
+    mPagePre <- flushPageIfNonEmpty pageAcc
+    -- Make the new page and overflow pages. Add the span of pages to the index.
+    let (page, overflowPages) = PageAcc.singletonPage key op
+
+    -- Combine the results with anything we flushed before
+    let !pages = selectPages mPagePre page
+    return (pages, overflowPages)
+
+-- | Internal helper. See 'RunAcc.flushPageIfNonEmpty'.
```
jorisdral, line 204: Use a fully qualified name for `flushPageIfNonEmpty`

src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,227 @@
...
+    keysCount <- PageAcc.keysCountPageAcc pageAcc
+    if keysCount > 0
+      then do
+        -- Serialise the page and reset the accumulator
+        page <- PageAcc.serialisePageAcc pageAcc
+        PageAcc.resetPageAcc pageAcc
+        pure $ Just page
+      else pure Nothing
+
+-- | Internal helper. See 'RunAcc.selectPagesAndChunks'.
```
jorisdral, line 216: Use a fully qualified name for `selectPages`

src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,227 @@
...
+                     -> RawPage
+                     -> [RawPage]
+selectPages mPagePre page =
+  maybeToList mPagePre ++ [page]
+
+-- -- | See 'RunBuilder.close'.
+-- close :: MonadSTM m => WriteBufferWriter m h -> m ()
+-- close WriteBufferWriter {..} = do
+--     for_ writeBufferWriterHandles $ closeHandle writeBufferWriterHasFS
+--     for_ (pathsForWriteBufferFiles writeBufferWriterFsPaths) $ FS.removeFile writeBufferWriterHasFS
```
jorisdral, line 227: We can remove this code if it is unused

src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,227 @@
...
+  assert (PageAcc.entryWouldFitInPage key op) $ do
+    pageBoundaryNeeded <-
+        -- Try adding the key/op to the page accumulator to see if it fits. If
+        -- it does not fit, a page boundary is needed.
+        not <$> PageAcc.pageAccAddElem pageAcc key op
+    if pageBoundaryNeeded
+      then do
+        mPage <- flushPageIfNonEmpty pageAcc
+        added <- PageAcc.pageAccAddElem pageAcc key op
+        assert added $ pure mPage
```
jorisdral, line 182: Is the idea to copy documentation from `RunAcc` and `RunBuilder`? If so, we're missing some here

src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,227 @@
+module Database.LSMTree.Internal.WriteBufferWriter
```
jorisdral, line 1: Could add some haddocks here

jorisdral (commented):
Looks promising!

I think this works for writing out the write buffer. It would be good to add a test that exercises `writeWriteBuffer` directly. Maybe we could test that the `WriteBufferWriter` and `RunBuilder` produce the same keyops and blob files?

wenkokke:
> Looks promising!
> 
> 
> 
> I think this works for writing out the write buffer. It would be good to add a test that exercises `writeWriteBuffer` directly. Maybe we could test that the `WriteBufferWriter` and `RunBuilder` produce the same keyops and blob files?

Is there any testing infrastructure that would allow me to easily set up the required inputs?

src/Database/LSMTree/Internal/Paths.hs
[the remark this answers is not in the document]
wenkokke: Fair enough, I just liked the consistency.

src/Database/LSMTree/Internal/WriteBufferWriter.hs
wenkokke: It's largely in there just in case we decide to export the incremental API, in which case it might be needed.

wenkokke: It's largely in there just in case we decide to export the incremental API, in which case it might be needed.

src/Database/LSMTree/Internal/WriteBufferWriter.hs
wenkokke: See previous comment.

wenkokke:
> Looks promising!
> 
> 
> 
> I think this works for writing out the write buffer. It would be good to add a test that exercises `writeWriteBuffer` directly. Maybe we could test that the `WriteBufferWriter` and `RunBuilder` produce the same keyops and blob files?

Do you have any initial feelings regarding the amount of duplication?

jorisdral:
> > Looks promising!
> > I think this works for writing out the write buffer. It would be good to add a test that exercises `writeWriteBuffer` directly. Maybe we could test that the `WriteBufferWriter` and `RunBuilder` produce the same keyops and blob files?
> 
> Is there any testing infrastructure that would allow me to easily set up the required inputs?

https://github.com/IntersectMBO/lsm-tree/blob/d11c709cb3c3b91d9f66e94620017e0edd997528/src-extras/Database/LSMTree/Extras/RunData.hs

^ contains some definitions for generating data for runs, which is equivalent to data for write buffers. It's just sorted key\/value data without duplicate keys

https://github.com/IntersectMBO/lsm-tree/blob/d11c709cb3c3b91d9f66e94620017e0edd997528/test/Test/Database/LSMTree/Internal/Run.hs#L179-L212

^ this test could give you an idea of how to use the run data.

jorisdral:
> > Looks promising!
> > I think this works for writing out the write buffer. It would be good to add a test that exercises `writeWriteBuffer` directly. Maybe we could test that the `WriteBufferWriter` and `RunBuilder` produce the same keyops and blob files?
> 
> Do you have any initial feelings regarding the amount of duplication?

Oops, yes, I meant to write about that but I forgot. I think it's okay to have the duplication for now, as long as we 
* Put down a TODO (maybe in the module header?) to investigate whether we can reduce the amount of duplication. I might have some ideas, but it would mean a larger refactoring effort which I don't think is worth it to do right away
* Establish some correspondence between `RunBuilder` and `WriteBufferWriter`, hence the suggestion to add a test

wenkokke:
> > > Looks promising!
> 
> > > I think this works for writing out the write buffer. It would be good to add a test that exercises `writeWriteBuffer` directly. Maybe we could test that the `WriteBufferWriter` and `RunBuilder` produce the same keyops and blob files?
> 
> > 
> 
> > Do you have any initial feelings regarding the amount of duplication?
> 
> 
> 
> Oops, yes, I meant to write about that but I forgot. I think it's okay to have the duplication for now, as long as we 
> 
> * Put down a TODO (maybe in the module header?) to investigate whether we can reduce the amount of duplication. I might have some ideas, but it would mean a larger refactoring effort which I don't think is worth it to do right away
> 
> * Establish some correspondence between `RunBuilder` and `WriteBufferWriter`, hence the suggestion to add a test

Fair! I still think the generalisation of the run infrastructure to permit empty filters and indices might be worthwhile, especially with SPJ's recent "specialise with value" extension.

jorisdral:
> > > > Looks promising!
> > 
> > 
> > > > I think this works for writing out the write buffer. It would be good to add a test that exercises `writeWriteBuffer` directly. Maybe we could test that the `WriteBufferWriter` and `RunBuilder` produce the same keyops and blob files?
> > 
> > 
> > > 
> > 
> > 
> > > Do you have any initial feelings regarding the amount of duplication?
> > 
> > 
> > Oops, yes, I meant to write about that but I forgot. I think it's okay to have the duplication for now, as long as we
> > 
> > * Put down a TODO (maybe in the module header?) to investigate whether we can reduce the amount of duplication. I might have some ideas, but it would mean a larger refactoring effort which I don't think is worth it to do right away
> > * Establish some correspondence between `RunBuilder` and `WriteBufferWriter`, hence the suggestion to add a test
> 
> Fair! I still think the generalisation of the run infrastructure to permit empty filters and indices might be worthwhile, especially with SPJ's recent "specialise with value" extension.

True, though we wouldn't be able to take advantage of that on older GHCs... unless that stuff is backported

wenkokke:
Added the test `prop_WriteRunEqWriteWriteBuffer` that compares serialisation as Run versus serialisation as WriteBuffer, which fails:
```
lsm-tree
  Database.LSMTree.Internal.Run
    Write buffer to disk
      prop_WriteRunEqWriteWriteBuffer: FAIL (0.01s)
        *** Failed! (after 4 tests and 2 shrinks):
        Exception:
          Assertion failed
            assert, called at src/Database/LSMTree/Internal/RawBytes.hs:157:5 in lsm-tree-0.1.0.0-inplace:Database.LSMTree.Internal.RawBytes
        WB {unWB = fromList [(SerialisedKey [],Delete)]}
        Use --quickcheck-replay="(SMGen 4977141403849015630 6713696729127803493,3)" to reproduce.
```

wenkokke:
I've added an implementation for `readWriteBuffer`. However, my test fails with a double free, which I cannot seem to hunt down.

src/Database/LSMTree/Internal/Paths.hs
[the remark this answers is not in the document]
wenkokke: I've kept it for now, since it served me well. Happy to remove once other rewrites are compete.

wenkokke:
@jorisdral This is ready for review.

wenkokke:
This is ready for review _but not for merge_. There appear to be some non-deterministic test failures.

src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -210,6 +222,74 @@ toSnapMergingRunState (MR.OngoingMerge rs (MR.SpentCreditsVar spentCreditsVar) m
...
+     (MonadMVar m, MonadMask m, MonadSTM m, MonadST m)
+  => TempRegistry m
+  -> ResolveSerialisedValue
+  -> HasFS m h
+  -> HasBlockIO m h
+  -> WriteBufferFsPaths
+  -> WriteBufferFsPaths
+  -> m (WriteBuffer, Ref (WriteBufferBlobs m h))
+openWriteBuffer reg resolve hfs hbio snapWriteBufferPaths activeWriteBufferPaths = do
+  -- Hard link the write buffer keyops and checksum files to the snapshot directory.
```
jorisdral, line 285:
```suggestion
  -- Hard link the write buffer keyops and checksum files to the active directory.
```

src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -210,6 +222,74 @@ toSnapMergingRunState (MR.OngoingMerge rs (MR.SpentCreditsVar spentCreditsVar) m
...
+  hardLinkTemp reg hfs hbio NoHardLinkDurable (writeBufferKOpsPath snapWriteBufferPaths) (writeBufferKOpsPath activeWriteBufferPaths)
+  hardLinkTemp reg hfs hbio NoHardLinkDurable (writeBufferChecksumsPath snapWriteBufferPaths) (writeBufferChecksumsPath activeWriteBufferPaths)
+  -- Copy the write buffer blobs file to the snapshot directory.
```
jorisdral, line 288:
```suggestion
  -- Copy the write buffer blobs file to the active directory.
```

src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -210,6 +222,74 @@ toSnapMergingRunState (MR.OngoingMerge rs (MR.SpentCreditsVar spentCreditsVar) m
...
+  -> ActiveDir
+  -> NamedSnapshotDir
+  -> WriteBuffer
+  -> Ref (WriteBufferBlobs m h)
+  -> m WriteBufferFsPaths
+snapshotWriteBuffer reg hfs hbio uc activeDir snapDir wb wbb = do
+  -- Write the write buffer and write buffer blobs to the active directory.
+  activeWriteBufferNumber <- uniqueToRunNumber <$> incrUniqCounter uc
+  let activeWriteBufferPaths = WriteBufferFsPaths (getActiveDir activeDir) activeWriteBufferNumber
+  WBW.writeWriteBuffer hfs hbio activeWriteBufferPaths wb wbb
```
jorisdral, line 256: In case an exception happens later while within the scope of the current `TempRegistry`, the write buffer files we've written to disk should be removed. So, this should be something along the lines of:

```suggestion
  allocateTemp reg 
    (WBW.writeWriteBuffer hfs hbio activeWriteBufferPaths wb wbb)
    (... {- remove files -} )
```

src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -210,6 +222,74 @@ toSnapMergingRunState (MR.OngoingMerge rs (MR.SpentCreditsVar spentCreditsVar) m
...
+  -> ResolveSerialisedValue
+  -> HasFS m h
+  -> HasBlockIO m h
+  -> WriteBufferFsPaths
+  -> WriteBufferFsPaths
+  -> m (WriteBuffer, Ref (WriteBufferBlobs m h))
+openWriteBuffer reg resolve hfs hbio snapWriteBufferPaths activeWriteBufferPaths = do
+  -- Hard link the write buffer keyops and checksum files to the snapshot directory.
+  hardLinkTemp reg hfs hbio NoHardLinkDurable (writeBufferKOpsPath snapWriteBufferPaths) (writeBufferKOpsPath activeWriteBufferPaths)
+  hardLinkTemp reg hfs hbio NoHardLinkDurable (writeBufferChecksumsPath snapWriteBufferPaths) (writeBufferChecksumsPath activeWriteBufferPaths)
```
jorisdral, line 287: Why not read straight from the files in the snapshot directory?

Moreover, if we hard link them in the active directory and they become unused right after `WBR.readWriteBuffer`, then we have to manually delete the files in the active directory

src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -396,12 +476,66 @@ hardLinkRunFiles ::
...
+    -> HasFS IO h
+    -> HasBlockIO IO h
+    -> HardLinkDurable
+    -> FS.FsPath
+    -> FS.FsPath
+    -> IO ()
+  #-}
+-- | @'hardLinkTemp' reg hfs hbio dur sourcePath targetPath@ creates a hard link
+-- from @sourcePath@ to @targetPath@.
+hardLinkTemp ::
```
jorisdral, line 498: Let's call it just `hardLink` instead. The `Temp` prefix is going to be out of date soon.

src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -396,12 +476,66 @@ hardLinkRunFiles ::
...
+  copyFileTemp ::
+       TempRegistry IO
+    -> HasFS IO h
+    -> HasBlockIO IO h
+    -> FS.FsPath
+    -> FS.FsPath
+    -> IO ()
+    #-}
+-- | @'copyFile' hfs hbio source target@ copies the @source@ path to the @target@ path.
+copyFileTemp ::
```
jorisdral, line 528: Let's call it just `copyFile` instead. The `Temp` prefix is going to be out of date soon.

src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -396,12 +476,66 @@ hardLinkRunFiles ::
...
+{-# SPECIALISE
+  copyFileTemp ::
+       TempRegistry IO
+    -> HasFS IO h
+    -> HasBlockIO IO h
+    -> FS.FsPath
+    -> FS.FsPath
+    -> IO ()
+    #-}
+-- | @'copyFile' hfs hbio source target@ copies the @source@ path to the @target@ path.
```
jorisdral, line 527:
```suggestion
-- | @'copyFile' reg hfs hbio source target@ copies the @source@ path to the @target@ path.
```

test/Test/Database/LSMTree/Internal/Run.hs
```diff
@@ -209,3 +229,98 @@ prop_WriteAndOpen fs hbio wb =
...
+
+-- | Serialise a 'WriteBuffer' and 'WriteBufferBlobs' to disk and perform an 'IO' action.
+withSerialisedWriteBuffer ::
+       FS.HasFS IO h
+    -> FS.HasBlockIO IO h
+    -> WriteBufferFsPaths
+    -> WriteBuffer
+    -> Ref (WriteBufferBlobs IO h)
+    -> IO a
+    -> IO a
```
jorisdral, line 318: Too many leading spaces in the signature

test/Test/Database/LSMTree/Internal/Run.hs
```diff
@@ -209,3 +229,98 @@ prop_WriteAndOpen fs hbio wb =
...
+            rdBlob <- FSL.hGetAll hfs rdBlobHandle
+            wbBlob <- FSL.hGetAll hfs wbBlobHandle
+            assertEqual "writtenRunBlob/writtenWriteBufferBlob" rdBlob wbBlob
+  -- TODO: return a proper Property instead of using assertEqual etc.
+  return (property True)
+
+-- | Use 'SerialisedRunData' to 'WriteBuffer' and 'WriteBufferBlobs'.
+withRunDataAsWriteBuffer ::
+    FS.HasFS IO h
+  -> ResolveSerialisedValue
```
jorisdral, line 297:
```suggestion
withRunDataAsWriteBuffer ::
     FS.HasFS IO h
  -> ResolveSerialisedValue
```

test/Test/Database/LSMTree/Internal/Run.hs
```diff
@@ -209,3 +229,98 @@ prop_WriteAndOpen fs hbio wb =
...
+    -> IO a
+    -> IO a
+withSerialisedWriteBuffer hfs hbio wbPaths wb wbb =
+  bracket_ (writeWriteBuffer hfs hbio wbPaths wb wbb) $ do
+    for_ [ Paths.writeBufferKOpsPath wbPaths
+         , Paths.writeBufferBlobPath wbPaths
+         , Paths.writeBufferChecksumsPath wbPaths
+         ] $ \fsPath -> do
+      fsPathExists <- FS.doesFileExist hfs fsPath
+      when fsPathExists $ FS.removeFile hfs fsPath
```
jorisdral, line 326: This can go into `Database.LSMTree.Extras.RunData`

test/Test/Database/LSMTree/Internal/Run.hs
```diff
@@ -209,3 +229,98 @@ prop_WriteAndOpen fs hbio wb =
 
       -- TODO: return a proper Property instead of using assertEqual etc.
       return (property True)
+
+-- | Writing and loading a 'WriteBuffer' gives the same in-memory
+--   representation as the original run.
```
jorisdral, line 234:
```suggestion
-- | Writing and loading a 'WriteBuffer' gives the same in-memory
--   representation as the original write buffer.
```

test/Test/Database/LSMTree/Internal/Run.hs
```diff
@@ -209,3 +229,98 @@ prop_WriteAndOpen fs hbio wb =
...
+    let inPaths = WrapRunFsPaths $ simplePath 1111
+    let resolve (SerialisedValue x) (SerialisedValue y) = SerialisedValue (x <> y)
+    withRunDataAsWriteBuffer hfs resolve inPaths srd $ \wb wbb -> do
+      -- Write write buffer to disk:
+      let wbPaths = WrapRunFsPaths $ simplePath 1312
+      withSerialisedWriteBuffer hfs hbio wbPaths wb wbb $ do
+        -- Read write buffer from disk:
+        (wb', wbb') <- readWriteBuffer reg resolve hfs hbio wbPaths
+        assertEqual "k/ops" wb wb'
+        releaseRef wbb'
```
jorisdral, line 253: This is not comparing the blobs, only the key/ops. Something similar to `readKOps` as used in `prop_WriteAndOpen` should work, I think?

test/Test/Database/LSMTree/Internal/Run.hs
```diff
@@ -209,3 +229,98 @@ prop_WriteAndOpen fs hbio wb =
...
+          FS.withFile hfs wbKOpsPath FS.ReadMode $ \wbKOpsHandle -> do
+            rdKOps <- FSL.hGetAll hfs rdKOpsHandle
+            wbKOps <- FSL.hGetAll hfs wbKOpsHandle
+            assertEqual "writtenRunKOps/writtenWriteBufferKOps" rdKOps wbKOps
+        -- Compare Blob:
+        FS.withFile hfs rdBlobFile FS.ReadMode $ \rdBlobHandle ->
+          FS.withFile hfs wbBlobPath FS.ReadMode $ \wbBlobHandle -> do
+            rdBlob <- FSL.hGetAll hfs rdBlobHandle
+            wbBlob <- FSL.hGetAll hfs wbBlobHandle
+            assertEqual "writtenRunBlob/writtenWriteBufferBlob" rdBlob wbBlob
```
jorisdral, line 290: This compares only the bytes, but we can probably compare both the keyops and blobs instead/in addition. If there are byte mismatches currently, then the assertion wouldn't create readable counter examples, because it would be thousands of bytes for even a single key/value/blob triplet

jorisdral (changes requested):
I haven't reviewed `WriteBufferReader.hs` and `WriteBufferWriter.hs` yet, reviews for those will follow

src/Database/LSMTree/Internal/WriteBufferReader.hs
```diff
@@ -0,0 +1,155 @@
...
+    (,) <$> readEntries reader <*> pure readerWriteBufferBlobs
+  where
+    readEntries reader = readEntriesAcc WB.empty
+      where
+        readEntriesAcc acc = next reader >>= \case
+          Empty -> pure acc
+          ReadEntry key entry -> readEntriesAcc $ WB.addEntry resolve key (rawBlobRefSpan <$> toFullEntry entry) acc
+
+-- | Allows reading the k\/ops of a run incrementally, using its own read-only
+-- file handle and in-memory cache of the current disk page.
```
jorisdral, line 65: Not a run

src/Database/LSMTree/Internal/WriteBufferReader.hs
```diff
@@ -0,0 +1,155 @@
...
+--
+-- New pages are loaded when trying to read their first entry.
+data WriteBufferReader m h = WriteBufferReader {
+      -- | The disk page currently being read. If it is 'Nothing', the reader
+      -- is considered closed.
+      readerCurrentPage      :: !(MutVar (PrimState m) (Maybe RawPage))
+      -- | The index of the entry to be returned by the next call to 'next'.
+    , readerCurrentEntryNo   :: !(PrimVar (PrimState m) Word16)
+    , readerKOpsHandle       :: !(FS.Handle h)
+    , readerWriteBufferBlobs :: !(Ref (WriteBufferBlobs m h))
```
jorisdral, line 75: Could this be a `BlobFile` instead? That would keep this code closer to the original `RunReader`

src/Database/LSMTree/Internal/WriteBufferReader.hs
```diff
@@ -0,0 +1,155 @@
...
+  => TempRegistry m
+  -> HasFS m h
+  -> HasBlockIO m h
+  -> WriteBufferFsPaths
+  -> m (WriteBufferReader m h)
+new reg readerHasFS readerHasBlockIO fsPaths = do
+  (readerKOpsHandle :: FS.Handle h) <- FS.hOpen readerHasFS (Paths.writeBufferKOpsPath fsPaths) FS.ReadMode
+  -- Double the file readahead window (only applies to this file descriptor)
+  FS.hAdviseAll readerHasBlockIO readerKOpsHandle FS.AdviceSequential
+  readerWriteBufferBlobs <- allocateTemp reg (WBB.open readerHasFS (Paths.writeBufferBlobPath fsPaths) FS.AllowExisting) releaseRef
```
jorisdral, line 100: It would be closer to how the `RunReader` works if we are required to pass in a `BlobFile`/`WriteBufferBlobs` instead of creating it here. It would mean that a `WriteBufferReader` behaves the same as a `RunReader` with regards to references: it duplicates input references, but it does not create them

src/Database/LSMTree/Internal/WriteBufferReader.hs
```diff
@@ -0,0 +1,155 @@
...
+            -- TODO: we know that we need the next page, could already load?
+            modifyPrimVar readerCurrentEntryNo (+1)
+            let entry' :: E.Entry SerialisedValue (RawBlobRef m h)
+                entry' = fmap (WBB.mkRawBlobRef readerWriteBufferBlobs) entry
+            overflowPages <- readOverflowPages readerHasFS readerKOpsHandle lenSuffix
+            let rawEntry = mkEntryOverflow entry' page lenSuffix overflowPages
+            return (ReadEntry key rawEntry)
+
+close :: WriteBufferReader m h -> m ()
+close WriteBufferReader{..} = FS.hClose readerHasFS readerKOpsHandle
```
jorisdral, line 155: `close` should also release the `readerWriteBufferBlobs` reference, if we make the changes I suggested above

src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,236 @@
...
+    writerHasFS      :: !(HasFS m h),
+    writerHasBlockIO :: !(HasBlockIO m h)
+  }
+
+{-# SPECIALISE
+  new ::
+      HasFS IO h
+    -> HasBlockIO IO h
+    -> WriteBufferFsPaths
+    -> IO (WriteBufferWriter IO h)
```
jorisdral, line 91:
```suggestion
  new ::
       HasFS IO h
    -> HasBlockIO IO h
    -> WriteBufferFsPaths
    -> IO (WriteBufferWriter IO h)
```

src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,236 @@
...
+      writerHasFS      = hfs,
+      writerHasBlockIO = hbio,
+      ..
+    }
+
+{-# SPECIALISE
+    unsafeFinalise ::
+        Bool
+      -> WriteBufferWriter IO h
+      -> IO (HasFS IO h, HasBlockIO IO h, WriteBufferFsPaths)
```
jorisdral, line 120:
```suggestion
         Bool
      -> WriteBufferWriter IO h
      -> IO (HasFS IO h, HasBlockIO IO h, WriteBufferFsPaths)
```

src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,236 @@
...
+  for_ writerHandles $ closeHandle writerHasFS
+  return (writerHasFS, writerHasBlockIO, writerFsPaths)
+
+
+{-# SPECIALIZE
+    addKeyOp ::
+        WriteBufferWriter IO h
+      -> SerialisedKey
+      -> Entry SerialisedValue (RawBlobRef IO h)
+      -> IO ()
```
jorisdral, line 155:
```suggestion
         WriteBufferWriter IO h
      -> SerialisedKey
      -> Entry SerialisedValue (RawBlobRef IO h)
      -> IO ()
```

src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,236 @@
...
+        -- Serialise the page and reset the accumulator
+        page <- PageAcc.serialisePageAcc pageAcc
+        PageAcc.resetPageAcc pageAcc
+        pure $ Just page
+      else pure Nothing
+
+-- | Internal helper. See 'Database.LSMTree.Internal.RunAcc.selectPagesAndChunks'.
+selectPages :: Maybe RawPage
+                     -> RawPage
+                     -> [RawPage]
```
jorisdral, line 234:
```suggestion
selectPages :: 
     Maybe RawPage
  -> RawPage
  -> [RawPage]
```

src/Database/LSMTree/Internal/WriteBufferWriter.hs
```diff
@@ -0,0 +1,236 @@
...
+    let (page, overflowPages) = PageAcc.singletonPage key op
+    -- Combine the results with anything we flushed before
+    let !pages = selectPages mPagePre page
+    return (pages, overflowPages)
+
+-- | Internal helper. See 'Database.LSMTree.Internal.RunAcc.flushPageIfNonEmpty'.
+flushPageIfNonEmpty :: PageAcc s -> ST s (Maybe RawPage)
+flushPageIfNonEmpty pageAcc = do
+    keysCount <- PageAcc.keysCountPageAcc pageAcc
+    if keysCount > 0
```
jorisdral, line 223: To keep this code close to its original in `RunAcc`, it would have been better not to rename identifiers, like changing `nkeys` to `keysCount`. It doesn't have to change now, but something to keep in mind going forward

jorisdral (changes requested):
I think this looks okay! There are some things we should change, but after that we can probably merge this PR. 

A general note: it would have been easier to review the PR if the commits had been more structured and informative

I think the next thing to tackle (in follow-up PRs) would be to deduplicate the code we have for write buffer serialisation, because it is in many aspects nearly identical.

jorisdral:
Regarding deduplication, there are two avenues I can think of. 

1. Deduplicating `WriteBufferReader` and `RunReader`. And my thinking is that we could create a more general `KOpsReader` type that can be instantiated to be either a `WriteBufferReader` or `RunReader`. We should probably do this first, because it requires fewer changes than the next avenue

2. Deduplicating `WriteBufferWriter` and `RunBuilder`. My thinking is that we could construct this hierarchy:

```
                      /-------- WriteBufferBuilder
                     /
KOpsAcc <- KOpsBuilder <---------- RunBuilder
                            /
ProbeAcc <- ProbeBuilder <-/
```

`RunAcc` is split up into `KOpsAcc` and `ProbeAcc`. The former takes care of building pages for key/ops and blob files. The latter takes care of building the bloom filters and fence pointer indices. `RunBuilder` is split in two in the same way, doing the `IO` for the `Acc`umulators. The *new* `RunBuilder` is the composition of a `KOPsBuilder` and a `ProbeBuilder`. The `WriteBufferBuilder` is just a small wrapper around a `KOpsBuilder`

src/Database/LSMTree/Internal/Entry.hs
```diff
@@ -47,6 +48,17 @@ onBlobRef def g = \case
...
+     Applicative t
+  => (blobref -> t blobref')
+  -> Entry v blobref
+  -> t (Entry v blobref')
+traverseBlobRef f = \case
+  Insert v -> pure (Insert v)
+  InsertWithBlob v blobref -> InsertWithBlob v <$> f blobref
+  Mupdate v -> pure (Mupdate v)
+  Delete -> pure Delete
+
```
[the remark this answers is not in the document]
wenkokke: I'll remove it. Needed it at some point, and didn't spot the Traversable instance.

wenkokke:
> Regarding deduplication, there are two avenues I can think of.
> 
> 1. Deduplicating `WriteBufferReader` and `RunReader`. And my thinking is that we could create a more general `KOpsReader` type that can be instantiated to be either a `WriteBufferReader` or `RunReader`. We should probably do this first, because it requires fewer changes than the next avenue
> 2. Deduplicating `WriteBufferWriter` and `RunBuilder`. My thinking is that we could construct this hierarchy:
> 
> ```
>                       /-------- WriteBufferBuilder
>                      /
> KOpsAcc <- KOpsBuilder <---------- RunBuilder
>                             /
> ProbeAcc <- ProbeBuilder <-/
> ```
> 
> `RunAcc` is split up into `KOpsAcc` and `ProbeAcc`. The former takes care of building pages for key/ops and blob files. The latter takes care of building the bloom filters and fence pointer indices. `RunBuilder` is split in two in the same way, doing the `IO` for the `Acc`umulators. The _new_ `RunBuilder` is the composition of a `KOPsBuilder` and a `ProbeBuilder`. The `WriteBufferBuilder` is just a small wrapper around a `KOpsBuilder`

I assume this is for future work rather than as part of this PR?

src/Database/LSMTree/Internal/WriteBufferReader.hs
wenkokke: I feel that if we pass in a `WriteBufferBlobs` (as per the next comment), then storing a `BlobFile` in the incremental reader would be odd, since it would require unpacking and repacking the `WriteBufferBlobs`. Given the second comment, would it make sense to leave the first one as a `WriteBufferBlobs`?

src/Database/LSMTree/Internal/WriteBufferReader.hs
wenkokke: I guess minor point against the refactor is that the RunReader API suggests (and is in fact mostly used) to read blobs into an existing WriteBufferBlobs, whereas doing so is always an error with the WriteBufferReader API... I think that's why I did it this way?

wenkokke:
> A general note: it would have been easier to review the PR if the commits had been more structured and informative

I'll admit I didn't structure my commit much. However, do we have any guidelines for structuring commits and commit messages? Something like [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)?

src/Database/LSMTree/Internal/Paths.hs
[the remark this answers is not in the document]
wenkokke: Done

src/Database/LSMTree/Internal.hs
```diff
@@ -1246,22 +1225,25 @@ openSnapshot sesh label tableType override snap resolve = do
...
-                                                  releaseRef
 
-        let actDir = Paths.activeDir (sessionRoot seshEnv)
+        let activeDir = Paths.activeDir (sessionRoot seshEnv)
+
+        -- Read write buffer
+        activeWriteBufferNumber <- uniqueToRunNumber <$> incrUniqCounter uc
+        let activeWriteBufferPaths = Paths.WriteBufferFsPaths (Paths.getActiveDir activeDir) activeWriteBufferNumber
+        let snapWriteBufferPaths = Paths.WriteBufferFsPaths (Paths.getNamedSnapshotDir snapDir) snapWriteBuffer
+        (tableWriteBuffer, tableWriteBufferBlobs) <- openWriteBuffer reg resolve hfs hbio snapWriteBufferPaths activeWriteBufferPaths
```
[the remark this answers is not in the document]
wenkokke: For clarify, that probably entails renaming `snapDir` to `namedSnapDir`, otherwise we'll have `snapDir` and `snapsDir`.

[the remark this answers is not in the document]
wenkokke: We have to create `snapWriteBufferPaths` (or at least somehow pass the `snapWriteBuffer` counter as well as `snapDir`. Is it really worth inlining just the `activeWriteBufferPaths` creation?

src/Database/LSMTree/Internal/WriteBufferReader.hs
wenkokke: I believe we cannot make this refactoring. In `openSnapshot`, we *copy* the blobs file, which means that the blob references in the key/ops file don't require updating. If we permit `readWriteBuffer` to take a pre-existing blobs buffer, then we can no longer *copy* the blobs file, since we need to append them to the pre-existing blobs buffer, which means we need to update all the blob references, which means we need to actually read and copy each individual blob.

test/Test/Database/LSMTree/Internal/Run.hs
wenkokke: Do we want to maintain a separate `readKOps` function for this?

test/Test/Database/LSMTree/Internal/Run.hs
wenkokke: The purpose of this test is to assert that writing runs is the same as writing write buffers, which means that if I read the data back in, I'm testing the writer and the reader at the same time, which would test something else... It would also require a new function to read back runs, which means that we're no longer testing the write buffer serialisation against the actual run serialisation mechanism.

src/Database/LSMTree/Internal.hs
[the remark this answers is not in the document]
wenkokke: Done as part of a different refactor

src/Database/LSMTree/Internal/Snapshot.hs
wenkokke: Done

src/Database/LSMTree/Internal/WriteBufferReader.hs
wenkokke: Done with an added NOTE in the comment

src/Database/LSMTree/Internal/WriteBufferReader.hs
wenkokke: Done

src/Database/LSMTree/Internal.hs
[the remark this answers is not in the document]
jorisdral: Looks like this is not changed yet

src/Database/LSMTree/Internal.hs
```diff
@@ -1160,42 +1158,22 @@ createSnapshot resolve snap label tableType t = do
...
         snapLevels <- toSnapLevels (tableLevels content)
+
         -- Hard link runs into the named snapshot directory
         snapLevels' <- snapshotRuns reg hbio snapDir snapLevels
 
-        let snapMetaData = SnapshotMetaData label tableType (tableConfig t) snapLevels'
+        let snapMetaData = SnapshotMetaData label tableType (tableConfig t) snapWriteBufferNumber snapLevels'
             SnapshotMetaDataFile contentPath = Paths.snapshotMetaDataFile snapDir
             SnapshotMetaDataChecksumFile checksumPath = Paths.snapshotMetaDataChecksumFile snapDir
         writeFileSnapshotMetaData hfs contentPath checksumPath snapMetaData
```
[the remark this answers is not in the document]
jorisdral: Looks like this is not changed yet

src/Database/LSMTree/Internal.hs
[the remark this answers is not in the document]
jorisdral: Looks like this is not changed yet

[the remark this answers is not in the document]
jorisdral: Looks like this is not changed yet

test/Test/Database/LSMTree/Internal/Run.hs
jorisdral: Let's leave it out for now

src/Database/LSMTree/Internal/WriteBufferReader.hs
jorisdral: I have a commit I'll push to this branch shortly (with your permission) to show you what I meant

jorisdral (commented):
Seems like almost everything is resolved. I have two commits to include in this PR, however, if that's okay with you @wenkokke ?

src/Database/LSMTree/Internal/WriteBufferReader.hs
jorisdral: See https://github.com/IntersectMBO/lsm-tree/pull/478/commits/66c7796bdf8b00ec96c7e2d2837f26047a95f30a

Issue #444, opened 2024-10-29
[BUG] `NoThunks` assertions failures when using `modifyMVarMasked`


jorisdral:
**Describe the bug**
The use of `strict-mvar`'s `modifyMVarMasked_` and `modifyMVarMasked` functions in the `Database.LSMTree.Internal.MergeSchedule` module make our `NoThunks` tests fail. We start to use these functions in #426, and subsequently `NoThunks` assertions started failing. As part of #426, the assertions are temporarily disabled.

**To Reproduce**
1. Go to `Test.Database.LSMTree.Normal.StateMachine.runIO`
2. Replace the line `const id (assertNoThunks sesh) $ pure ()` by `assertNoThunks sesh $ pure ()`
3. Run `cabal run lsm-tree-test -- -p '$NF=="propLockstep_RealImpl_MockFS_IO"'`

**Expected behavior**
These `strict-mvar` functions are expected to be strict enough to not leave any (shallow) thunks in our code. We should investigate why the `NoThunks` assertions fail, and then enable them again.

**Output**
```
❯ cabal run lsm-tree-test -- -p '$NF=="propLockstep_RealImpl_MockFS_IO"'
lsm-tree
  Normal.StateMachine
    propLockstep_RealImpl_MockFS_IO: FAIL (4.87s)
      *** Failed! (after 27 tests and 79 shrinks):
      Exception:
        Assertion failed: found thunkThunkInfo {thunkContext = ["MergingRunState","StrictMVar IO","MergingRun","incomingRuns","Level","Boxed.Vector","tableLevels","TableContent","RWState","StrictTVar IO","RWVar","tableContent","TableHandleEnv IO HandleMock","TableHandleState","RWState","StrictTVar IO","RWVar","tableHandleState","TableHandle","Map","StrictMVar IO","sessionOpenTables","SessionEnv","SessionState","RWState","StrictTVar IO","RWVar","sessionState","Session'"], thunkInfo = Nothing}
        CallStack (from HasCallStack):
          error, called at src-extras/Database/LSMTree/Extras/NoThunks.hs:75:33 in lsm-tree-0.1.0.0-inplace-extras:Database.LSMTree.Extras.NoThunks
[...]

PR #479, opened 2024-11-25
Re-enable NoThunks tests (#444) 

jorisdral:
The `NoThunks` assertions are re-enabled, but in cases where I expect failures related to `StrictMVar`s, we ignore the thunk and keep checking further into the `MVar` contents.

This is arguably better than not checking `NoThunks` at all, though still unfortunate.

### DIFF
--- a/src-extras/Database/LSMTree/Extras/NoThunks.hs
+++ b/src-extras/Database/LSMTree/Extras/NoThunks.hs
@@ -68,6 +68,8 @@ import           KMerge.Heap
 import           NoThunks.Class
 import           System.FS.API
 import           System.FS.BlockIO.API
+import           System.FS.IO
+import           System.FS.Sim.MockFS
 import           Test.QuickCheck (Property, Testable (..), counterexample)
 import           Unsafe.Coerce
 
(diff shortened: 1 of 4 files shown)


test/Test/Database/LSMTree/Normal/StateMachine/DL.hs
```diff
@@ -1,3 +1,5 @@
+{-# LANGUAGE OverloadedLists #-}
```
mheinzel, line 1: Where is that used?

mheinzel (approved):

test/Test/Database/LSMTree/Normal/StateMachine/DL.hs
jorisdral: Nothing! Forgot to remove it after I made some other changes

src-extras/Database/LSMTree/Extras/NoThunks.hs
```diff
@@ -564,11 +566,29 @@ instance NoThunks a => NoThunks (StrictTVar IO a) where
...
-      noThunks ctx x
+-- TODO: in some cases, strict-mvar functions leave thunks behind, in particular
+-- modifyMVarMasked and modifyMVarMasked_. So in some specific cases we evaluate
+-- the contents of the MVar to WHNF, and keep checking nothunks from there. See
+-- lsm-tree#444.
+instance (NoThunks a, Typeable a) => NoThunks (StrictMVar IO a) where
+  showTypeOf (p :: Proxy (StrictMVar IO a)) = show $ typeRep p
+  wNoThunks ctx var
+    | Just (Proxy :: Proxy (MergingRunState IO HandleIO))
+        <- gcast (Proxy @a)
```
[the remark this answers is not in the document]
jorisdral: The quantified constraint in `NoThunksIOLike'` will throw a compiler error telling us to mark the instances for `StrictMVar` as incoherent, which then makes the tests fail... I haven't found a solution for it, but I'll add a note on the class describing why we're using casts (for now)

PR #480, opened 2024-11-26
Rename `blob` and `blobref` type variables to `b`

jorisdral:
Follow-up to #469. It's a rather pervasive change, so I'm keeping this PR in draft mode until more important PRs have landed on `main`


jorisdral:
Should be ready to review now!

dcoutts (approved):
LGTM!

You decide when is the best moment to merge, wrt the other in-flight PRs.

PR #495, opened 2024-12-10
refcount: define macro-constraint `HasCallStackIfDebug`

wenkokke:
@dcoutts This PR defines a `HasCallStackIfDebug` macro which

```c
#ifdef NO_IGNORE_ASSERTS
#define HasCallStackIfDebug HasCallStack
#else
#define HasCallStackIfDebug ()
#endif
```

which does not throw a warning when used as

```hs
newRef ::
     (RefCounted obj, FinaliserM obj ~ m, PrimMonad m)
  => HasCallStackIfDebug
  => m ()
  -> (RefCounter m -> obj)
  -> m (Ref obj)
```

but is closer to Haskell syntax than the previous `HAS_CALL_STACK` macro.

### DIFF
--- a/src-control/Control/RefCount.hs
+++ b/src-control/Control/RefCount.hs
@@ -200,9 +200,9 @@ class RefCounted obj where
   getRefCounter       :: obj -> RefCounter (FinaliserM obj)
 
 #ifdef NO_IGNORE_ASSERTS
-#define HAS_CALL_STACK => HasCallStack
+#define HasCallStackIfDebug HasCallStack
 #else
-#define HAS_CALL_STACK
+#define HasCallStackIfDebug ()
 #endif
 
 -- GHC says specialising is too complicated! But it's ok, each of these can
@@ -219,7 +219,7 @@ class RefCounted obj where
 --
 newRef ::
      (RefCounted obj, FinaliserM obj ~ m, PrimMonad m)
-     HAS_CALL_STACK
+  => HasCallStackIfDebug
   => m ()
   -> (RefCounter m -> obj)
   -> m (Ref obj)
@@ -234,7 +234,7 @@ newRef finaliser mkObject = do
 --
 releaseRef ::
      (RefCounted obj, FinaliserM obj ~ m, PrimMonad m, MonadMask m)
-     HAS_CALL_STACK
+  => HasCallStackIfDebug
   => Ref obj
   -> m ()
 releaseRef ref@Ref{refobj} = do
@@ -268,7 +268,7 @@ deRef ref@Ref{refobj} =
 withRef ::
      forall m obj a.
      PrimMonad m
-     HAS_CALL_STACK
+  => HasCallStackIfDebug
   => Ref obj
   -> (obj -> m a)
   -> m a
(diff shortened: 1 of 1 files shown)


wenkokke:
Note that all other variants _do_ throw a warning.

The following use of the above macro throws a warning:

```hs
newRef ::
     (RefCounted obj, FinaliserM obj ~ m, PrimMonad m, HasCallStackIfDebug) 
  => m ()
  -> (RefCounter m -> obj)
  -> m (Ref obj)
```

The following definition always throws a warning, regardless of use:

```hs
type HasCallStackIfDebug :: Constraint
#ifdef NO_IGNORE_ASSERTS
type HasCallStackIfDebug = HasCallStack
#else
type HasCallStackIfDebug = ()
#endif
```

src-control/Control/RefCount.hs
```diff
@@ -200,9 +200,9 @@ class RefCounted obj where
   getRefCounter       :: obj -> RefCounter (FinaliserM obj)
 
 #ifdef NO_IGNORE_ASSERTS
-#define HAS_CALL_STACK => HasCallStack
+#define HasCallStackIfDebug HasCallStack
 #else
-#define HAS_CALL_STACK
+#define HasCallStackIfDebug ()
 #endif
```
jorisdral, line 206: I prefer the all-caps macros because it makes it clear that it is not actual haskell syntax

src-control/Control/RefCount.hs
```diff
@@ -219,7 +219,7 @@ class RefCounted obj where
 --
 newRef ::
      (RefCounted obj, FinaliserM obj ~ m, PrimMonad m)
-     HAS_CALL_STACK
+  => HasCallStackIfDebug
```
jorisdral, line 222: What's the reason that this does not throw a warning?

wenkokke: By the looks of it, GHC combines successive constraint arrows, so by the time it reaches the warning subsystem, `(A, B) => () => x` has turned into `(A, B) => x` by concatenation, whereas it does no other normalisation, so `(A, B, ()) => x` remains unchanged.

src-control/Control/RefCount.hs
wenkokke: Do you feel that with the all-caps name, this is an improvement?

wenkokke: There is, of course, the risk that this is highly dependent on what is arguably a GHC bug.

wenkokke:
My preferred version would be to use CPP to define a type and simply disable the warning, since we're purposefully creating an empty constraint, so disabling the warning is sensible. That way, the code we're writing actually can follow the general rule that each CPP section should be parseable Haskell.

dcoutts (approved):
LGTM!

I have no problem with using a "normal" name rather than a CPP_ISH name. I did the CPP_ISH name because the thing I did was clearly not valid syntax, so it provided a hint that some evil CPP was at play.

And I don't mind relying on GHC warning quirks here.

Related: we should file a ghc bug/feature request to ask that redundant () constraints not get warned about but just get eliminated.

PR #497, opened 2024-12-10
Test generators and shrinkers for snapshot metadata

jorisdral:
I wrote these commits when I was debugging a looping shrinker, but it turned out it was not the shrinkers for metadata that was looping. It was the shrinker for the  `Errors` type instead (see [fs-sim#84](https://github.com/input-output-hk/fs-sim/pull/84)). So even though the new tests in this PR did not find any bugs, we might as well include the tests on `main`


dcoutts (approved):
LGTM.

PR #501, opened 2024-12-18
Remove commented out `levelsInvariant` code

jorisdral:
The code is available in the git history if we ever want to reuse it.


mheinzel (approved):
:100:

dcoutts (approved):

PR #502, opened 2024-12-18
Make it possible to choose the index type

jeltsch:
This pull requests enables choice of index types via table configurations.

### DIFF
--- a/cabal.project.release
+++ b/cabal.project.release
@@ -20,9 +20,7 @@ constraints: bloomfilter <0
 -- import in case they don't have liburing installed... Currently, it is not
 -- possible to have conditionals on package flags in a project file. Otherwise,
 -- we could add a conditional on (+serialblockio) to remove this import automatically.
---import: cabal.project.blockio-uring
-
-flags: +serialblockio
+import: cabal.project.blockio-uring
 
 if impl(ghc >=9.12)
   allow-newer: base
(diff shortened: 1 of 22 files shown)


jorisdral (commented):
The PR is a draft, but I thought I'd write down my thoughts anyway.

### Type parameter `i` or `j`

We parameterise over `j` (index accumulator) in a lot of places, where I think `i` (the index itself) would be nicer conceptually. I don't think it actually complicates the code to have an additional type family

```haskell
type family Acc i
type instance Acc IndexOrdinary = IndexOrdinaryAcc
type instance Acc IndexCompact = IndexCompactAcc
```

### Existentially quantifying at the bottom

We can still give this a try to see if it works and whether we think it's nicer to have. I'd be happy to try if you want me to

jeltsch:
> ### Existentially quantifying at the bottom
> 
> We can still give this a try to see if it works and whether we think it's nicer to have. I’d be happy to try if you want me to.

Actually, I’m meanwhile eager to try it myself. 🙂 I think that I have some rather clear idea of how to do it and would be happy to try it out.

Thank you for all your other comments. I’ve read through all of them but would like to address them only after having explored that alternative approach of existentially quantifying at the bottom, as we might decide to follow this approach and this would make many of those comments obsolete.

jorisdral:
Closing in favour of #522

PR #506, opened 2024-12-20
Make it possible to choose the index type (opacity)

jeltsch:
This pull requests enables choice of index types via table configurations. It solves the same problem as #502 but using an alternative implementation, one that doesn’t involve adding index and index accumulator type parameters but rather immediately quantifies index and index accumulator types.

### DIFF
--- a/src/Database/LSMTree/Internal/Config.hs
+++ b/src/Database/LSMTree/Internal/Config.hs
@@ -19,8 +19,6 @@ module Database.LSMTree.Internal.Config (
   , BloomFilterAlloc (..)
   , defaultBloomFilterAlloc
   , bloomFilterAllocForLevel
-    -- * Fence pointer index
-  , FencePointerIndex (..)
     -- * Disk cache policy
   , DiskCachePolicy (..)
   , diskCachePolicyForLevel
@@ -36,6 +34,7 @@ import           Data.Word (Word64)
 import           Database.LSMTree.Internal.Assertions (assert,
                      fromIntegralChecked)
 import           Database.LSMTree.Internal.Entry (NumEntries (..))
+import           Database.LSMTree.Internal.Index.Some (IndexType (CompactIndex))
 import           Database.LSMTree.Internal.Run (RunDataCaching (..))
 import           Database.LSMTree.Internal.RunAcc (RunBloomFilterAlloc (..))
 import qualified Monkey
@@ -65,7 +64,7 @@ data TableConfig = TableConfig {
     -- applications.
   , confWriteBufferAlloc  :: !WriteBufferAlloc
   , confBloomFilterAlloc  :: !BloomFilterAlloc
-  , confFencePointerIndex :: !FencePointerIndex
+  , confFencePointerIndex :: !IndexType
     -- | The policy for caching key\/value data from disk in memory.
   , confDiskCachePolicy   :: !DiskCachePolicy
   , confMergeSchedule     :: !MergeSchedule
@@ -275,39 +274,6 @@ bloomFilterAllocForLevel conf (LevelNo l) =
       | otherwise = foldr (\x r k -> case k of
                                       0 -> Just x
                                       _ -> r (k-1)) (const Nothing) xs n
-
-{-------------------------------------------------------------------------------
-  Fence pointer index
--------------------------------------------------------------------------------}
-
--- | Configure the type of fence pointer index.
---
--- TODO: this configuration option currently has no effect: 'CompactIndex' is
--- always used.
-data FencePointerIndex =
-    -- | Use a compact fence pointer index.
-    --
-    -- The compact index type is designed to work with keys that are large
-    -- cryptographic hashes, e.g. 32 bytes.
-    --
-    -- When using the 'IndexCompact', additional constraints apply to the
-    -- 'Database.LSMTree.Internal.Serialise.Class.serialiseKey' function. The
-    -- __Minimal size__ law should be satisfied:
-    --
-    -- [Minimal size] @'Database.LSMTree.Internal.RawBytes.size'
-    --   ('Database.LSMTree.Internal.Serialise.Class.serialiseKey' x) >= 8@
-    --
-    -- Use 'Database.LSMTree.Internal.Serialise.Class.serialiseKeyMinimalSize'
-    -- to test this law.
-    CompactIndex
-    -- | Use an ordinary fence pointer index, without any constraints on
-    -- serialised keys.
-  | OrdinaryIndex
-  deriving stock (Show, Eq)
-
-instance NFData FencePointerIndex where
-  rnf CompactIndex  = ()
-  rnf OrdinaryIndex = ()
 
 {-------------------------------------------------------------------------------
   Disk cache policy
--- a/src/Database/LSMTree/Internal/Config.hs
+++ b/src/Database/LSMTree/Internal/Config.hs
@@ -19,6 +19,9 @@ module Database.LSMTree.Internal.Config (
   , BloomFilterAlloc (..)
   , defaultBloomFilterAlloc
   , bloomFilterAllocForLevel
+    -- * Fence pointer index
+  , FencePointerIndex (..)
+  , indexTypeForRun
     -- * Disk cache policy
   , DiskCachePolicy (..)
   , diskCachePolicyForLevel
@@ -34,7 +37,7 @@ import           Data.Word (Word64)
 import           Database.LSMTree.Internal.Assertions (assert,
                      fromIntegralChecked)
 import           Database.LSMTree.Internal.Entry (NumEntries (..))
-import           Database.LSMTree.Internal.Index.Some (IndexType (CompactIndex))
+import           Database.LSMTree.Internal.Index.Some (IndexType (..))
 import           Database.LSMTree.Internal.Run (RunDataCaching (..))
 import           Database.LSMTree.Internal.RunAcc (RunBloomFilterAlloc (..))
 import qualified Monkey
@@ -64,7 +67,7 @@ data TableConfig = TableConfig {
     -- applications.
   , confWriteBufferAlloc  :: !WriteBufferAlloc
   , confBloomFilterAlloc  :: !BloomFilterAlloc
-  , confFencePointerIndex :: !IndexType
+  , confFencePointerIndex :: !FencePointerIndex
     -- | The policy for caching key\/value data from disk in memory.
   , confDiskCachePolicy   :: !DiskCachePolicy
   , confMergeSchedule     :: !MergeSchedule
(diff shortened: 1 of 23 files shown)


src/Database/LSMTree/Internal/Config.hs
```diff
@@ -276,39 +275,6 @@ bloomFilterAllocForLevel conf (LevelNo l) =
...
-    CompactIndex
-    -- | Use an ordinary fence pointer index, without any constraints on
-    -- serialised keys.
-  | OrdinaryIndex
-  deriving stock (Show, Eq)
-
-instance NFData FencePointerIndex where
-  rnf CompactIndex  = ()
-  rnf OrdinaryIndex = ()
-
```
jorisdral, line 311: I'd prefer to keep this as its own config option, and then have a small `FencePointerIndex -> IndexType` translation here like we have `bloomFilterAllocForLevel` for the bloom filter allocation option.  If we do this, it means the index config option and index type flag can then change independently in the future.

```haskell
  Fence pointer index

-- | Configure the type of fence pointer index.
--
-- TODO: this configuration option currently has no effect: 'CompactIndex' is
-- always used.
data FencePointerIndex =
    -- | Use a compact fence pointer index.
    --
    -- The compact index type is designed to work with keys that are large
    -- cryptographic hashes, e.g. 32 bytes.
    --
    -- When using the 'IndexCompact', additional constraints apply to the
    -- 'Database.LSMTree.Internal.Serialise.Class.serialiseKey' function. The
    -- __Minimal size__ law should be satisfied:
    --
    -- [Minimal size] @'Database.LSMTree.Internal.RawBytes.size'
    --   ('Database.LSMTree.Internal.Serialise.Class.serialiseKey' x) >= 8@
    --
    -- Use 'Database.LSMTree.Internal.Serialise.Class.serialiseKeyMinimalSize'
    -- to test this law.
    CompactIndex
    -- | Use an ordinary fence pointer index, without any constraints on
    -- serialised keys.
  | OrdinaryIndex
  deriving stock (Show, Eq)

instance NFData FencePointerIndex where
  rnf CompactIndex  = ()
  rnf OrdinaryIndex = ()
  
indexTypeForRun :: FencePointerIndex -> Indextype
indexTypeForRun = ...
```

src/Database/LSMTree/Internal/Index.hs
```diff
@@ -49,38 +49,13 @@ class Index i where
...
     sizeInPages :: i -> NumPages
 
-    {-|
-        Yields the header of the serialised form of an index.
-
-        See the documentation of the 'Index' class for how to generate a
-        complete serialised index.
-    -}
-    headerLBS :: Proxy# i -> LazyByteString
-
```
jorisdral, line 59: Why is this removed from the class?

src/Database/LSMTree/Internal/Index.hs
```diff
@@ -49,38 +49,13 @@ class Index i where
...
-        The contents of the byte string may be directly used as the backing
-        memory for the constructed index. Currently, this is done for compact
-        indexes.
-
-        For deserialising numbers, the endianness of the host system is used. If
-        serialisation has been done with a different endianness, this mismatch
-        is detected by looking at the type–version indicator.
-    -}
-    fromSBS :: ShortByteString -> Either String (NumEntries, i)
```
jorisdral, line 83: Why is this removed from the class?

src/Database/LSMTree/Internal/Index.hs
```diff
@@ -95,7 +70,7 @@ class Index i where
         (Currently, construction of compact indexes needs the former and
         construction of ordinary indexes needs the latter bound.)
 -}
-class Index (ResultingIndex j) => IndexAcc j where
+class (Index (ResultingIndex j), NFData (ResultingIndex j)) => IndexAcc j where
```
jorisdral, line 73: What do you think about replacing this constraint on the class with a constraint in the `SomeIndexAcc` existential? For example:

```haskell
data SomeIndexAcc s = forall j . (IndexAcc j, NFData (ResultingIndex j)) => SomeIndexAcc (j s)
```

src/Database/LSMTree/Internal/Index/Some.hs
```diff
@@ -0,0 +1,133 @@
...
+    | OrdinaryIndex
+    deriving stock (Show, Eq)
+
+instance NFData IndexType where
+
+    rnf CompactIndex  = ()
+    rnf OrdinaryIndex = ()
+
+-- | The type of indexes.
+data SomeIndex = forall i . (Index i, NFData i) => SomeIndex i
```
jorisdral, line 69: We could also go for explicitly listing the indexes:

```haskell
data SomeIndex = SomeCompactIndex CompactIndex | SomeOrdinaryIndex OrdinaryIndex
```

The family of available indexes is (indirectly) closed anyway, so maybe this is nicer to work with?

src/Database/LSMTree/Internal/Index/Some.hs
```diff
@@ -0,0 +1,133 @@
...
+    or trailing space. In the case of a compact index, the contents of the byte
+    string must be stored 64-bit-aligned.
+
+    For deserialising numbers, the endianness of the host system is used. If
+    serialisation has been done with a different endianness, this mismatch is
+    detected by looking at the type–version indicator.
+-}
+fromSBS :: IndexType -> ShortByteString -> Either String (NumEntries, SomeIndex)
+fromSBS CompactIndex  = fmap (second SomeIndex) . IndexCompact.fromSBS
+fromSBS OrdinaryIndex = fmap (second SomeIndex) . IndexOrdinary.fromSBS
```
jorisdral, line 105: Okay, I see that this is what replaced the class methods of the same name. You could also keep the class methods and use proxies. What do you think?

src/Database/LSMTree/Internal/Index/Some.hs
jorisdral, line 1: So, by changing from `CompactIndex` to `SomeIndex` we do have some extra indirections and missed specialisations in our code paths, like unwrapping `SomeIndex` into an index `i` with `Index i` in scope every time we `search`. Maybe there are ways to minimise this overhead? In any case, we should definitely run the benchmarks to see if it adds considerable performance overhead.

jorisdral (commented):
Looks like an okay alternative to #502 to me

src/Database/LSMTree/Internal/Index.hs
```diff
@@ -49,38 +49,13 @@ class Index i where
...
-    -}
-    headerLBS :: Proxy# i -> LazyByteString
-
     {-|
         Yields the footer of the serialised form of an index.
 
         See the documentation of the 'Index' class for how to generate a
         complete serialised index.
     -}
     finalLBS :: NumEntries -> i -> LazyByteString
```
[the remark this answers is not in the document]
jeltsch: “Key–op pairs” is the same as “entries”, right?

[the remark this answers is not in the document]
jeltsch: I’ve created issue #517 for this.

src/Database/LSMTree/Internal/Config.hs
jeltsch: Yes, I think this is a better solution. I’ve made corresponding changes in 6ba79a6577fb681a4d556895eb8e00818a81d368.

src/Database/LSMTree/Internal/Index.hs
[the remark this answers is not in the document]
jorisdral: > “Key–op pairs” is the same as “entries”, right?

An `Entry` does not contain a key, but the count of key-op pairs is equivalent to the count of entries

src/Database/LSMTree/Internal/Index.hs
jeltsch: Actually, there is not only the constraint in the declaration of the `IndexAcc` class but also one in the declaration of the `SomeIndex` type, similar to the one that you propose for `SomeIndexAcc`. I came up with this solution to satisfy the need to fully evaluate indexes as part of fully evaluating runs, but I have to say that I don’t think that this was a particularly good solution.

I’ve developed a different approach now: have `NFData` as a superclass of `Index`. This is slightly more restrictive, as it requires that *all* indexes can be fully evaluated, not just those that arise from index accumulators. Then again, we always construct indexes via index accumulators; so this restriction doesn’t restrict us in practice. Furthermore, this new approach has two advantages:
* It is simpler.
* It expresses the actual idea behind constraining indexes: it should be possible to fully evaluate indexes, because indexes appear as part of runs, which we want to be able to fully evaluate.

I’ve created 09ca9f35e8742d85271379dc833abefbd96dda08, which implements this change.

src/Database/LSMTree/Internal/Index/Some.hs
jeltsch: I don’t really think that this would make it easier to work with values of `SomeIndex`. At the same time, it might make it more difficult to add other index types in case we wish to do so.

Furthermore, it would encourage us to bundle the implementations of each individual operation for all index types. For example, we may define a single `search` function that contains the search functionality for the compact and the ordinary index and picks the correct one based on pattern matching. However, I think that it’s better to bundle the implementations of all operations for each individual index type, like we do currently.

All in all, I think that avoiding data types with cases for the different index types is the way to go, which made me also rethink the current handling of `headerLBS`, `fromSBS`, and `newWithDefaults` (see [below](https://github.com/IntersectMBO/lsm-tree/pull/506#discussion_r1907396556)).

src/Database/LSMTree/Internal/Index/Some.hs
jeltsch: I’ve implemented the solution based on type proxies in 6ce5f35da37f569fa74511c621a6a3f3c16e8d09.

src/Database/LSMTree/Internal/Index/Some.hs
jeltsch: I had a quick chat with @dcoutts about this, and he suggested that the extra overhead would likely be bearable, given that index operations are only invoked rather rarely. Also, I don’t see ways to reduce this overhead at the moment. I agree that we should compare the new version to the compact-index-only one using the benchmarks. Could you perhaps point me to the benchmarks that are relevant in this regard?

src/Database/LSMTree/Internal/Index/Some.hs
jeltsch: Meanwhile, I consider the above-mentioned approach of having a `SomeIndex` type comprising a fixed set of index types to be an alternative to consider. This change of mind has to do with my further work on this pull request.

When trying to adapt the parts of the `lsm-tree` package that are outside of the core library to the current solution of existentially quantifying index and index accumulator types right away, I immediately ran into a problem: We need to make `SomeIndex` and `SomeIndexAcc` instances of `NoThunks` in the `extras` library, but this only works if it is known that the concretely used index and index accumulator types are instances of `NoThunks` as well. How could we guarantee this? We would have to insert `NoThunks` constraints into the index-related modules in the core library: as superclass constraints to the `Index` and `IndexAcc` classes or as constraints on the existentially quantified type variables in the definitions of `SomeIndex` and `SomeIndexAcc`. However, I understand that `NoThunks` should not be mentioned in the core library, as it is only related to testing. Furthermore, adding such constraints would result in `NoThunks` dictionaries carried around in production, which shouldn’t happen.

Your alternative solution with its fixed, known set of index types wouldn’t have such issues, as we would just have to make sure that the concrete index and index accumulator types that are used are instances of `NoThunks`. Therefore, I plan to realize this solution as a third variant of allowing for choice of index type. As part of this, I’d like to drop the type classes `Index` and `IndexAcc`, as they don’t seem to have a purpose with this solution anymore.

jeltsch: This 3rd variant is now implemented in #522, and we’ve decided to go for this solution.

PR #509, opened 2024-12-22
Resolving TODO: Converted FilePointer to Word representation

recursion-ninja:
# Description

A small refactoring addressing the TODO task of converting the `FilePointer` representation to `Word`, while exposing a `Word64` API.

The implementation in the PR replicates the definition of [`Data.Primitive.PrimVar.fetchAddInt`](https://hackage.haskell.org/package/primitive-0.9.0.0/docs/Data-Primitive-PrimVar.html#v:fetchAddInt), adapted for the unsigned `Word` type.

### DIFF
--- a/src/Database/LSMTree/Internal/WriteBufferBlobs.hs
+++ b/src/Database/LSMTree/Internal/WriteBufferBlobs.hs
@@ -197,9 +197,10 @@ mkWeakBlobRef (DeRef WriteBufferBlobs {blobFile}) blobspan =
 
 
 -- | A mutable file offset, suitable to share between threads.
+--
+-- This pointer is limited to 31-bit file offsets on 32-bit systems. This should
+-- be a sufficiently large limit that we never reach it in practice.
 newtype FilePointer m = FilePointer (PrimVar (PrimState m) Int)
---TODO: this would be better as Word64
--- this will limit to 31bit file sizes on 32bit arches
 
 instance NFData (FilePointer m) where
   rnf (FilePointer var) = var `seq` ()


src/Database/LSMTree/Internal/WriteBufferBlobs.hs
```diff
@@ -169,22 +189,47 @@ mkWeakBlobRef (DeRef WriteBufferBlobs {blobFile}) blobspan =
...
+  let !delta = fromIntegral n
+
+      -- Rely on the machine's binary reprentation of Int forming an additive ring
+      -- which is homomorphic with Word64 (such as 2's compliment); i.e.:
+      -- let limit = toEnum (maxBound :: Int) :: Word
+      -- in  limit + x === coerce ( (maxBound :: Int) + x )
+      convert :: Int -> Word64
+      convert = unsafeCoerce
+  in  convert <$> fetchAddInt var delta
+#endif
```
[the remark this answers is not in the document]
jorisdral: Of course, this does not solve the problem for 32 bit architectures, but we're already assuming architectures to be `64` bits (see `README.md`)

src/Database/LSMTree/Internal/WriteBufferBlobs.hs
```diff
@@ -169,22 +189,47 @@ mkWeakBlobRef (DeRef WriteBufferBlobs {blobFile}) blobspan =
...
     }
 
-
 -- | A mutable file offset, suitable to share between threads.
+#if MIN_VERSION_base(4,16,1)
+newtype FilePointer m = FilePointer (PrimVar (PrimState m) Word)
+#else
 newtype FilePointer m = FilePointer (PrimVar (PrimState m) Int)
---TODO: this would be better as Word64
--- this will limit to 31bit file sizes on 32bit arches
```
jorisdral, line 176: I wonder: is it even possible to do atomic operations on `Word64` on 32 bit architectures? Doesn't a 32-bit architecture need to represent a `64-bit` word as two machine words, losing the ability to modify both at the same time atomically?

src/Database/LSMTree/Internal/WriteBufferBlobs.hs
jorisdral, line 235: We can rely on `Int`s being 64 bits, because we require that from our users (see `README.md`). With that assumption, we can probably just use `fromIntegral <$> P.fetchAddInt var n`

src/Database/LSMTree/Internal/WriteBufferBlobs.hs
recursion-ninja: While it is probably chip dependant, I would imagine the answer is generally "yes," a 32-bit machine cannot be guaranteed to operate on `Word64` in an atomic manner. Is this a deal breaker for us?

src/Database/LSMTree/Internal/WriteBufferBlobs.hs
recursion-ninja: Hmm... in that case I think you are right. 

This is a case of *"6` hours of coding can save you 5 minutes of reading the documentation"*, haha.

We can mark this PR and "will not merge" and stick with the existing implementation, maybe adding a comment reminding the reader that we can assume 64-bit `Int`s.

src/Database/LSMTree/Internal/WriteBufferBlobs.hs
jorisdral: If we really want to have 64-bit file pointers on 32-bit systems, then we'd have to use something like `MVar` on those systems. Seeing as we don't support 32-bit systems, I think we don't have to consider that case

src/Database/LSMTree/Internal/WriteBufferBlobs.hs
jorisdral: Sure, let's do that. Maybe you can repurpose this PR for removing the TODO?

dcoutts (dismissed):
I don't think it's worth making this change. It does not fix the theoretical problem (of 31bit max file size for the WBB blob file on 32bit systems), which itself is hardly a practical limitation.

It doesn't solve it because it just switches from Int to Word, which is the same size. It'd need Word64 atomic support on 32bit systems (which ghc does not support, and which may not be available on all 32bit arches that ghc does support).

So I propose we just close this.

jorisdral:
I repurposed this PR to replace the TODO with a general comment about word size limits

jorisdral (approved):

PR #511, opened 2024-12-27
Add missing `HasCallStackIfDebug` pragmas in `Ref` specialisations

jorisdral:
Otherwise the call stacks will not be properly propagated


wenkokke (approved):
LGTM! Good catch on that double SPECIALISE pragmatic, odd that GHC doesn't complain about that.

PR #515, opened 2025-01-02
Add fault testing to the `StateMachine` tests

jorisdral:
See the commit messages for an explanation of the changes.

Currently, we only add fault injection to the `CreateSnapshot` and `OpenSnapshot` actions. This can be extended to other actions, but it would require even more infrastructure. For now, the snapshot functions are the main focus, because we want to test that corruption/disk faults are handled properly.


dcoutts (approved):
This looks great. I think the strategy is pretty reasonable.

PR #526, opened 2025-01-14
Do not put `::` on a newline

jorisdral:
It messes with syntax highlighting in GitHub and VSCode


mheinzel (approved):
> It messes with syntax highlighting in GitHub and VSCode

That's unfortunate. I generally like aligning the `::` with the `->`, but I think the PR makes sense, given that we already widely use the style with `::` directly after the function name.

PR #535, opened 2025-01-20
Add type parameter to merging runs in prototype

mheinzel:
Based on #477, which needs to be merged first.

This statically enforces that merging runs in different parts of the LSM can only have specific merge types.

Note the individual commits. There are multiple possible ways of going from the individual merge types to how the merge is performed. Two are shown here.

### DIFF
--- a/prototypes/ScheduledMerges.hs
+++ b/prototypes/ScheduledMerges.hs
@@ -93,12 +93,13 @@ type Levels s = [Level s]
 -- completed, the resulting run will become a resident run at this level.
 data Level s = Level !(IncomingRun s) ![Run]
 
--- | An additional optional last level, created as a result of 'union'. It can
--- not only contain an ongoing merge of multiple runs, but a nested tree of
--- merges. See Note [Table Unions].
-data UnionLevel s = NoUnion
-                    -- | We track the debt to make sure it never increases.
-                  | Union !(MergingTree s) !(STRef s Debt)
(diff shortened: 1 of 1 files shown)


dcoutts (approved):
LGTM and it looks like it would be relatively straightforward to apply the same merge type approach in the real impl.

prototypes/ScheduledMerges.hs
```diff
@@ -93,58 +93,79 @@ type Levels s = [Level s]
...
+-- to avoid supplying merge credits. It's not essential, but simplifies things
+-- somewhat.
+data IncomingRun s = Merging !MergePolicy !(MergingRun LevelMergeType s)
+                   | Single  !Run
 
 -- | The merge policy for a LSM level can be either tiering or levelling.
 -- In this design we use levelling for the last level, and tiering for
 -- all other levels. The first level always uses tiering however, even if
--- it's also the last level. So 'MergePolicy' and 'MergeLastLevel' are
+-- it's also the last level. So 'MergePolicy' and 'IncomingMergeType' are
```
jorisdral, line 107: `IncomingMergeType` does not seem to exist

mheinzel: Good catch. I renamed it to `LevelMergeType`, but seems like my search missed this occurrence because it considers the tick `'` as part of identifier names. I'll fix it as part of another PR.

PR #536, opened 2025-01-20
Implement MergeUnion merges

mheinzel:
The first commit is from #538, so maybe that should be merged first.

### DIFF
--- a/src/Database/LSMTree/Internal/Merge.hs
+++ b/src/Database/LSMTree/Internal/Merge.hs
@@ -252,7 +252,7 @@ steps ::
   => Merge m h
   -> Int  -- ^ How many input entries to consume (at least)
   -> m (Int, StepResult)
-steps Merge {..} requestedSteps = assertStepsInvariant <$> do
+steps m@Merge {..} requestedSteps = assertStepsInvariant <$> do
     -- TODO: ideally, we would not check whether the merge was already done on
     -- every call to @steps@. It is important for correctness, however, that we
     -- do not call @steps@ on a merge when it was already done. It is not yet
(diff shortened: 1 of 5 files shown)


mheinzel:
```
benchmarking Bench.Database.LSMTree.Internal.Merge/word64-mix-collisions-x4-midlevel/merge
time                 12.60 ms   (12.23 ms .. 12.99 ms)
                     0.995 R²   (0.991 R² .. 0.998 R²)
mean                 12.44 ms   (12.28 ms .. 12.63 ms)
std dev              453.7 μs   (377.2 μs .. 543.8 μs)
allocated:           1.000 R²   (1.000 R² .. 1.000 R²)
  iters              1.817e7    (1.817e7 .. 1.817e7)
  y                  -1617.647  (-3687.656 .. 1.456e-7)
variance introduced by outliers: 14% (moderately inflated)

benchmarking Bench.Database.LSMTree.Internal.Merge/word64-mix-collisions-x4-lastlevel/merge
time                 10.69 ms   (10.47 ms .. 10.92 ms)
                     0.996 R²   (0.991 R² .. 0.999 R²)
mean                 10.70 ms   (10.55 ms .. 11.02 ms)
std dev              557.2 μs   (289.9 μs .. 984.6 μs)
allocated:           1.000 R²   (1.000 R² .. 1.000 R²)
  iters              1.612e7    (1.612e7 .. 1.612e7)
  y                  -5.591e-8  (-2.268e-7 .. 2.475e-7)
variance introduced by outliers: 23% (moderately inflated)

benchmarking Bench.Database.LSMTree.Internal.Merge/word64-mix-collisions-x4-union/merge
time                 12.64 ms   (12.08 ms .. 13.20 ms)
                     0.988 R²   (0.979 R² .. 0.995 R²)
mean                 12.21 ms   (11.97 ms .. 12.59 ms)
std dev              749.4 μs   (538.0 μs .. 1.165 ms)
allocated:           1.000 R²   (1.000 R² .. 1.000 R²)
  iters              1.800e7    (1.800e7 .. 1.800e7)
  y                  -1075.871  (-2205.797 .. 1.442e-7)
variance introduced by outliers: 28% (moderately inflated)
```

In this benchmark, allocations and performance of union merges are similar to mid level merges. When there are no collisions, it is similar to last level merges (since both don't need to write Deletes), but in the presence of collisions there is some expected overhead: While the level merges can skip all but the first non-mupsert entry for each key, union merge has to resolve them all monoidally.

src/Database/LSMTree/Internal/Merge.hs
```diff
@@ -329,6 +342,47 @@ steps Merge {..} requestedSteps = assertStepsInvariant <$> do
...
+            -- resolved all entries for this key, write it
+            writeReaderEntry mergeType mergeBuilder key entry
+            go n
+          else do
+            (_, nextEntry, hasMore) <- Readers.pop mergeReaders
+            -- for resolution, we need the full second value to be present
+            let resolved = combineUnion mergeMappend
+                             (Reader.toFullEntry entry)
+                             (Reader.toFullEntry nextEntry)
+            handleEntry (n + 1) key (Reader.Entry resolved) hasMore
```
mheinzel, line 384: Note that if 'mergeMappend' is @const@ (e.g. when doing a union on a normal/non-monoidal table), it seems like we potentially do more work than necessary: Once we find the first non-delete entry, we could ignore the remaining entries for that key. However, this depends on union's semantics for blobs. With the current model, if any input table has a blob for a key, the result should also have a blob, so we have to keep looking for blobs.

Maybe that's not the right choice of semantics. If we have multiple blobs for a key, which one should we choose?

Maybe this is also a (rare) example of a situation where the combination of blobs and monoidal resolution (which can't appear with a normal/monoidal table distinction) make an optimisation difficult.

dcoutts (approved):
LGTM!

src/Database/LSMTree/Internal/Merge.hs
mheinzel: > So if I understand correctly there are two questions

Yes, although they are related: if we don't change the semantics for blobs, and don't make the normal/monoidal table distinction explicit in the internals here, then I don't think we gain much from optimising the `const` case.

PR #537, opened 2025-01-21
Renumber run and write buffer files starting from 0 in snapshots

jorisdral:
This feature is described in the `format-directory.md` document, but we were not doing it yet in the implementation.


mheinzel (approved):
Not strictly necessary, but nice that it now matches the documentation (and I do think it's a little nicer). The change adds a little bit of potential for confusion with the two counters, but I think it's clear from the variable names and comments.

dcoutts (approved):

PR #539, opened 2025-01-22
Test snapshot corruption

jorisdral:
A next step would be to apply the same explicit corruption method to the statemachine tests.

In parallel, we can add silent corruption to `fs-sim`, which currently only simulates noisy corruption. We could use this silent corruption simulation to eventually replace the explicit `flipFileBit` calls

### DIFF
--- a/test/Test/Database/LSMTree/Internal/Snapshot/FS.hs
+++ b/test/Test/Database/LSMTree/Internal/Snapshot/FS.hs
@@ -1,17 +1,36 @@
+{-# LANGUAGE OverloadedStrings #-}
+
 -- | Tests for snapshots and their interaction with the file system
 module Test.Database.LSMTree.Internal.Snapshot.FS (tests) where
 
 import           Codec.CBOR.Read (DeserialiseFailure)
-import           Control.Exception
+import           Control.Monad.Class.MonadThrow
+import           Control.Monad.IOSim (runSimOrThrow)
+import           Control.Tracer
+import           Data.Bifunctor (Bifunctor (..))
+import           Data.Maybe (fromJust)
+import qualified Data.Set as Set
+import qualified Data.Vector as V
+import           Data.Word
+import           Database.LSMTree.Extras (showPowersOf10)
+import           Database.LSMTree.Extras.Generators ()
+import           Database.LSMTree.Internal
+import           Database.LSMTree.Internal.Config
 import           Database.LSMTree.Internal.CRC32C
+import           Database.LSMTree.Internal.Entry
+import           Database.LSMTree.Internal.Paths
+import           Database.LSMTree.Internal.Serialise
 import           Database.LSMTree.Internal.Snapshot
 import           Database.LSMTree.Internal.Snapshot.Codec
+import qualified System.FS.API as FS
 import           System.FS.API
 import           System.FS.Sim.Error hiding (genErrors)
 import qualified System.FS.Sim.MockFS as MockFS
 import           Test.Database.LSMTree.Internal.Snapshot.Codec ()
+import           Test.QuickCheck
+import           Test.QuickCheck.Gen (genDouble)
 import           Test.Tasty
-import           Test.Tasty.QuickCheck as QC
+import           Test.Tasty.QuickCheck
 import           Test.Util.FS
 
 tests :: TestTree
@@ -20,6 +39,7 @@ tests = testGroup "Test.Database.LSMTree.Internal.Snapshot.FS" [
         prop_fsRoundtripSnapshotMetaData
     , testProperty "prop_fault_fsRoundtripSnapshotMetaData"
         prop_fault_fsRoundtripSnapshotMetaData
+    , testProperty "prop_flipSnapshotBit" prop_flipSnapshotBit
     ]
 
 -- | @readFileSnapshotMetaData . writeFileSnapshotMetaData = id@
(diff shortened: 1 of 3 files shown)


jorisdral:
Ready for review now!

test/Test/FS.hs
```diff
@@ -128,6 +137,47 @@ prop_numDirEntries (Path dir) isFiles (UniqueList paths) = runSimOrThrow $
     n = length paths
     xs = zip (getInfiniteList isFiles) paths
 
+{-------------------------------------------------------------------------------
+  Corruption
+-------------------------------------------------------------------------------}
+
+data WithBitOffset a = WithBitOffset Int a
```
wenkokke, line 124: Does this need to be an Int?

test/Test/Util/FS.hs
```diff
@@ -228,6 +237,74 @@ approximateEqStream (UnsafeStream infoXs xs) (UnsafeStream infoYs ys) =
...
+          absPath' = absPath </> p
+        isFile <- doesFileExist hfs absPath'
+        if isFile then
+          pure (File relPath' `Set.insert` acc)
+        else do
+          isDirectory <- doesDirectoryExist hfs absPath'
+          if isDirectory then
+            go (Directory relPath' `Set.insert` acc) relPath' absPath'
+          else
+            error "impossible"
```
wenkokke, line 283: Under what conditions would this happen? I suppose if the testing directory contains a node that isn't a file or directory, which can happen, but indicates a failure... so should perhaps this error have a slightly clearer message, in case it does actually happen?

test/Test/Util/FS.hs
```diff
@@ -228,6 +237,74 @@ approximateEqStream (UnsafeStream infoXs xs) (UnsafeStream infoYs ys) =
...
+{-------------------------------------------------------------------------------
+  List directory
+-------------------------------------------------------------------------------}
+
+data DirEntry a = Directory a | File a
+  deriving stock (Show, Eq, Ord, Functor)
+
+-- | List all files and directories in the given directory and recusively in all
+-- sub-directories.
+listDirectoryRecursive ::
```
wenkokke, line 259: Do we need some sort of test to make sure that these listings are actually complete?

test/Test/Util/FS.hs
```diff
@@ -368,6 +408,27 @@ instance Arbitrary NoCleanupErrors where
...
+      shrinkAsComponent s = fsPathComponentString <$>
+          shrink (FsPathComponent $ NE.fromList s)
+
+-- >>> [ c | c <- [minBound..maxBound], isPathChar c ]
+-- "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
+isPathChar :: Char -> Bool
+isPathChar c = isAscii c && (isLetter c || isDigit c)
+
+genPathChar :: Gen Char
+genPathChar = arbitraryASCIIChar `suchThat` isPathChar
```
wenkokke, line 427: Is this generating and filtering rather than generating from a list of options? Wouldn't that have really poor performance when generating?

test/Test/Database/LSMTree/Internal/Snapshot/FS.hs
```diff
@@ -137,3 +157,100 @@ instance Arbitrary TestErrors where
...
+  deriving stock (Show, Eq)
+
+instance Arbitrary Double_0_1 where
+  arbitrary = Double_0_1 <$> genDouble
+  shrink (Double_0_1 x) = [Double_0_1 x' | x' <- shrink x, 0 <= x', x' < 1]
+
+prop_flipSnapshotBit ::
+     Positive (Small Int)
+  -> V.Vector (Word64, Entry Word64 Word64)
+  -> Double_0_1 -- ^ Used to pick which file to corrupt
```
wenkokke, line 176: Isn't it possible to use an actual arbitrary choice over the set of files?

test/Test/Database/LSMTree/Internal/Snapshot/FS.hs
```diff
@@ -137,3 +157,100 @@ instance Arbitrary TestErrors where
...
+  -> Double_0_1 -- ^ Used to pick which bit to flip in the file we picked
```
wenkokke, line 177: Likewise isn't it possible to use an actual choice bounded by the length of the file?

wenkokke (approved):
LGTM. Left a few questions.

test/Test/FS.hs
jorisdral: Doesn't have to be, but if it were a `Word` we would have to convert to an `Int` at some point anyway

test/Test/Util/FS.hs
jorisdral: Preferably, but I have them on a separate branch. The test are quite chunky so I want to add them in a different PR at some point. For now I can only give you my word that they work well enough :stuck_out_tongue:

test/Test/Util/FS.hs
jorisdral: The performance is fine, but I agree it can be faster. I've changed it to pick `Char`s from a fixed list instead.

test/Test/Database/LSMTree/Internal/Snapshot/FS.hs
jorisdral: The set of files depends on multiple factors, and it's hard to predict from the test inputs. This seemed to simplest way to pick from a list that you don't know the size of beforehand. Do you have an idea for how to approach it differently?

test/Test/Database/LSMTree/Internal/Snapshot/FS.hs
jorisdral: Similarly, it's hard to predict the sizes of files from just the test inputs.

test/Test/Util/FS.hs
wenkokke: Maybe leave a TODO?

test/Test/Database/LSMTree/Internal/Snapshot/FS.hs
jorisdral: Left a TODO on the property

test/Test/Database/LSMTree/Internal/Snapshot/FS.hs
jorisdral: Left a TODO on the property

PR #540, opened 2025-01-22
fix: move checkCRC to CRC32C

wenkokke:


mheinzel (commented):
Is there a particular reason for moving the code, e.g. to enable re-using the code for in another place? If so, that can be useful to have in the commit message.

Seems like a sensible change overall.

wenkokke:
> Is there a particular reason for moving the code, e.g. to enable re-using the code for in another place? If so, that can be useful to have in the commit message.
> 
> Seems like a sensible change overall.

See #542. Note that this is merging into @jorisdral's active branch, so he should probably do the reviewing and merging.

mheinzel:
> Note that this is merging into @jorisdral's active branch, so he should probably do the reviewing and merging

Ah, sorry, I didn't see that. Thanks!

PR #542, opened 2025-01-22
fix: validate checksums in openWriteBuffer

wenkokke:
Should be merged after #539

### DIFF
--- a/src/Database/LSMTree/Internal/Paths.hs
+++ b/src/Database/LSMTree/Internal/Paths.hs
@@ -52,10 +52,8 @@ module Database.LSMTree.Internal.Paths (
 
 import           Control.Applicative (Applicative (..))
 import           Control.DeepSeq (NFData (..))
-import           Data.Bifunctor (Bifunctor (..))
 import qualified Data.ByteString.Char8 as BS
 import           Data.Foldable (toList)
-import           Data.Function ((&))
 import qualified Data.Map as Map
 import           Data.Maybe (fromMaybe)
 import           Data.String (IsString (..))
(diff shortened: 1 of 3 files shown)


src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -301,6 +304,12 @@ openWriteBuffer ::
   -> WriteBufferFsPaths
   -> m (WriteBuffer, Ref (WriteBufferBlobs m h))
 openWriteBuffer reg resolve hfs hbio uc activeDir snapWriteBufferPaths = do
+  -- Check the checksums
+  (expectedChecksumForKOps, expectedChecksumForBlob) <-
+    CRC.expectValidFile (writeBufferChecksumsPath snapWriteBufferPaths) . fromChecksumsFileForWriteBufferFiles
+      =<< CRC.readChecksumsFile hfs (writeBufferChecksumsPath snapWriteBufferPaths)
+  checkCRC hfs hbio False (unForKOps expectedChecksumForKOps) (writeBufferKOpsPath snapWriteBufferPaths)
+  checkCRC hfs hbio False (unForBlob expectedChecksumForBlob) (writeBufferBlobPath snapWriteBufferPaths)
```
jorisdral, line 312:
```suggestion
  checkCRC hfs hbio True (unForBlob expectedChecksumForBlob) (writeBufferBlobPath snapWriteBufferPaths)
```

src/Database/LSMTree/Internal/Paths.hs
```diff
@@ -322,16 +320,16 @@ writeBufferFilePathWithExt (WriteBufferFsPaths dir n) ext =
 -------------------------------------------------------------------------------}
 
 toChecksumsFileForWriteBufferFiles :: (ForKOps CRC.CRC32C, ForBlob CRC.CRC32C) -> CRC.ChecksumsFile
-toChecksumsFileForWriteBufferFiles checksums =
-  Map.fromList . toList $ checksums & bimap
```
jorisdral, line 325: Ah, the `toList (x,y) = [y]`, was that the problem?

src/Database/LSMTree/Internal/Paths.hs
```diff
@@ -321,10 +320,16 @@ writeBufferFilePathWithExt (WriteBufferFsPaths dir n) ext =
...
+    ]
   where
-    toChecksumsFileName :: String -> CRC.ChecksumsFileName
     toChecksumsFileName = CRC.ChecksumsFileName . BS.pack
+
+fromChecksumsFileForWriteBufferFiles :: CRC.ChecksumsFile -> Either String (ForKOps CRC.CRC32C, ForBlob CRC.CRC32C)
+fromChecksumsFileForWriteBufferFiles file = do
+  (,) <$> (ForKOps <$> fromChecksumFile writeBufferKOpsExt) <*> (ForBlob <$> fromChecksumFile writeBufferBlobExt)
+  where
+    fromChecksumFile key = maybe (Left $ "key not found: " <> key) Right (Map.lookup (CRC.ChecksumsFileName . fromString $ key) file)
```
jorisdral, line 335: Long lines: rewrap some

jorisdral (approved):
Nice :smile:

I have some small suggestions, but we can merge once they are resolved. We can probably also squash a few of the commits together

src/Database/LSMTree/Internal/Snapshot.hs
jorisdral, line 312: Hmm, I just noticed that we're in fact reading the blob file twice: once to check the CRC, then once below to copy it from the snapshot directory to the active directory. Maybe leave a TODO that we could optimise this by computing the CRC and copying the file in one traversal of the file?

src/Database/LSMTree/Internal/Paths.hs
wenkokke: Yes, indeed. For a foggy moment, my brain believed that the traversable instance of homogeneous tuples would do the right thing.

src/Database/LSMTree/Internal/Snapshot.hs
wenkokke: Shouldn't the fs cache help with this?

jorisdral: It will probably faster than if were to do actual I/O, but still a double traversal. And on more memory-constrained systems, we might not be able to benefit from the cache much

PR #544, opened 2025-01-22
fix: move FileFormatError to Paths

wenkokke:
Move FileFormatError to Paths module to allow reuse in validating the write buffer files.

### DIFF
--- a/src/Database/LSMTree/Internal/CRC32C.hs
+++ b/src/Database/LSMTree/Internal/CRC32C.hs
@@ -47,6 +47,10 @@ module Database.LSMTree.Internal.CRC32C (
   ChecksumError (..),
   checkCRC,
   expectChecksum,
+
+  -- * File format errors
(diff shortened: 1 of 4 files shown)


src/Database/LSMTree/Internal/Paths.hs
```diff
@@ -328,3 +333,20 @@ toChecksumsFileForWriteBufferFiles checksums =
...
+  deriving stock Show
+  deriving anyclass Exception
+
+expectValidFile ::
+     MonadThrow f
+  => FsPath
+  -> Either String a
+  -> f a
+expectValidFile _  (Right x)  = pure x
+expectValidFile fp (Left err) = throwIO $ FileFormatError fp err
```
jorisdral, line 352: I'd probably put this in the CRC module as well

jorisdral (approved):

PR #546, opened 2025-01-23
Test that `Entry` forms a `Semigroup` with `combineUnion`

jorisdral:
This is a follow-up to #536


test/Test/Database/LSMTree/Internal/Merge.hs
```diff
@@ -159,19 +159,8 @@ prop_MergeUnion fs hbio stepSize (SmallList rds) =
...
-
-    getValueAndBlob :: Entry.Entry v b -> Maybe (v, Maybe b)
-    getValueAndBlob = \case
-        Entry.Insert v           -> Just (v, Nothing)
-        Entry.InsertWithBlob v b -> Just (v, Just b)
-        Entry.Mupdate v          -> Just (v, Nothing)
-        Entry.Delete             -> Nothing
-
-    resolveValueAndBlob (v', Nothing) (v, b) = (mappendValues v' v, b)
-    resolveValueAndBlob (v', Just b)  (v, _) = (mappendValues v' v, Just b)
```
jorisdral, line 174: The reason I think this can be simplified is because we are doing almost the same as in `prop_MergeDistributes`, but we're also testing the semantics of `combineUnion`. We can test the semantics of `combineUnion` separately. What do you think?

jorisdral: e.g., see `prop_resolveEntriesNormalSemantics`: we could do something similar for the `combineUnion` semantics

test/Test/Database/LSMTree/Internal/Entry.hs
```diff
@@ -47,16 +54,32 @@ prop_resolveEntriesNormalSemantics ::
...
+  where
+    expected = from . sconcat . to $ es
+      where
+        to = fmap NormalUpdateSG
+        from = Just . unNormalUpdateSG
+
+    real = from . sconcat . to $ es
+      where
+        to = fmap (EntrySG . updateToEntryNormal)
+        from (EntrySG x) = entryToUpdateNormal x
```
mheinzel, line 66: Small nit, feel free to ignore: The two definitions looking that similar makes it a little unclear what the difference is, although I can see it when looking at each `to`. Maybe it would be clearer if called `toUpdate` and `toEntry`, or explicitly having `sconcat @EntrySG` etc.?

test/Test/Database/LSMTree/Internal/Merge.hs
mheinzel: That's true. I could add a property test like that, or do you want to do it?

I also think we could remove the whole `prop_MergeUnion`. It doesn't add much over the existing `prop_MergeDistributes`.

mheinzel (approved):
Nice!

test/Test/Database/LSMTree/Internal/Entry.hs
jorisdral: I've added some type signatures to make it more clear

test/Test/Database/LSMTree/Internal/Merge.hs
jorisdral: I'll add the property test now

test/Test/Database/LSMTree/Internal/Entry.hs
```diff
@@ -25,96 +28,352 @@ tests :: TestTree
...
+                       => Arbitrary (Union (Monoidal.Update v))
+
+instance Semigroup v => Semigroup (Union (Monoidal.Update v)) where
+  Union up1 <> Union up2 = Union $ case (up1, up2) of
+      (Monoidal.Delete     , _                  ) -> up2
+      (_                   , Monoidal.Delete    ) -> up1
+      (Monoidal.Insert v1  , Monoidal.Insert v2 ) -> Monoidal.Insert (v1 <> v2)
+      (Monoidal.Insert v1  , Monoidal.Mupsert v2) -> Monoidal.Insert (v1 <> v2)
+      (Monoidal.Mupsert v1 , Monoidal.Insert v2 ) -> Monoidal.Insert (v1 <> v2)
+      (Monoidal.Mupsert v1 , Monoidal.Mupsert v2) -> Monoidal.Insert (v1 <> v2)
```
mheinzel, line 333: I think the semantics we want to compare `combineUnion` to might not be this one on `{Normal,Monoidal}.Update` (which almost copies what `combineUnion` itself does), but the `Semigroup` instance of `Maybe v` and `Maybe (v, Monoid.First b)` instead. I'd say that makes the intended behaviour clearer.

PR #548, opened 2025-01-24
`First` semigroup instead of `First` monoid for blob semantics

jorisdral:


jorisdral:
> Looks good, but there is probably some documentation that should also get updated.

I've taken some of your suggestions from https://github.com/IntersectMBO/lsm-tree/commit/6790d82d2d63e0732dcca9779324b5bcb1a0d236. If there are other places where documenattion has to be updated, could you point me to them?

mheinzel:
> If there are other places where documenattion has to be updated, could you point me to them?

I briefly looked for other places, but it seems like we have never really been precise about what monoidal behaviour to expect from blobs, mainly because these cases aren't possible via the `Normal`/`Monoidal` API. Maybe a short sentence could be added to the docs of `combine` and `combineUnion` to clarify that blobs are taken from the left/new value only, but that's all I think.

mheinzel (approved):

PR #551, opened 2025-01-28
Fix out-of-bounds array access in `hFlipBit`.

### LINKED ISSUE #549: [BUG] "Access violation" errors in `lsm-tree-test`
**Describe the bug**

There are test failures on `main` because the tests throw errors with the message "Access violation in generated code when writing ...". It seems to be only happening on Windows.

**To Reproduce**
Follow the steps defined by the `haskell.yml` Github Actions workflow. It is not clear whether the bugs can reproduced locally, since I (@jorisdral) don't have a Windows machine to try it out on.

**Expected behavior**
Tests should pass

**Output**

https://github.com/IntersectMBO/lsm-tree/actions/runs/12958624976/job/36149389798
https://github.com/IntersectMBO/lsm-tree/actions/runs/12958624976/job/36149392812
https://github.com/IntersectMBO/lsm-tree/actions/runs/12958624976/job/36149392812
https://github.com/IntersectMBO/lsm-tree/actions/runs/12958624976/job/36149389798

The following might also be related, but does not produce the same error message:

https://github.com/IntersectMBO/lsm-tree/actions/runs/12951447199/job/36126516458

 - OS: Windows
 - OS version: Windows Server 2022
* GHC: 8.10.7, 9.8.2

**Additional context**
These tests have started showing up after https://github.com/IntersectMBO/lsm-tree/pull/539 was merged. Maybe there are some out-of-bounds byte array accesses introduced by that PR?

jorisdral:
Resolves #549

### DIFF
--- a/test/Test/FS.hs
+++ b/test/Test/FS.hs
@@ -124,16 +124,19 @@ createFile hfs p = withFile hfs p (WriteMode MustBeNew) $ \_ -> pure ()
 data WithBitOffset a = WithBitOffset Int a
   deriving stock Show
 
+bitLength :: BS.ByteString -> Int
+bitLength bs = BS.length bs * 8
+
 instance Arbitrary (WithBitOffset ByteString) where
   arbitrary = do
       bs <- arbitrary `suchThat` (\bs -> BS.length bs > 0)
-      bitOffset <- chooseInt (0, BS.length bs - 1)
+      bitOffset <- chooseInt (0, bitLength bs - 1)
       pure $ WithBitOffset bitOffset bs
   shrink (WithBitOffset bitOffset bs) =
       [ WithBitOffset bitOffset' bs'
       | bs' <- shrink bs
       , BS.length bs' > 0
-      , let bitOffset' = max 0 $ min (BS.length bs' - 1) bitOffset
+      , let bitOffset' = max 0 $ min (bitLength bs' - 1) bitOffset
       ] ++ [
         WithBitOffset bitOffset' bs
       | bitOffset' <- max 0 <$> shrink bitOffset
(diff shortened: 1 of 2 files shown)


test/Test/FS.hs
```diff
@@ -124,16 +124,19 @@ createFile hfs p = withFile hfs p (WriteMode MustBeNew) $ \_ -> pure ()
...
       bs <- arbitrary `suchThat` (\bs -> BS.length bs > 0)
-      bitOffset <- chooseInt (0, BS.length bs - 1)
+      bitOffset <- chooseInt (0, bitLength bs - 1)
       pure $ WithBitOffset bitOffset bs
   shrink (WithBitOffset bitOffset bs) =
       [ WithBitOffset bitOffset' bs'
       | bs' <- shrink bs
       , BS.length bs' > 0
-      , let bitOffset' = max 0 $ min (BS.length bs' - 1) bitOffset
+      , let bitOffset' = max 0 $ min (bitLength bs - 1) bitOffset
```
mheinzel, line 139: I think it should be
```suggestion
      , let bitOffset' = max 0 $ min (bitLength bs' - 1) bitOffset
```

mheinzel (commented):
Looks reasonable, but I think one small mistake snuck in.

test/Test/FS.hs
jorisdral: Good catch, fixed

jorisdral:
I've turned on auto-merge, so once this is approved it should move into the merge queue automatically

wenkokke (approved):

mheinzel (approved):

PR #555, opened 2025-01-29
feat: add corruptSnapshot to model

wenkokke:


test/Database/LSMTree/Model/Session.hs
```diff
@@ -280,6 +284,7 @@ instance Eq Err where
     where
       _coveredAllCases x = case x of
           ErrTableClosed{}          -> ()
+          ErrSnapshotCorrupted{}    -> ()
```
jorisdral, line 287: You didn't add a case `(==) ErrSnapshotCorrupted ErrSnapshotCorrupted`

jorisdral (changes requested):

test/Database/LSMTree/Model/Session.hs
wenkokke: Woops. Fixed.

wenkokke: What's the deal with `_coveredAllCases`?

test/Database/LSMTree/Model/Session.hs
```diff
@@ -575,9 +586,11 @@ openSnapshot label name = do
     case Map.lookup name snaps of
       Nothing ->
         throwError ErrSnapshotDoesNotExist
-      Just (Snapshot conf label' tbl) -> do
+      Just (Snapshot conf label' tbl corrupted) -> do
         when (label /= label') $
           throwError ErrSnapshotWrongType
+        when corrupted $
+          throwError ErrSnapshotCorrupted
```
jorisdral, line 593: I think the order should also be the other way around. The SUT will not be able to compare labels until it has read/deserialised/checksummed all the snapshot files

test/Database/LSMTree/Model/Session.hs
jorisdral: The `Eq` instance is custom so if `Err` changes, the compiler doesn't always tell you to modify the `Eq` instance. With `_coveredAllCases`, the code won't compile when you've added or changed constructors

jorisdral (approved):
Last thing, and then we can merge

test/Database/LSMTree/Model/Session.hs
jorisdral: So it's just a of future-proofing of the code

PR #556, opened 2025-01-29
Some state machine test fixes and refactorings

jorisdral:


test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -351,6 +340,14 @@ propLockstep_RealImpl_MockFS_IO tr =
         )
       tagFinalState'
 
+-- We can not use @bracket@ inside @PropertyM@, so @acuire_RealImpl_MockFS@ and
```
mheinzel, line 343:
```suggestion
-- We can not use @bracket@ inside @PropertyM@, so @acquire_RealImpl_MockFS@ and
```

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -351,6 +340,14 @@ propLockstep_RealImpl_MockFS_IO tr =
...
+-- @release_RealImpl_MockFS@ are not run in a maksed state and it is not
```
mheinzel, line 344:
```suggestion
-- @release_RealImpl_MockFS@ are not run in a masked state and it is not
```

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -440,48 +455,65 @@ realHandler = Handler $ pure . handler'
...
 
-createSystemTempDirectory ::  [Char] -> IO (FilePath, HasFS IO HandleIO, HasBlockIO IO HandleIO)
-createSystemTempDirectory prefix = do
-    systemTempDir <- getCanonicalTemporaryDirectory
-    tempDir <- createTempDirectory systemTempDir prefix
-    let hasFS = ioHasFS (MountPoint tempDir)
-    hasBlockIO <- ioHasBlockIO hasFS defaultIOCtxParams
-    pure (tempDir, hasFS, hasBlockIO)
+-- | When combined with other handlers, 'catchAllErrorHandler' has to go last
+-- because it matches on 'SomeException', an no other handlers are run after
```
mheinzel, line 512:
```suggestion
-- because it matches on 'SomeException', and no other handlers are run after
```

test/Database/LSMTree/Model/Session.hs
```diff
@@ -266,12 +271,31 @@ instance Show Err where
...
+--
+-- These laws should be sufficient for the purpose of comparing an error
+-- returned from the SUT to an error returned from the model.
+--
+-- Where @'Eq' 'Err'@ deviates from a typical instance is that 'ErrDiskFault' is
+-- equal to all other errors. This is important for comparing SUT exceptions to
+-- model exceptions. The model only knows /something/ went wrong, but the SUT
+-- will probably have a very specific error that it throws. In such cases where
+-- the SUT error is more specific then the model error, the errors are
+-- considered equal.
```
mheinzel, line 298: This unlawful instance seems quite hacky. Is using a plain function difficult because `quickcheck-lockstep` asks for an `Eq` instance?

mheinzel (approved):

test/Database/LSMTree/Model/Session.hs
jorisdral: The current approach piggybacks on the `quickcheck-lockstep` framework, so that we don't have to write a custom `postcondition`. 

The default postcondition is: `Either Err a == Either Err a` for some `a` that was the result of action. But with injected errors, we need approximate equality instead of syntactic equality, so I thought instead of augmenting the `postcondition` I'd just make `Eq Err` approximate.


'll try to see what it looks like if I use a plain function

jorisdral: So, using a plain function is not great because we'd have to rewrite the `quickcheck-lockstep` postcondition, but it would be doable. 

However, while trying out the plain function approach I re-discovered that we had used the `Eq (Obs h a)` in the past to allow disagreements between the model and SUT, instead of the `Eq Err` instance. In particular, we had already encoded in `Eq (Obs h a)` that the model invalidates blob references conservatively, while the SUT may keep them valid for a while longer. So, I moved the check into `Eq (Obs h a)` and that makes it all simpler

jorisdral: We could argue that `quickcheck-lockstep` should require just a comparison function, instead of requiring `Eq (Obs h a)`. But having a slightly unlawful `Eq (Obs h a)` achieves the same

PR #557, opened 2025-01-30
Bugfix: synchronise the snapshot directory and its contents

jorisdral:
Not all the snapshot directory contents were being synchronised, and arguably not in the correct order. The synchronisation now only happens at the very end of taking a snapshot. Because of this, if other parts of the `createSnapshot` function fail, then we won't have flushed file and directory contents to disk unnecessarily.


dcoutts (approved):
Yes this is a simpler approach: sync everything in the named snapshot dir once at the end, rather than doing bits as we go along (which is more intrusive in the code).

PR #564, opened 2025-02-06
feat(issue558): test silent snapshot corruption

### LINKED ISSUE #558: Snapshot corruption in the state machine tests
#539 introduced a classic property test for detecting corruption in snapshots. #542 fixed a bug in the code that #539 tested. The next step would be to detect corruption in the state machine tests as well. First, I'll describe the current approach to fault testing in the state machine tests, and then how to add snapshot corruption to the state machine tests.

### Overview of fault testing in the state machine

At the time of writing this issue (87549dd86d4b068cb6b145f809045f7f73be2af2), the state machine already has the possibility to inject *noisy* disk faults using `fs-sim` into the SUT. These faults are noisy because they produce IO exceptions. Even with fault injection, it is not guaranteed that the SUT will fail with an exception. There is non-determinism in the outcome. However, the model is deterministic, which means that the model must decide to fail with an exception, or to succeed. We solve this discrepancy as follows:

1. Fault injection is explicit: either we inject faults into the SUT, or we do not. This is represented by a `Maybe Errors` in relevant state machine actions.
2. If an action has faults to inject, the model assumes the SUT will throw an exception, and so the model will also throw an exception.

https://github.com/IntersectMBO/lsm-tree/blob/87549dd86d4b068cb6b145f809045f7f73be2af2/test/Database/LSMTree/Model/Session.hs#L202-L237

3. If an action has faults to inject, the SUT will run with those injected faults using `withErrors` from the `fs-sim` package.
  * If the SUT threw an exception, then the model and SUT both return an exception so we're okay.
  * If the SUT did not throw an exception, then it accidentally succeeded in performing the action. The model and SUT will not be in sync, so what we do is that we roll back the successful action. What to roll back differs per action. For example if `CreateSnapshot` accidentally succeeded, then the roll back action is to delete the snapshot. This makes sure the model and SUT are in sync again, and then we throw a dummy error so that the model and SUT have matching responses to the action.

https://github.com/IntersectMBO/lsm-tree/blob/87549dd86d4b068cb6b145f809045f7f73be2af2/test/Test/Database/LSMTree/StateMachine.hs#L1301-L1339

### Tasks

* (#555) The model has to be made aware of snapshot corruption, so that it can respond appropriately when opening a snapshot that is corrupt. Moreover, we have to add a function to the model that explicitly corrupts a snapshot. 
* Add support for corrupting snapshots explicitly in the SUT. There are two options, and I'd probably recommend going with option (i) because it's the most straightforward to do and requires minimal code changes. We can implement option (ii) afterwards, but then we'll already have all the necessary boilerplate in place.
  1. Use the `flipSnapshotBit` approach from #539. The code to corrupt the snapshot can be factored out from `prop_flipSnapshotBit`, and then we'd probably need to add a `corruptSnapshot` function to the `IsTable` class and implement it using `flipSnapshotBit`.
  2. Modify `fs-sim` to also perform *silent* corrupted writes instead of the *noisy* corrupted writes it does now. 
* Represent the instruction to corrupt a snapshot in the `CreateSnapshot` action.
  1. Something like `Maybe (Double, Double)` should be added, which describes which file and bit to corrupt in a snapshot
  2. A `Maybe Errors` with silent corruption instructions.
  Then make the appropriate changes to handle updated `CreateSnapshot` action. Stuff like modifying generators, modifying `runModel`, modifying `runIO` and `runIOSim`, and so forth.
* Finally, run the tests, make fixes to the testing framework, the model, or the SUT, and so forth.

wenkokke:
Fixes #558.

- [X] add `corruptSnapshot` to the model
- [X] add `corruptSnapshot` to the `IsTable` class
- [X] extract `flipRandomBitInRandomFile` from `prop_flipSnapshotBit`
- [x] implement `corruptSnapshot` for the SUT
- [x] add silent corruptions to the state machine tests

### DIFF
--- a/test/Database/LSMTree/Class.hs
+++ b/test/Database/LSMTree/Class.hs
@@ -28,7 +28,7 @@ import           Database.LSMTree.Class.Common as Common
 import qualified Database.LSMTree.Internal as RI (SessionEnv (..), Table (..),
                      Table' (..), withOpenSession)
 import qualified Database.LSMTree.Internal.Paths as RIP
-import           Test.Util.FS (flipRandomBitInRandomFile)
+import           Test.Util.FS (flipRandomBitInRandomFileHardlinkSafe)
 import           Test.Util.QC (Choice)
 
 -- | Class abstracting over table operations.
@@ -249,7 +249,7 @@ rCorruptSnapshot choice name (RI.Table' t) =
       let hfs = RI.sessionHasFS seshEnv
           root = RI.sessionRoot seshEnv
           namedSnapDir = RIP.getNamedSnapshotDir (RIP.namedSnapshotDir root name)
-       in void $ flipRandomBitInRandomFile hfs choice namedSnapDir
+       in void $ flipRandomBitInRandomFileHardlinkSafe hfs choice namedSnapDir
 
 instance IsTable R.Table where
     type Session R.Table = R.Session
--- a/test/Database/LSMTree/Model/Session.hs
+++ b/test/Database/LSMTree/Model/Session.hs
@@ -32,7 +32,7 @@ module Database.LSMTree.Model.Session (
   , runModelM
   , runModelMWithInjectedErrors
     -- ** Errors
-  , Err (.., DefaultErrDiskFault)
+  , Err (.., DefaultErrDiskFault, DefaultErrSnapshotCorrupted)
     -- * Tables
   , Table
   , TableConfig (..)
@@ -241,19 +241,23 @@ runModelMWithInjectedErrors (Just _) _ onErrors st =
 pattern DefaultErrDiskFault :: Err
 pattern DefaultErrDiskFault = ErrDiskFault "default"
 
+-- | The default 'ErrSnapshotCorrupted' that model operations will throw.
+pattern DefaultErrSnapshotCorrupted :: Err
+pattern DefaultErrSnapshotCorrupted = ErrSnapshotCorrupted "default"
+
 --
 -- Errors
 --
 
 data Err =
     ErrTableClosed
-  | ErrSnapshotCorrupted
   | ErrSnapshotExists
   | ErrSnapshotDoesNotExist
   | ErrSnapshotWrongType
   | ErrBlobRefInvalidated
   | ErrCursorClosed
     -- | Something went wrong with the file system.
+  | ErrSnapshotCorrupted String
   | ErrDiskFault String
   | ErrOther String
   deriving stock Eq
@@ -262,8 +266,6 @@ instance Show Err where
   showsPrec d = \case
       ErrTableClosed ->
         showString "ErrTableClosed"
-      ErrSnapshotCorrupted ->
-        showString "ErrSnapshotCorrupted"
       ErrSnapshotExists ->
         showString "ErrSnapshotExists"
       ErrSnapshotDoesNotExist ->
@@ -274,6 +276,10 @@ instance Show Err where
         showString "ErrBlobRefInvalidated"
       ErrCursorClosed ->
         showString "ErrCursorClosed"
+      ErrSnapshotCorrupted s ->
+        showParen (d > appPrec) $
+        showString "ErrSnapshotCorrupted " .
+        showParen True (showString s)
       ErrDiskFault s ->
         showParen (d > appPrec) $
         showString "ErrDiskFault " .
@@ -577,7 +583,7 @@ openSnapshot label name = do
         throwError ErrSnapshotDoesNotExist
       Just (Snapshot conf label' tbl corrupted) -> do
         when corrupted $
-          throwError ErrSnapshotCorrupted
+          throwError DefaultErrSnapshotCorrupted
         when (label /= label') $
           throwError ErrSnapshotWrongType
         case fromSomeTable tbl of
@@ -592,9 +598,9 @@ openSnapshot label name = do
           Just table' ->
             newTableWith conf table'
 
--- TODO: to match the implementation of the real table, this should not corrupt the
---       snapshot if there are _no non-empty files_; however, since there are no such
---       snapshots, this is probably fine
+-- To match the implementation of the real table, this should not corrupt the
+-- snapshot if there are _no non-empty files_; however, since there are no such
+-- snapshots, this is probably fine.
 corruptSnapshot ::
      (MonadState Model m, MonadError Err m)
   => SnapshotName
--- a/test/Test/Database/LSMTree/StateMachine.hs
+++ b/test/Test/Database/LSMTree/StateMachine.hs
@@ -66,8 +66,9 @@ module Test.Database.LSMTree.StateMachine (
   ) where
 
 import           Control.ActionRegistry (AbortActionRegistryError (..),
-                     AbortActionRegistryReason (..), ActionError,
-                     CommitActionRegistryError (..), getActionError)
+                     ActionError, CommitActionRegistryError (..),
+                     getActionError, getReasonExitCaseException)
+import           Control.Applicative (Alternative (..))
 import           Control.Concurrent.Class.MonadMVar.Strict
 import           Control.Concurrent.Class.MonadSTM.Strict
 import qualified Control.Exception
@@ -83,13 +84,13 @@ import           Control.Tracer (Tracer, nullTracer)
 import           Data.Bifunctor (Bifunctor (..))
 import           Data.Constraint (Dict (..))
 import           Data.Either (partitionEithers)
-import           Data.Foldable (for_)
 import           Data.Kind (Type)
 import           Data.List.NonEmpty (NonEmpty (..))
 import qualified Data.List.NonEmpty as NE
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import           Data.Maybe (catMaybes, fromJust, fromMaybe)
+import           Data.Monoid (First (..))
 import           Data.Primitive.MutVar
 import           Data.Set (Set)
 import qualified Data.Set as Set
@@ -103,7 +104,8 @@ import           Database.LSMTree.Extras.Generators (KeyForIndexCompact)
 import           Database.LSMTree.Extras.NoThunks (propNoThunks)
 import           Database.LSMTree.Internal (LSMTreeError (..))
 import qualified Database.LSMTree.Internal as R.Internal
-import           Database.LSMTree.Internal.CRC32C (FileFormatError (..))
+import           Database.LSMTree.Internal.CRC32C (ChecksumError (..),
+                     ChecksumsFileFormatError (..), FileFormatError (..))
 import           Database.LSMTree.Internal.Serialise (SerialisedBlob,
                      SerialisedValue)
 import qualified Database.LSMTree.Model.IO as ModelIO
@@ -137,10 +139,10 @@ import qualified Test.QuickCheck.StateModel.Lockstep.Defaults as Lockstep.Defaul
 import qualified Test.QuickCheck.StateModel.Lockstep.Run as Lockstep.Run
 import           Test.Tasty (TestTree, testGroup)
 import           Test.Tasty.QuickCheck (testProperty)
-import           Test.Util.FS (SilentCorruption (..), SilentCorruptions (..),
-                     approximateEqStream, noRemoveDirectoryRecursiveE,
+import           Test.Util.FS (approximateEqStream, noRemoveDirectoryRecursiveE,
                      propNoOpenHandles, propNumOpenHandles)
 import           Test.Util.PrettyProxy
+import           Test.Util.QC (Choice)
 import qualified Test.Util.QLS as QLS
 import           Test.Util.TypeFamilyWrappers (WrapBlob (..), WrapBlobRef (..),
                      WrapCursor (..), WrapTable (..))
@@ -444,6 +446,8 @@ realErrorHandlers = [
     , abortActionRegistryErrorHandler
     , fsErrorHandler
     , fileFormatErrorHandler
+    , checksumsFileFormatErrorHandler
+    , checksumErrorHandler
     , catchAllErrorHandler
     ]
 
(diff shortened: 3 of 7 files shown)


test/Test/Util/QC.hs
```diff
@@ -15,3 +18,16 @@ testClassLaws typename Laws {lawsTypeclass, lawsProperties} =
     | (name, prop) <- lawsProperties ]
 
 
+-- | A 'Choice' of a uniform random number in a range where shrinking picks smaller numbers.
+newtype Choice = Choice Word64
+  deriving stock (Show, Eq)
+  deriving newtype (Arbitrary)
```
jorisdral: The `Arbitrary` instance for `Word64` uses [`arbitrarySizedBoundedIntegral`](https://hackage.haskell.org/package/QuickCheck-2.15.0.1/docs/Test-QuickCheck.html#v:arbitrarySizedBoundedIntegral), which generates smaller numbers more often than big numbers. Maybe we should use [`arbitraryBoundedIntegral`](https://hackage.haskell.org/package/QuickCheck-2.15.0.1/docs/Test-QuickCheck.html#v:arbitraryBoundedIntegral) instead? That would generate uniformly distributed `Word64`s, which is arguably better I think, because then each file bit is equally likely to get picked

Another argument in favour of an explicit instance is that the `Arbitrary` instance can't change under our feet in the future, which would happen currently if the `Arbitrary` instance for `Word64` changes in a new version of `QuickCheck`

test/Test/Util/QC.hs
```diff
@@ -15,3 +18,16 @@ testClassLaws typename Laws {lawsTypeclass, lawsProperties} =
...
+
+getChoice :: (Integral a) => Choice -> (a, a) -> a
+getChoice (Choice n) (l, u) = fromIntegral (((ni * (ui - li)) `div` mi) + li)
+  where
+    ni = toInteger n
+    li = toInteger l
+    ui = toInteger u
+    mi = toInteger (maxBound :: Word64)
```
jorisdral, line 35: In `(a, a)`, are these inclusive or exclusive bounds? We should mention this in the haddocks

test/Test/Util/QC.hs
jorisdral: This is the current output for `prop_flipSnapshotBit`. I'd hope that we flip other bits than `0` more often if we make the suggested change

```
    prop_flipSnapshotBit:                   OK (0.32s)
      +++ OK, passed 100 tests.

      Corrupted file (100 in total):
      46% snapshots/snap/0.checksums
      29% snapshots/snap/0.blobs
      18% snapshots/snap/0.keyops
       4% snapshots/snap/1.keyops
       2% snapshots/snap/2.keyops
       1% snapshots/snap/3.keyops

      Flipped bit (100 in total):
      63% n == 0
      14% 10000 <= n < 100000
      11% 1000 <= n < 10000
       6% 10 <= n < 100
       4% 100 <= n < 1000
       2% 1 <= n < 10

      Result (100 in total):
      49% Corruption detected: ChecksumError
      42% Corruption detected: ChecksumsFileFormatError
       7% Corruption detected: AbortActionRegistryError
       2% Corruption detected: FileFormatError
 ```

File: test/Test/Util/QC.hs [hunk 3]
```diff
@@ -15,3 +18,16 @@ testClassLaws typename Laws {lawsTypeclass, lawsProperties} =
     | (name, prop) <- lawsProperties ]
 
 
+-- | A 'Choice' of a uniform random number in a range where shrinking picks smaller numbers.
+newtype Choice = Choice Word64
+  deriving stock (Show, Eq)
+  deriving newtype (Arbitrary)
+
+getChoice :: (Integral a) => Choice -> (a, a) -> a
+getChoice (Choice n) (l, u) = fromIntegral (((ni * (ui - li)) `div` mi) + li)
```
jorisdral, line 30: Neat function :+1:

jorisdral, line 30: Can we expect for all `Integral a` that the `fromIntegral` conversion preserves uniformity? We should maybe add a comment about it

test/Test/Util/FS.hs
```diff
@@ -347,6 +351,43 @@ listDirectoryFiles hfs = go Set.empty
...
+    -- Handle the situation where there are no non-empty files
+    if totalFileSizeBits == 0 then pure Nothing else do
+      assert (totalFileSizeBits > 0) $ pure ()
+      -- Interpret `index` to point to a bit between `0` and `totalFileSize - 1`
+      let bitIndex = getChoice bitChoice (0, totalFileSizeBits - 1)
+      -- Flip the bit
+      Just <$> flipFileBitAt hfs bitIndex filesAndFileSizeBits
+
+-- | Internal helper: flip a single bit in a given list of files and sizes.
+flipFileBitAt :: (MonadThrow m, PrimMonad m) => HasFS m h -> Word64 -> [(FsPath, Word64)] -> m (FsPath, Int)
```
jorisdral, line 390:
```suggestion
flipFileBitAt :: 
     (MonadThrow m, PrimMonad m) 
  => HasFS m h 
  -> Word64 
  -> [(FsPath, Word64)] 
  -> m (FsPath, Int)
```

test/Test/Util/FS.hs
```diff
@@ -347,6 +351,43 @@ listDirectoryFiles hfs = go Set.empty
...
+flipFileBitAt _hfs bitIndex [] =
+  error $ printf "flipFileBitAt: bit index out of bounds (%d)" bitIndex
+flipFileBitAt hfs bitIndex ((file, fileSize) : filesAndSizes)
+  | bitIndex < fileSize = do
+    let bitIndexAsInt = fromIntegral bitIndex
```
jorisdral, line 395: This could technically overflow (though highly unlikely). Maybe for correctness clamp the result to the `(0, maxBound)` range?

test/Test/Util/FS.hs
```diff
@@ -347,6 +351,43 @@ listDirectoryFiles hfs = go Set.empty
...
+    flipFileBit hfs file bitIndexAsInt
+    pure (file, bitIndexAsInt)
+  | otherwise = do
```
jorisdral, line 398: Redundant do

test/Database/LSMTree/Model/Session.hs
```diff
@@ -603,6 +603,9 @@ openSnapshot label name = do
           Just table' ->
             newTableWith conf table'
 
+-- TODO: to match the implementation of the real table, this should not corrupt the
+--       snapshot if there are _no non-empty files_; however, since there are no such
+--       snapshots, this is probably fine
```
jorisdral, line 609: Convert into a regular comment instead of a TODO. I don't think this is something we'll have to change in the future.

test/Database/LSMTree/Class.hs
```diff
@@ -231,6 +236,21 @@ withCursor offset tbl = bracket (newCursor offset tbl) (closeCursor (Proxy @h))
   Real instance
 -------------------------------------------------------------------------------}
 
+-- | Snapshot corruption for the real instance.
+--   Implemented here, instead of as part of the public API.
```
jorisdral, line 240: :+1:

test/Database/LSMTree/Model/Session.hs
```diff
@@ -233,9 +233,10 @@ runModelMWithInjectedErrors ::
   -> Model -> (Either Err a, Model)
 runModelMWithInjectedErrors Nothing onNoErrors _ st =
     runModelM onNoErrors st
-runModelMWithInjectedErrors (Just _) _ onErrors st =
+runModelMWithInjectedErrors (Just _errors) _ onErrors st =
     runModelM (onErrors >> throwError (ErrFsError "modelled FsError")) st
 
+
```
jorisdral, line 239:
```suggestion
```

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -1211,10 +1214,18 @@ runIO action lookUp = ReaderT $ \ !env -> do
...
-          runRealWithInjectedErrors "CreateSnapshot" env merrs
-            (Class.createSnapshot label name (unwrapTable $ lookUp' tableVar))
-            (\() -> Class.deleteSnapshot session name)
+        CreateSnapshot mcorrsOrErrs label name tableVar ->
+          let rCreateSnapshot = Class.createSnapshot label name (unwrapTable $ lookUp' tableVar) in
+          case sequence mcorrsOrErrs of
+            Left (SilentCorruptions corrs) -> do
+              rCreateSnapshot
+              for_ corrs $ \corr ->
+                Class.corruptSnapshot (bitChoice corr) name (unwrapTable $ lookUp' tableVar)
```
jorisdral, line 1223: We could technically flip a bit we previously flipped, un-corrupting the snapshot again. It's verrrry unlikely, but it might deserve a comment. What do you think?

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -1273,10 +1284,18 @@ runIOSim action lookUp = ReaderT $ \ !env -> do
           Class.mupserts (unwrapTable $ lookUp' tableVar) kmups
         RetrieveBlobs blobRefsVar -> catchErr handlers $
           fmap WrapBlob <$> Class.retrieveBlobs (Proxy @h) session (unwrapBlobRef <$> lookUp' blobRefsVar)
-        CreateSnapshot merrs label name tableVar ->
-          runRealWithInjectedErrors "CreateSnapshot" env merrs
-            (Class.createSnapshot label name (unwrapTable $ lookUp' tableVar))
-            (\() -> Class.deleteSnapshot session name)
+        CreateSnapshot mcorrsOrErrs label name tableVar ->
```
jorisdral, line 1287: The generator for `CreateSnapshot` generates `Nothing` by default, so we're currently not running tests with coruption at all

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -604,7 +606,7 @@ instance ( Show (Class.TableConfig h)
     -- Snapshots
     CreateSnapshot ::
          C k v b
-      => Maybe Errors
+      => Maybe (Either SilentCorruptions Errors)
       -> R.SnapshotLabel -> R.SnapshotName -> Var h (WrapTable h IO k v b)
       -> Act h ()
```
jorisdral, line 611: Would be good to add a `Tag` for snapshot corruption so we can see from the test output that corruption happens (or not)

jorisdral (changes requested):
Nice! Here are some comments. In particular, the tests don't seem to run with corruption just yet because the actions generator was not updated yet.

BTW, I pushed ec1061dc6c9363fcadafd4e803db75cec78e3d7b to the branch, which includes some changes I made while reviewing the PR. Feel free to remove the commit and/or incorporate some of the suggestions there

jorisdral:
With my changes from https://github.com/IntersectMBO/lsm-tree/commit/ec1061dc6c9363fcadafd4e803db75cec78e3d7b, I see this failure:

```
✦ ❯ cabal run lsm-tree-test -- -p '$NF=="propLockstep_RealImpl_MockFS_IO"'
Configuration is affected by the following files:
- cabal.project.debug
- cabal.project.local
lsm-tree
  Test.Database.LSMTree.StateMachine
    propLockstep_RealImpl_MockFS_IO: FAIL (4.43s)
      *** Failed! Exception: 'ChecksumsFileFormatError snapshots/snap2/0.checksums "BRC32C (blobs) = 00000000"' (after 20 tests and 111 shrinks):
      do var1 <- action $ New (PrettyProxy @((Key,Value,Blob))) (TableConfig {confMergePolicy = MergePolicyLazyLevelling, confSizeRatio = Four, confWriteBufferAlloc = AllocNumEntries (NumEntries 1), confBloomFilterAlloc = AllocFixed 10, confFencePointerIndex = CompactIndex, confDiskCachePolicy = DiskCacheNone, confMergeSchedule = Incremental})
         action $ CreateSnapshot (Just (Left (SilentCorruptions {unSilentCorruptions = SilentCorruption {bitChoice = Choice 0} :| []}))) (SnapshotLabel "Key Value Blob") "snap2" (unsafeMkGVar var1 (OpFromRight `OpComp` OpId))
         action $ OpenSnapshot (PrettyProxy @((Key,Value,Blob))) Nothing (SnapshotLabel "Key Value Blob") "snap2"
         pure ()
      Use --quickcheck-replay="(SMGen 7156249841088144899 16260318356288160751,19)" to reproduce.

1 out of 1 tests failed (4.43s)
```

There are two parts to solving this: making sure `ChecksumsFileFormatError` is caught by the handlers, and making sure that it is mapped to the correct model `Err`, which I think is `ErrSnapshotCorrupted`

I'm also trying to merge #556 today, which refactors the error handlers a bit, which may conflict with this PR

test/Test/Util/QC.hs
wenkokke: Inclusive—I'll add a Haddock.

test/Test/Util/QC.hs
wenkokke: Right, it's uniform as long as the range isn't larger than the largest `Word64`.

test/Database/LSMTree/Model/Session.hs
wenkokke: This feels like the kind of change we should leave up to formatters, rather than people.

test/Test/Database/LSMTree/StateMachine.hs
wenkokke: Ugh, yes. Fine, I'll `Set` it! ^__^"

test/Test/Database/LSMTree/StateMachine.hs
wenkokke: I see you fixed this.

wenkokke:
Remaining test failure under:

```bash
cabal run lsm-tree-test -- --quickcheck-replay="(SMGen 10878425437647127804 1790301305950248185,70)" -p '/propLockstep_RealImpl_RealFS_IO/'
```

Gives the error:

```
lsm-tree
  Test.Database.LSMTree.StateMachine
    propLockstep_RealImpl_RealFS_IO: FAIL (12.22s)
      *** Failed! Falsified (after 1 test and 97 shrinks):
      do pure ()
      Reference is never released: RefId 27368
      Allocation site: CallStack (from HasCallStack):
        newRefWithTracker, called at src-control/Control/RefCount.hs:309:5 in lsm-tree-0.1.0.0-inplace-control:Control.RefCount
        dupRef, called at src/Database/LSMTree/Internal/RunReader.hs:121:23 in lsm-tree-0.1.0.0-inplace:Database.LSMTree.Internal.RunReader
      Use --quickcheck-replay="(SMGen 10878425437647127804 1790301305950248185,70)" to reproduce.
```

test/Test/Util/FS.hs
```diff
@@ -343,10 +352,114 @@ listDirectoryFiles hfs = go Set.empty
         else
           pure acc
 
+{-------------------------------------------------------------------------------
+  Hard Links
+-------------------------------------------------------------------------------}
+
```
jorisdral, line 358:
```suggestion
```

test/Test/Util/FS.hs
```diff
@@ -343,10 +352,114 @@ listDirectoryFiles hfs = go Set.empty
...
+-- | Pick a random bit in a random file in a given directory.
+pickRandomBitInRandomFile ::
+     (PrimMonad m, MonadThrow m)
+  => HasFS m h
+  -> Choice
+  -> FsPath
+  -> m (Maybe (FsPath, Int))
+pickRandomBitInRandomFile hfs bitChoice dir = do
+  -- List all files and their sizes
```
jorisdral, line 407:
```suggestion
  -- List all files
```

test/Test/Util/FS.hs
```diff
@@ -343,10 +352,114 @@ listDirectoryFiles hfs = go Set.empty
...
+      assert (totalFileSizeBits > 0) $ pure ()
+      -- Internal helper: find the file/bit that a choice points to.
+      let pickFileBitAt bitIndex [] =
+            error $ printf "flipFileBitAt: bit index out of bounds (%d)" bitIndex
+          pickFileBitAt bitIndex ((file, fileSize) : filesAndSizes)
+            | bitIndex < fileSize = pure (file, fromIntegral $ bitIndex `min` fromIntegral (maxBound @Int))
+            | otherwise = pickFileBitAt (bitIndex - fileSize) filesAndSizes
+      -- Interpret `index` to point to a bit between `0` and `totalFileSize - 1`
+      let bitIndex = getChoice bitChoice (0, totalFileSizeBits - 1)
+      -- Flip the bit
```
jorisdral, line 427:
```suggestion
```

test/Test/Util/FS.hs
```diff
@@ -343,10 +352,114 @@ listDirectoryFiles hfs = go Set.empty
...
+  -> FsPath
+  -> Int -- ^ Bit offset.
+  -> m ()
+flipFileBitHardlinkSafe hfs hbio fileOrig bitOffset = do
+  -- Compute the CRC of fileOrig:
+  let readHandleCRC !crc !h =
+        hGetSome hfs h 65504 >>= \case
+                    bs | BS.null bs -> pure crc
+                       | otherwise  -> readHandleCRC (CRC.crc32c_update crc bs) h
+  crc <- withFile hfs fileOrig ReadMode (readHandleCRC 0)
```
jorisdral, line 444: Why compute the CRC? Seems redundant

jorisdral (changes requested):
Almost ready. Just a few more small changes to make, and then we can proceed

test/Test/Util/FS.hs
```diff
@@ -423,7 +536,8 @@ fsPathComponentString :: FsPathComponent -> String
 fsPathComponentString (FsPathComponent s) = NE.toList s
 
 instance Arbitrary FsPathComponent where
-  arbitrary = resize 5 $ -- path components don't have to be very long
+  arbitrary = resize 5 $ -- path components don't have to be very long -- path components don't have to be very long
+       -- path components don't have to be very long
```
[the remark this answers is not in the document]
wenkokke: I think this is a stylish Haskell artifact.

test/Test/Util/FS.hs
```diff
@@ -459,7 +549,8 @@ instance Arbitrary NoCleanupErrors where
 -------------------------------------------------------------------------------}
 
 instance Arbitrary FsPath where
-  arbitrary = scale (`div` 10) $ -- paths don't have to be very long
+  arbitrary = scale (`div` 10) $ -- paths don't have to be very long -- paths don't have to be very long
+       -- paths don't have to be very long
```
jorisdral, line 553:
```suggestion
  arbitrary = scale (`div` 10) $ -- paths don't have to be very long 
```

test/Test/Util/FS.hs
```diff
@@ -342,11 +350,92 @@ listDirectoryFiles hfs = go Set.empty
...
+  -> FsPath
+  -> Int -- ^ Bit offset.
+  -> m ()
+flipFileBitHardlinkSafe hfs fileOrig bitOffset = do
+  -- Compute the CRC of fileOrig:
+  let readHandleCRC !crc !h =
+        hGetSome hfs h 65504 >>= \case
+                    bs | BS.null bs -> pure crc
+                       | otherwise  -> readHandleCRC (CRC.crc32c_update crc bs) h
+  crc <- withFile hfs fileOrig ReadMode (readHandleCRC 0)
```
jorisdral, line 424: Okay, if we keep the CRC, we can probably use this function instead
```suggestion
  crc <- readFileCRC32C hfs fileOrig
```

jorisdral (approved):
LGTM! I have two small suggestions + we should rewrite/squash the commit history, and then we can merge

test/Test/Util/FS.hs
wenkokke: This is again a stylish-haskell issue.

PR #567, opened 2025-02-11
Enable version updates for GitHub Actions using Dependabot

jorisdral:
Dependabot will automatically open PRs if some dependencies in the GHA workflow files need updating.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  # Enable version updates for GitHub Actions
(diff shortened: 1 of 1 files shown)


wenkokke (approved):

wenkokke:
LGTM. Do we want to set an explicit time of week, so they don't come in over the weekend?

jorisdral:
> LGTM. Do we want to set an explicit time of week, so they don't come in over the weekend?

Good point, I checked and the default is on Monday, which should be okay

PR #573, opened 2025-02-13
Add type parameter to MergingRun to limit possible merge types

mheinzel:
# Description

This is just a phantom type parameter, the `Merge` inside still stores a general `MergeType` which is not statically guaranteed to be part of the restricted type. We can also go all the way, but then we get the type parameter and `IsMergeType` constraints in a lot of places, e.g. on `Merge.steps` and the `NoThunks` instances (which get quite a bit more cumbersome). So I decided against it for now, but I'm happy to revisit that choice.

Snapshots can always encode the full `MergeType`. At first I decided to do the partial conversion in `fromSnapIncomingRun`, potentially throwing an exception (first commit). Later I switched it to failing in `decodeVersioned`, which makes the serialisation code less mechanical, but seems to fit into our error handling better. I'm open to both solutions, so let me know if you have a preference.


src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -734,7 +735,8 @@ addRunToLevels tr conf@TableConfig{..} resolve hfs hbio root uc r0 reg levels ul
             !alloc = bloomFilterAllocForLevel conf ln
             !runPaths = Paths.runPath root (uniqueToRunNumber n)
         traceWith tr $ AtLevel ln $
-          TraceNewMerge (V.map Run.size rs) (runNumber runPaths) caching alloc mergePolicy mergeType
+          TraceNewMerge (V.map Run.size rs) (runNumber runPaths) caching alloc
+          mergePolicy mergeType
```
jorisdral, line 739: Indentation

src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -159,16 +157,16 @@ newtype SuppliedCredits = SuppliedCredits { getSuppliedCredits :: Int }
...
-  | SnapOngoingMerge !(V.Vector r) !MergeType
-    -- ^ While 'MR.MergingRun' can statically restrict the type of merge that
-    -- is possible in a specific part of a table (see 'MR.IsMergeType'), the
-    -- serialisation format always allows all of them. This is more future proof
-    -- and avoids accidentally deserialising one merge type as the other.
+  | SnapOngoingMerge !(V.Vector r) !t
+    -- ^ While we use a specific, more restrictive merge type here (see
+    -- 'MR.IsMergeType'), we should always serialise it as a 'MergeType'.
+    -- This is more future proof and avoids accidentally deserialising one merge
+    -- type as the other.
```
jorisdral, line 166: Alright, I see you did the parameterisation here. I don't think it's necessarily more future proof though to always serialise it as a `MergeType`. Why not enforce level merge vs. tree merge in the serialisation as well?

src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -216,11 +214,26 @@ toSnapIncomingRun (Merging mergePolicy mergingRun) = do
...
+  -> SnapMergingRunState MR.LevelMergeType (Ref (Run m h))
 toSnapMergingRunState (MR.CompletedMerge r)  = SnapCompletedMerge r
-toSnapMergingRunState (MR.OngoingMerge rs m) = SnapOngoingMerge rs (Merge.mergeType m)
+toSnapMergingRunState (MR.OngoingMerge rs m) =
+    -- The merge type conversion should never fail, since the Merge was
+    -- constructed based on the restricted LevelMergeType.
+    -- To make the types align, we could either:
+    -- * store the restricted type inside MergingRun itself, but this would
+    --   just duplicate information, not provide any safety.
+    -- * Also add the type parameter to Merge, adding complexity there.
```
jorisdral, line 232: Does it really make the code more complex, or is it a noisy change? If it's mostly noise, I think we should/could go for it

src/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -389,6 +388,7 @@ iforLevelM_ lvls k = V.iforM_ lvls $ \i lvl -> k (LevelNo (i + 1)) lvl
 -- * not stored in snapshots
 -- * not loaded from snapshots
 -- * ignored in lookups
+-- * never merged (by supplying credits to it)
```
jorisdral, line 391: Maybe:
```suggestion
-- * never completed (by supplying credits to it)
```

src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -161,6 +162,10 @@ newtype SuppliedCredits = SuppliedCredits { getSuppliedCredits :: Int }
 data SnapMergingRunState r =
     SnapCompletedMerge !r
   | SnapOngoingMerge !(V.Vector r) !MergeType
+    -- ^ While 'MR.MergingRun' can statically restrict the type of merge that
+    -- is possible in a specific part of a table (see 'MR.IsMergeType'), the
+    -- serialisation format always allows all of them. This is more future proof
+    -- and avoids accidentally deserialising one merge type as the other.
```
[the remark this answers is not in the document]
jorisdral: I see you revisited this in a later commit

src/Database/LSMTree/Internal/Snapshot.hs
```diff
@@ -407,6 +412,10 @@ releaseRuns reg = traverse_ $ \r -> delayedCommit reg (releaseRef r)
   Opening from levels snapshot format
 -------------------------------------------------------------------------------}
 
+data FromSnapError = InvalidLevelMergeType !MergeType
+  deriving stock Show
+  deriving anyclass Exception
```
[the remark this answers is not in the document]
jorisdral: I see you revisited this in a later commit

jorisdral (commented):
Nice nice, some suggestions/questions

src/Database/LSMTree/Internal/MergingRun.hs
```diff
@@ -66,7 +67,7 @@ import           Database.LSMTree.Internal.RunAcc (RunBloomFilterAlloc)
 import           System.FS.API (HasFS)
 import           System.FS.BlockIO.API (HasBlockIO)
 
-data MergingRun m h = MergingRun {
+data MergingRun t m h = MergingRun {
```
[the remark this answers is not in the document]
mheinzel: My thinking was that now the partially applied `MergingRun LevelMergeType` is exactly what it was before generalising for unions. But good point, both ways of thinking about it make sense.

src/Database/LSMTree/Internal/MergingTree.hs
```diff
@@ -130,11 +128,12 @@ newPendingLevelMerge prs mmt = do
     -- are masked then there can be no async exceptions here at all.
     mergeTreeState <- case (prs, mmt) of
       ([PreExistingRun r], Nothing) ->
+        -- If there is just a single PreExistingMergingRun, we could in theory
+        -- also directly turn that into a MergingTree, but it's not necessary
+        -- and a little complicated because of the LevelMergeType/TreeMergeType
+        -- mismatch.
```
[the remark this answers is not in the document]
mheinzel: `MergeLastLevel :: LevelMergeType` and `MergeLevel :: TreeMergeType` are semantically the same, so some kind of coercion would be fine. But turning a  `MergeMidLevel` merge into a `MergeLevel` one is dishonest, even though the unexpected Deletes in the output wouldn't cause a problem in practice. It could be done, but simply creating a one-input pending merge is not really a problem.

Thinking about it, if there is only a single `PreExistingRun`, then there can only be one level in the input table, and at level 1 there are no merging runs. So the whole case should never come up. Maybe that argument should be in the comment instead?

src/Database/LSMTree/Internal/Snapshot.hs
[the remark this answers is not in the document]
mheinzel: The point remains: The on-disk format should always talk about the full `MergeType`. Otherwise I see a risk of encoding `MergeLastLevel :: LevelMergeType` as 1 and later decoding that as `MergeUnion :: TreeMergeType` if we ever make some mistake. If serialisation goes via `MergeType`, we can detect the mismatch.

src/Database/LSMTree/Internal/Snapshot.hs
mheinzel: https://github.com/IntersectMBO/lsm-tree/pull/573#discussion_r1958175190

src/Database/LSMTree/Internal/Snapshot.hs
mheinzel: It's mainly noisy, and we also get more indirection in `Merge.steps` (an additional type class dictionary call for every entry we merge). That's probably okay, but not sure how it impacts other optimisations.

src/Database/LSMTree/Internal/MergeSchedule.hs
mheinzel: I don't just mean full completion, but doing any merging work in the tree. So maybe "progressed"?

src/Database/LSMTree/Internal/Snapshot.hs
mheinzel: > I don't think it's necessarily more future proof though to always serialise it as a MergeType.

Do you mean future proof as in "easy to make changes while keeping the serialisation format compatible" or "less risk of introducing a bug"?

It's safer than serialising several different merge types and requiring anyone reading them to exactly use the same type there as well. If there is just one `MergeType` with a `DecodeVersioned` instance, it's clear which one to use and the compiler checks the conversion to more specific types. It also makes it easy to e.g. relax the merge type invariant in the future without having to touch the serialisation format.

> Why not enforce level merge vs. tree merge in the serialisation as well?

Do you mean encoding our expectation of `LevelMergeType` vs `TreeMergeType`, so one can't be deserialised as the other? This could be done by reflecting the type parameter into an additional value `MergeTypeType = Level | Tree` and also serialising that. Or just tweaking the serialisation of the types themselves:

```haskell
-- | Ensure 'LevelMergeType' and 'TreeMergeType' can't be deserialised as the other.
tagLevelMergeType, tagTreeMergeType :: Word
tagLevelMergeType = 0
tagTreeMergeType  = 1

instance Encode LevelMergeType  where
  encode MergeMidLevel  = encodeTag tagLevelMergeType <> encodeWord 0
  encode MergeLastLevel = encodeTag tagTreeMergeType  <> encodeWord 1

instance DecodeVersioned LevelMergeType where
  decodeVersioned V0 = do
      tag <- decodeTag
      when (tag /= tagLevelMergeType) $
        fail _
      decodeWord >>= \case
        0 -> pure MergeMidLevel
        1 -> pure MergeLastLevel
        t -> fail ("[LevelMergeType] Unexpected tag: " <> show t)
```

Or what do you have in mind?

src/Database/LSMTree/Internal/Snapshot.hs
jorisdral: You can precalculate `t -> MergeType` and store the result somewhere, no need to compute it on every call to `Merge.steps`

src/Database/LSMTree/Internal/Snapshot.hs
jorisdral: > Do you mean future proof as in "easy to make changes while keeping the serialisation format compatible" or "less risk of introducing a bug"?

Probably more along the lines of "less risk of introducing a bug". 

> ... relax the merge type invariant ...

What is the merge type invariant?

> Do you mean encoding our expectation of LevelMergeType vs TreeMergeType, so one can't be deserialised as the other?

This is kind of what I had in mind -- we serialise extra stuff to ensure we use the correct encoder/decoder

But okay, I can see where you're coming from. By serialising to a `MergeType`, we can change `LevelMergeType` and `TreeMergeType` and the mapping between those and `MergeType`, without having to change the serialisation format. Sounds good

jorisdral: Something I realised is that I prefer an explicit format because it might prevent getting bad tables from opening snapshots (i.e., tables with inconsistent state). However, maybe getting bad tables from opening old snapshots is something we should solve separately from the snapshot format. 

What I'm worried about is that we could have a newer version of `lsm-tree` that opens an older snapshot, but then even if the format was still compatible the resulting table can be in a bad state. One example might be that in an old `lsm-tree` we did some `x` amount of merging work, and then created a snapshot. But when we open the same snapshot in a newer version of `lsm-tree` and that version assumes that there was `x < y` amount of merging work, which can lead to failures down the line. In general, if the `lsm-tree` logic changes, how do we decide which snapshot version is still compatible?

This may seem a bit vague, we can also talk about it in call or something at some point :yum:

jorisdral (approved):

src/Database/LSMTree/Internal/MergingTree.hs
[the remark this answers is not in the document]
mheinzel: Yes. I'll do that.

src/Database/LSMTree/Internal/Snapshot.hs
mheinzel: > What is the merge type invariant?

It's what `LevelMergeType` and `TreeMergeType` enforce: In some places we can't have union merges and in some we can't have mid-level merges.

> But when we open the same snapshot in a newer version of lsm-tree and that version assumes that there was y amount of merging work

We can't really enforce all of the assumptions statically, but we can hopefully find good solution on a case-by-case basis. Some might be recoverable relatively easily (e.g. calculating from the tree shape how far along each merge should be and performing any remaining work), some might require making the merging algorithm a little more lenient. Keeping two merge algorithms sounds problematic, since we'd need to migrate the old tables to the new invariants/algorithm at some point to ever drop the old algorithm.

mheinzel: > Something I realised is that I prefer an explicit format because it might prevent getting bad tables from opening snapshots (i.e., tables with inconsistent state). However, maybe getting bad tables from opening old snapshots is something we should solve separately from the snapshot format compatibility question.

I agree with that, and the snapshot's version number can help us navigate this.

src/Database/LSMTree/Internal/MergingTree.hs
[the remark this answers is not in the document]
mheinzel: I updated the comment here and in the prototype. In the prototype, I also check that the claim about "no merges in level 1"  holds.

src/Database/LSMTree/Internal/Snapshot.hs
mheinzel: I'll merge this PR as it is now and open a new one for propagating the type parameter further inwards. Then we can discuss that separately.

PR #578, opened 2025-02-18
QLS: give all actions errors

jorisdral:
Instead of adding `Maybe Errors` to each constructor, which would probably lead to quite a bit of duplicated code, I opted to add a single top-level constructor that contains a `Maybe Errors` and an `Action'`. This `Action'` is a new data type that contains the interesting constructors, like `New` and `Updates`.

As a result of this change, there quite a few mechanical changes, but not much of it is interesting. However, having made this change, we are now ready to enable fault injection for all the actions whenever we want to.


wenkokke (approved):
Oof, the dynamic dict is rough.
Otherwise, LGTM.

PR #579, opened 2025-02-18
Remove bias from generation of `RawBytes` values

jeltsch:
Previously, the default generator for `RawBytes` had a tendency to generate small blocks of bytes. The reason was that the length of an underlying buffer could be at most the size parameter but the generated block of bytes constituted only a slice of the buffer. This pull request resolves this issue by allowing the length of the final slice to range up to the size parameter.

Furthermore, this pull request fixes an issue in the tests of transitivity of serialized-key equality and antisymmetry of serialized-key order. The previous implementations of these tests were supposed to work with a fixed, small size parameter, but they invoked the `mapSize` function for each particular test case and thus with keys already chosen instead of invoking it for the entire properties. This issue probably didn’t show up because of the bias towards small keys, likely resulting in most of the test cases using the empty key.

### DIFF
--- a/src-extras/Database/LSMTree/Extras/Generators.hs
+++ b/src-extras/Database/LSMTree/Extras/Generators.hs
@@ -42,6 +42,7 @@ import           Control.DeepSeq (NFData)
 import           Control.Exception (assert)
 import           Data.Coerce (coerce)
 import           Data.Containers.ListUtils (nubOrd)
+import           Data.Function ((&))
 import           Data.List (sort)
 import qualified Data.Primitive.ByteArray as BA
 import qualified Data.Vector.Primitive as VP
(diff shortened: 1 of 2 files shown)


src-extras/Database/LSMTree/Extras/Generators.hs
```diff
@@ -423,7 +424,12 @@ chunkSizeInvariant (ChunkSize csize) = chunkSizeLB <= csize && csize <= chunkSiz
 -------------------------------------------------------------------------------}
 
 instance Arbitrary RawBytes where
-  arbitrary = genRawBytes >>= genSlice
+  arbitrary = do
+    QC.NonNegative (QC.Small prefixLength)  <- arbitrary
+    QC.NonNegative (QC.Small payloadLength) <- arbitrary
+    QC.NonNegative (QC.Small suffixLength)  <- arbitrary
```
jorisdral, line 430: `Arbitrary Int` and `Arbitrary (Small Int)` are the same

jorisdral (approved):

jorisdral:
BTW, I added 8c829aae5e541ee615af91b9efa6188063949503. What do you think? Feel free to do with the commit what you want

jeltsch:
> BTW, I added [8c829aa](https://github.com/IntersectMBO/lsm-tree/commit/8c829aae5e541ee615af91b9efa6188063949503). What do you think? Feel free to do with the commit what you want.

I don’t think that using `quickcheck-classes` here is a good idea, because of what the documentation of `eqLaws` says:

> Some of these properties involve implication. In the case that the left hand side of the implication arrow does not hold, we do not retry. Consequently, these properties only end up being useful when the data type has a small number of inhabitants.

In our setting, we would get lots of test cases where the condition in the transitivity property is not fulfilled and only very few test cases where we actually test transitivity.

Interestingly, the implementation of transitivity of `(==)` in `quickcheck-classes` is not correct: it implements a property that’s stronger than transitivity, although this is harmless, because `(==)` must have that stronger property as well.

Issue #580, opened 2025-02-19
[BUG] Compact index tests no longer generate clashing keys


jorisdral:
The pages that generated for compact index tests no longer include pages that have clashing keys. It is important to restore these clashes: without clashes, we are not testing clash recovery logic that is baked into index construction and index search. To restore clashes, the generators have to be updated to generate clashing keys with a higher probability or to fix whatever is preventing those clashing keys from being generated. Moreover, we should guard against any similar future regressions by including a property like `coverTable` to ensure that clashes occur *at least* some percenage of the time.

On `main`, commit 02e3bf602e676bfdc9eac9c381e6de7bffc098d6:
```
❯ cabal run lsm-tree-test -- -p '$NF=="prop_distribution @KeyForIndexCompact"'
lsm-tree
  Test.Database.LSMTree.Internal.Index.Compact
    prop_distribution @KeyForIndexCompact: OK (0.02s)
      +++ OK, passed 100 tests (0% Has clashing multi-page values).

      # Clashes (100 in total):
      100% n == 0

      # Contiguous clash runs (100 in total):
      100% n == 0

      # Logical pages (100 in total):
      51% 10 <= n < 100
      41% 1 <= n < 10
       8% n == 0

      # MultiPageOneKey logical pages (100 in total):
      68% n == 0
      32% 1 <= n < 10

      # OnePageManyKeys logical pages (100 in total):
      48% 10 <= n < 100
      38% 1 <= n < 10
      14% n == 0

      # OnePageOneKey logical pages (100 in total):
      56% 1 <= n < 10
      44% n == 0

      # True pages (100 in total):
      56% 10 <= n < 100
      36% 1 <= n < 10
       8% n == 0

All 1 tests passed (0.02s)
```

PR #582, opened 2025-02-19
Make generator of serialized keys respect size constraints

jeltsch:
This pull request changes the generator of `SerialisedKey` values such that it only produces keys that respect the constraints on key sizes. It also changes the tests such that they use `SerialisedKey` instead of `KeyForIndexCompact` where the high probability of key conflicts provided by `KeyForIndexCompact` is not needed and `KeyForIndexCompact` was only used to enforce the above-mentioned size constraints.


test/Test/Database/LSMTree/Internal/Index/Compact.hs
```diff
@@ -49,14 +49,14 @@ import           Text.Printf (printf)
...
+      prop_distribution @BiasedKeyForIndexCompact
   , testProperty "prop_searchMinMaxKeysAfterConstruction" $
-      prop_searchMinMaxKeysAfterConstruction @KeyForIndexCompact 100
+      prop_searchMinMaxKeysAfterConstruction @BiasedKeyForIndexCompact 100
   , testProperty "prop_differentChunkSizesSameResults" $
-      prop_differentChunkSizesSameResults @KeyForIndexCompact
+      prop_differentChunkSizesSameResults @BiasedKeyForIndexCompact
   , testProperty "prop_singlesEquivMulti" $
-      prop_singlesEquivMulti @KeyForIndexCompact
+      prop_singlesEquivMulti @BiasedKeyForIndexCompact
```
mheinzel, line 59: Are there any other tests using `KeyForCompactIndex` that might benefit from handling collisions? Like `RunReader` for example. Probably the best default would be to keep everything at `BiasedKeyForIndexCompact` unless it is specifically moved to the unbiased version because we are sure it doesn't matter.

mheinzel (commented):
I have one important question, but apart from that it looks good to me.

test/Test/Database/LSMTree/Internal/Index/Compact.hs
jeltsch: > Are there any other tests using `KeyForCompactIndex` that might benefit from handling collisions? Like `RunReader` for example.

I fear that this is a question that I cannot answer, because I’m not familiar enough with the codebase.

> Probably the best default would be to keep everything at `BiasedKeyForIndexCompact` unless it is specifically moved to the unbiased version because we are sure it doesn't matter.

I’m not sure that this is a good idea, because it could be that having too much bias also causes problems. Best to decide on a case-by-case basis, not necessarily test-by-test but rather module-by-module. Maybe @jorisdral has an opinion about this. I’d also like to hear your suggestions, if possible.

jorisdral: > > Are there any other tests using `KeyForCompactIndex` that might benefit from handling collisions? Like `RunReader` for example.
> 
> I fear that this is a question that I cannot answer, because I’m not familiar enough with the codebase.

`RunReader` sounds like a good example where collisions would be useful. But note that the collisions that are interesting for the compact index are *clashses* in the 64 bits, not whole keys matching. So the preconditions on keys are slightly different

> > Probably the best default would be to keep everything at `BiasedKeyForIndexCompact` unless it is specifically moved to the unbiased version because we are sure it doesn't matter.
> 
> I’m not sure that this is a good idea, because it could be that having too much bias also causes problems. 

Is this hypothetical, or do you have an example?

> Best to decide on a case-by-case basis, not necessarily test-by-test but rather module-by-module. Maybe @jorisdral has an opinion about this. I’d also like to hear your suggestions, if possible.

It could be an option to use the opposite approach of the current PR. In the current PR, we make unbiased the default and use biased in a few places. It might be more useful to keep biased as the default, but switch to unbiased on a case-by-case basis.

test/Test/Database/LSMTree/Internal/RawBytes.hs
```diff
@@ -0,0 +1,86 @@
...
+
+-- * Utilities
+
+withFirstBlockSizeInfo :: RawBytes -> Property -> Property
+withFirstBlockSizeInfo firstBlock
+    = collect ("Size of first block is " ++ show (RawBytes.size firstBlock))
+
+-- * Properties to test
+
+-- ** 'Eq' laws
```
jorisdral, line 43: With the current (and previous) method of generating `RawBytes`, it's unlikely to generate two `RawBytes` that are equal. It means we mostly test `block1 /= block2` cases. If we want, we could try to change that

test/Test/Database/LSMTree/Internal/Index/Compact.hs
jeltsch: > `RunReader` sounds like a good example where collisions would be useful. But note that the collisions that are interesting for the compact index are _clashses_ in the 64 bits, not whole keys matching. So the preconditions on keys are slightly different

Those clashes of 64-bit prefixes is what the generator of `BiasedKeyForIndexCompact` mainly provokes. It also makes clashes of whole keys much more likely than the generator of `KeyForIndexCompact` but maybe not enough so to have them reliably occur to a significant extend. Are the clashes we would seek for `RunReader` tests clashes of whole keys?

jeltsch: > > > Probably the best default would be to keep everything at `BiasedKeyForIndexCompact` unless it is specifically moved to the unbiased version because we are sure it doesn't matter.
> > 
> > 
> > I’m not sure that this is a good idea, because it could be that having too much bias also causes problems.
> 
> Is this hypothetical, or do you have an example?

As of now, it’s just hypothetical. 🙂

jeltsch: > For the compact index tests, it would be useful to generate keys with clashes. I also already created #580 because we don't seem to be generating clashing keys anyway

Well, the generator used in the example run of #580 is the one that does provoke clashes, from what I can see: it should be the one associated with `KeyForIndexCompact`, which is the type that this pull requests renames into `BiasedKeyForIndexCompact`. Are you talking about whole-key clashes here? Maybe the generator’s tendency to induce those is not high enough, but it certainly should be there. Was there a higher clash probability in the past?

mheinzel: > Are the clashes we would seek for `RunReader` tests clashes of whole keys?

Yes, but hopefully the same generator would also make them sufficiently likely. We should label key collisions to check that (if we don't do that already).

jeltsch: Based on what the two of you, @jorisdral and @mheinzel, wrote above, I’ll switch from `KeyForCompactIndex` to `BiasedKeyForCompactIndex` in the tests for the following types:

* `RunReader`
* `RunReaders`
* `MergingRun`
* `Cursor`
* `Lookup`

jeltsch: > > Are the clashes we would seek for `RunReader` tests clashes of whole keys?
> 
> Yes, but hopefully the same generator would also make them sufficiently likely.

Hmm, #580 suggests that this generator does not make clashes of whole keys sufficiently likely. Note that #580 apparently uses said generator, the one that introduces bias.

jorisdral: ^ It's a regression

jeltsch: > Based on what the two of you, @jorisdral and @mheinzel, wrote above, I’ll switch from `KeyForCompactIndex` to `BiasedKeyForCompactIndex` in the tests for the following types:
> 
> * `RunReader`
> * `RunReaders`
> * `MergingRun`
> * `Cursor`
> * `Lookup`

I’ve made this change for `RunReader` and `RunReaders` in 7e3cecd875e2809f4fecdff3657e92ec6328cca3. Regarding the other ones, I hit the following issues:

* The test module for `MergingRun` doesn’t seem to refer to serialized keys at all.
* I couldn’t find a test module for `Cursor`.
* The test module for `Lookup` doesn’t specifically use serialized keys for compact indexes but serialized keys without constraints, for which we currently don’t have a biased generator, and I’m not sure which of the following ways to deal with this would be best:
    - Leave the `Lookup` test module alone.
    - Use the biased generator for compact-index keys in the `Lookup` test module.
    - Introduce a biased generator for keys without constraints and use it in the `Lookup` test module.

Thoughts?

jeltsch: Okay, at least @jorisdral thinks that those other modules can be left alone, as mentioned outside of GitHub.

test/Test/Database/LSMTree/Internal/RawBytes.hs
jeltsch: I’ve added output of statistics about block equality. It turns out that typically about 1.5 % of any two generated blocks are equal. Since this is quite a small share, I’ve increased the number of test cases for the above-mentioned properties to 10,000. All these changes are in be22c3f202fe0bca61ea94dd0d6949fb99ced9aa.

mheinzel (approved):
Looks good to me!

PR #585, opened 2025-02-20
prototype: carefully bound size of arbitrary trees

### LINKED ISSUE #570: [BUG] Sporadic timeouts for `prop_MergingTree`
**Describe the bug**
`prop_MergingTree` times out sporadically on GHA CI.

**To Reproduce**
`cabal run lsm-prototypes-tests -- -p 'prop_MergingTree'`. It's only sporadically timing out though, so probably one would have to try a few times to see it happen.

**Expected behavior**
Test should not time out.

**Output**

https://github.com/IntersectMBO/lsm-tree/actions/runs/13286590286/job/37096606225

I've seen the timeout happen a few times, and I think it's not linked to a particular OS or GHC version.

**Additional context**
A likely culprit for tests timing out is a loopy shrinker. There are a few to verify this, for example
* Run the test with `--quickcheck-shrinks=0` to prevent shrinking, and see if the test fails without timing out.
* Write tests for generators/shrinkers

If it's not the shrinker, then I'm not sure what the cause could be, since I don't know the tests very well (@mheinzel and @dcoutts probably do)

mheinzel:
# Description

Fixes #570 

Previously, occasionally a tree would get very large, leading to test timeouts. Now, the number of runs in a tree (as well as the number of tree constructors!) is bounded by the size parameter.


prototypes/ScheduledMergesTest.hs
```diff
@@ -294,20 +307,50 @@ completeM (MCompleted _ _ r)   = r
...
+          (first :) <$> arbitraryPartition (n - first)
+
+      -- Split into smaller positive numbers.
+      arbitraryPartition :: Int -> QC.Gen [Int]
+      arbitraryPartition n
+            | n <  1 = return []
+            | n == 1 = return [1]
+            | otherwise = do
+              first <- QC.chooseInt (1, n)
+              (first :) <$> arbitraryPartition (n - first)
```
jorisdral, line 387: These are not very fair partitions. Is that bad?

prototypes/ScheduledMergesTest.hs
```diff
@@ -23,6 +23,13 @@ tests = testGroup "Unit and property tests"
     [ testCase "test_regression_empty_run" test_regression_empty_run
     , testCase "test_merge_again_with_incoming" test_merge_again_with_incoming
     , testProperty "prop_union" prop_union
+    , testGroup "T"
+        [ testProperty "Arbitrary satisfies invariant" invariantT
+        , testProperty "Shrinking satisfies invariant" $ \t ->
+            case shrink t of
+              [] -> QC.label "no shrinks" $ QC.property True
+              ts -> QC.forAll (QC.growingElements ts) invariantT
```
jorisdral, line 33: I added 2778bff149ced163cb3912e20d975d568ef08718 to get a sense of what the generated trees look like. What surprised me is that the shrink test only tests *one* shrink value per property run. Is that what we intended when we started using `QC.growingElements`? In my mind we intended to test a subset of the shrinking values, but that's not the case

```
❯ cabal run lsm-prototypes-tests -- -p "T"
prototype
  ScheduledMerges
    Unit and property tests
      T
        Arbitrary satisfies invariant: OK (0.02s)
          +++ OK, passed 100 tests.

          Tree depth (100 in total):
          39% 4 <= n < 8
          28% 1 <= n < 2
          24% 2 <= n < 4
           9% n == 0

          Tree size (100 in total):
          25% 16 <= n < 32
          18% 32 <= n < 64
          18% 8 <= n < 16
          15% 4 <= n < 8
           9% 1 <= n < 2
           8% 64 <= n < 128
           7% 2 <= n < 4
        Shrinking satisfies invariant: OK (1.38s)
          +++ OK, passed 100 tests.

          Tree depth (100 in total):
          40% 2 <= n < 4
          34% 4 <= n < 8
          20% 1 <= n < 2
           5% n == 0
           1% 8 <= n < 16

          Tree size (100 in total):
          23% 32 <= n < 64
          20% 16 <= n < 32
          18% 8 <= n < 16
          14% 2 <= n < 4
          12% 4 <= n < 8
           7% 64 <= n < 128
           6% 1 <= n < 2
      prop_MergingTree:                OK (0.11s)
        +++ OK, passed 100 tests; 15 discarded.

All 3 tests passed (1.51s)
```

jorisdral (approved):
LGTM! 

BTW, I added 2778bff for tabulating some statistics about the trees, but feel free to do with the commit what you want. It seemed easier to push the commit here than share the code through other channels.

prototypes/ScheduledMergesTest.hs
mheinzel: I think it's generally good to have varying sizes of sub-trees and these get shuffled afterwards, so there is no correlation between size and order of the children. It also has the interesting effect that for overall size `n` we get `log n` children on average, each of which has average size `n / log n`. So increasing the size both leads to larger branching factor as well as deeper trees.

prototypes/ScheduledMergesTest.hs
mheinzel: > the shrink test only tests _one_ shrink value per property run. Is that what we intended when we started using `QC.growingElements`?

Not sure, but maybe the argument was that there is a larger chance to find something interesting when generating 100 inputs and checking one shrink of each, than generating one input and checking 100 of its shrinks. However,
1. that's a bit wasteful, so probably the best ratio of interesting exploration to runtime required is a bit higher than testing 1 shrink per input
2. we should increase the number of test cases so that we check much more than 100 shrinks in total, no matter how they are distributed.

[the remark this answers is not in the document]
mheinzel: I think that would be problematic. If we find out that the shrinker is broken, we should probably not rely on it to find a minimal example for us.

mheinzel: Looking at the [source code for `growingElements`](https://hackage.haskell.org/package/QuickCheck-2.15.0.1/docs/src/Test.QuickCheck.Gen.html#growingElements), I'm not convinced it offers us benefits over plain `elements`. It just skews towards the front, which makes it unlikely to find bugs towards the end of the possible shrinks.

I think originally the idea was to make it faster for very long lists of shrinks (or even make it terminate for infinite ones?), but using `length xs` requires to traverse the whole list either way.

So should I switch it (and `prop_forAllArbitraryAndShrinkPreserveInvariant`) to something like `vectorOf 10 (elements ts)`, or maybe something equivalent that doesn't calculate the length ten times?

jorisdral: Something like `listOf (elements _)`/`vectorOf _` sounds good to me.

mheinzel: @jorisdral I had another related idea: Instead of just picking a few possible shrinks of the original value, we could iteratively shrink, at least a few times. This can also find cases where only shrinking multiple times breaks an invariant. The first shrink could result in a state that does not violate the invariant, but is not covered by the generator and triggers problematic behaviour on the next shrink.

```haskell
prop_shrinkSatisfiesInvariant :: T -> Property
prop_shrinkSatisfiesInvariant t =
    QC.forAll (genShrinkTrace 4 t) $ \trace ->
      QC.tabulate "Trace length" [showPowersOf 2 $ length trace] $
      QC.conjoin $ flip map trace $ \(numAlternatives, t') ->
        QC.tabulate "Shrink alternatives" [showPowersOf 2 numAlternatives] $
          Right () === invariantT t'

-- | Iterative shrinks, and how many alternatives were possible at each point.
genShrinkTrace :: Arbitrary a => Int -> a -> QC.Gen [(Int, a)]
genShrinkTrace !n x
  | n <= 0 = return []
  | otherwise =
    case shrink x of
      [] -> return []
      xs -> do
        -- like QC.elements, but we want access to the length
        let len = length xs
        x' <- (xs !!) <$> QC.chooseInt (0, len - 1)
        ((len, x') :) <$> genShrinkTrace (n - 1) x'
```

Seems to work well and the runtime might still be acceptable?

```
        Shrinking satisfies invariant: OK (6.82s)
          +++ OK, passed 100 tests.
          
          Shrink alternatives (387 in total):
          18.3% 8192 <= n < 16384
          10.6% 512 <= n < 1024
          10.3% 1024 <= n < 2048
          10.3% 16384 <= n < 32768
           9.8% 4096 <= n < 8192
           7.0% 32768 <= n < 65536
           6.7% 2048 <= n < 4096
           5.4% 256 <= n < 512
           5.2% 128 <= n < 256
           5.2% 32 <= n < 64
           4.1% 64 <= n < 128
           2.3% 2 <= n < 4
           1.6% 16 <= n < 32
           1.3% 4 <= n < 8
           1.0% 1 <= n < 2
           0.8% 8 <= n < 16
          
          Trace length (100 in total):
          94% 4 <= n < 8
           5% 2 <= n < 4
           1% n == 0
```

Lots of possible shrinks, mainly due to having a lot of `Data.Map`s that can be shrunk. If I reduce QuickCheck's maximum size to 60, the tests only need about 2 seconds. Seems like a good tradeoff?

mheinzel: I think I'll go ahead with that version here (check arbitrary shrinking trace of depth 4, max size 60), get it merged, and then later we could do something similar in `prop_forAllArbitraryAndShrinkPreserveInvariant`.

PR #586, opened 2025-02-20
Loosen the size constraints on serialized keys

jeltsch:
Currently, the documentation of the general index interface requires serialized keys to satisfy the key size constraints of compact and ordinary indexes. This pull request changes the documentation such that it requires only the key size constraint of the actually used index type to be satisfied.


mheinzel (approved):
:+1:

PR #592, opened 2025-02-25
Introduce MergingRunData and MergingTreeData

mheinzel:
# Description

Similarly to the existing `RunData`, these are simplified versions of `MergingRun` and `MergingTree`, which can be generated, shrunk and turned into their full on-disk versions.

I also added labelling and tests for the generators and shrinkers, as well as the code creating the `MergingTree` etc. from the data.

The tests are quite slow even after halving the maximum size. The main issue is that `RunData` has a lot of possible shrinks (sometimes over 100k), which gets even worse when there are dozens of them in the tree.

<details>
<summary>test output</summary>

```
  Test.Database.LSMTree.Generators
    RunData
      Arbitrary satisfies invariant:          OK (0.60s)
        +++ OK, passed 100 tests.
        
        run has large k/ops (100 in total):
        52% has large k/op
        48% no large k/op
        
        run length (100 in total):
        68% 10 <= n < 100
        26% 1 <= n < 10
         6% n == 0
        
        value size (1974 in total):
        53.80% 10 <= n < 100
        17.88% 100 <= n < 1000
        15.75% 1000 <= n < 10000
        11.65% 1 <= n < 10
         0.91% n == 0
      Shrinking satisfies invariant:          OK (6.67s)
        +++ OK, passed 100 tests (3% no shrinks).
        
        number of shrinks (97 in total):
        45% 10000 <= n < 100000
        32% 1000 <= n < 10000
        20% 100 <= n < 1000
         3% 10 <= n < 100
      withRun doesn't leak resources:         OK (0.65s)
        +++ OK, passed 100 tests.
    NonEmptyRunData
      Arbitrary satisfies invariant:          OK (0.69s)
        +++ OK, passed 100 tests.
        
        run has large k/ops (100 in total):
        57% has large k/op
        43% no large k/op
        
        run length (100 in total):
        67% 10 <= n < 100
        33% 1 <= n < 10
        
        value size (2475 in total):
        54.71% 10 <= n < 100
        18.83% 100 <= n < 1000
        15.27% 1000 <= n < 10000
        10.79% 1 <= n < 10
         0.40% n == 0
      Shrinking satisfies invariant:          OK (14.08s)
        +++ OK, passed 100 tests (1% no shrinks).
        
        number of shrinks (99 in total):
        45% 10000 <= n < 100000
        38% 1000 <= n < 10000
        12% 100 <= n < 1000
         2% 1 <= n < 10
         2% 10 <= n < 100
    MergingRunData
      Arbitrary satisfies invariant:          OK (0.56s)
        +++ OK, passed 100 tests.
        
        merge type (100 in total):
        50% MergeLastLevel
        50% MergeMidLevel
        
        merging run inputs (53 in total):
        55% 4 <= n < 8
        36% 2 <= n < 4
         9% 8 <= n < 16
        
        merging run state (100 in total):
        53% OngoingMerge
        47% CompletedMerge
        
        run has large k/ops (295 in total):
        75.3% no large k/op
        24.7% has large k/op
        
        run length (295 in total):
        74.6% 1 <= n < 10
        23.7% 10 <= n < 100
         1.7% n == 0
        
        value size (2138 in total):
        33.77% 10 <= n < 100
        30.17% 1 <= n < 10
        17.87% 100 <= n < 1000
        16.09% 1000 <= n < 10000
         2.10% n == 0
      Shrinking satisfies invariant:          OK (4.02s)
        +++ OK, passed 100 tests.
        
        number of shrinks (100 in total):
        47% 1000 <= n < 10000
        24% 100 <= n < 1000
        24% 10000 <= n < 100000
         3% 1 <= n < 10
         2% 10 <= n < 100
      withMergingRun doesn't leak resources:  OK (0.80s)
        +++ OK, passed 100 tests.
    MergingTreeData
      Arbitrary satisfies invariant:          OK (4.25s)
        +++ OK, passed 100 tests.
        
        merge type (593 in total):
        40.6% MergeMidLevel
        37.1% MergeLastLevel
        12.5% MergeLevel
         9.8% MergeUnion
        
        merging run inputs (278 in total):
        52.5% 4 <= n < 8
        33.1% 2 <= n < 4
        14.4% 8 <= n < 16
        
        merging run state (593 in total):
        53.1% CompletedMerge
        46.9% OngoingMerge
        
        merging tree state (604 in total):
        29.8% PendingLevelMerge
        25.0% PendingUnionMerge
        23.3% CompletedTreeMerge
        21.9% OngoingTreeMerge
        
        run has large k/ops (2214 in total):
        72.36% no large k/op
        27.64% has large k/op
        
        run length (2214 in total):
        70.28% 1 <= n < 10
        28.27% 10 <= n < 100
         1.45% n == 0
        
        tree depth (100 in total):
        51% 2 <= n < 4
        38% 4 <= n < 8
        11% 1 <= n < 2
        
        value size (16784 in total):
        37.786% 10 <= n < 100
        27.359% 1 <= n < 10
        17.165% 100 <= n < 1000
        15.473% 1000 <= n < 10000
         2.216% n == 0
      Shrinking satisfies invariant:          OK (22.70s)
        +++ OK, passed 100 tests (2% no shrinks).
        
        number of shrinks (98 in total):
        39% 10000 <= n < 100000
        24% 1000 <= n < 10000
        21% 100000 <= n < 1000000
        13% 100 <= n < 1000
         1% 1 <= n < 10
         1% 10 <= n < 100
      withMergingTree doesn't leak resources: OK (6.39s)
        +++ OK, passed 100 tests.
```

</details>

### DIFF
--- a/src-extras/Database/LSMTree/Extras/MergingRunData.hs
+++ b/src-extras/Database/LSMTree/Extras/MergingRunData.hs
@@ -1,0 +1,222 @@
+-- | Utilities for generating 'MergingRun's. Tests and benchmarks should
+-- preferably use these utilities instead of (re-)defining their own.
+module Database.LSMTree.Extras.MergingRunData (
+    -- * Create merging runs
+    withMergingRun
+  , unsafeCreateMergingRun
+    -- * MergingRunData
+  , MergingRunData (..)
+  , mergingRunDataMergeType
+  , mergingRunDataInvariant
+  , mapMergingRunData
+  , SerialisedMergingRunData
+  , serialiseMergingRunData
+    -- * QuickCheck
+  , labelMergingRunData
+  , genMergingRunData
+  , shrinkMergingRunData
+  ) where
+
+import           Control.Exception (bracket)
+import           Control.RefCount
+import qualified Data.Vector as V
+import           Database.LSMTree.Extras (showPowersOf)
+import           Database.LSMTree.Extras.Generators ()
+import           Database.LSMTree.Extras.RunData
+import           Database.LSMTree.Internal.Index (IndexType)
+import           Database.LSMTree.Internal.Lookup (ResolveSerialisedValue)
+import           Database.LSMTree.Internal.MergingRun (MergingRun)
+import qualified Database.LSMTree.Internal.MergingRun as MR
+import           Database.LSMTree.Internal.Paths
+import           Database.LSMTree.Internal.Run (RunDataCaching (..))
+import qualified Database.LSMTree.Internal.Run as Run
+import           Database.LSMTree.Internal.RunAcc (RunBloomFilterAlloc (..))
+import           Database.LSMTree.Internal.RunNumber
+import           Database.LSMTree.Internal.Serialise
+import           Database.LSMTree.Internal.UniqCounter
+import qualified System.FS.API as FS
+import           System.FS.API (HasFS)
+import           System.FS.BlockIO.API (HasBlockIO)
+import           Test.QuickCheck as QC
+
+{-------------------------------------------------------------------------------
+  Create merging runs
+-------------------------------------------------------------------------------}
+
+-- | Create a temporary 'MergingRun' using 'unsafeCreateMergingRun'.
+withMergingRun ::
+     MR.IsMergeType t
+  => HasFS IO h
+  -> HasBlockIO IO h
+  -> ResolveSerialisedValue
+  -> IndexType
+  -> FS.FsPath
+  -> UniqCounter IO
+  -> SerialisedMergingRunData t
+  -> (Ref (MergingRun t IO h) -> IO a)
+  -> IO a
+withMergingRun hfs hbio resolve indexType path counter mrd = do
+    bracket
+      (unsafeCreateMergingRun hfs hbio resolve indexType path counter mrd)
+      releaseRef
+
+-- | Flush serialised merging run data to disk.
+--
+-- This might leak resources if not run with asynchronous exceptions masked.
+-- Consider using 'withMergingRun' instead.
+--
+-- Use of this function should be paired with a 'releaseRef'.
+unsafeCreateMergingRun ::
+     MR.IsMergeType t
+  => HasFS IO h
+  -> HasBlockIO IO h
+  -> ResolveSerialisedValue
+  -> IndexType
+  -> FS.FsPath
+  -> UniqCounter IO
+  -> SerialisedMergingRunData t
+  -> IO (Ref (MergingRun t IO h))
+unsafeCreateMergingRun hfs hbio resolve indexType path counter = \case
+    CompletedMergeData _ numRuns rd -> do
+      withRun hfs hbio indexType path counter rd $ \run -> do
+        -- slightly hacky, generally it's larger
+        let totalDebt = MR.numEntriesToMergeDebt (Run.size run)
+        MR.newCompleted numRuns totalDebt run
+
+    OngoingMergeData mergeType rds -> do
+      withRuns hfs hbio indexType path counter (toRunData <$> rds)
+        $ \runs -> do
+          n <- incrUniqCounter counter
+          let fsPaths = RunFsPaths path (RunNumber (uniqueToInt n))
+          MR.new hfs hbio resolve CacheRunData (RunAllocFixed 10) indexType
+            mergeType fsPaths (V.fromList runs)
+
+{-------------------------------------------------------------------------------
+  MergingRunData
+-------------------------------------------------------------------------------}
+
+-- | A data structure suitable for creating arbitrary 'MergingRun's.
+--
+-- Note: 'b ~ Void' should rule out blobs.
+--
+-- Currently, ongoing merges are always \"fresh\", i.e. there is no merge work
+-- already performed.
+--
+-- TODO: Generate merge credits and supply them in 'unsafeCreateMergingRun',
+-- similarly to how @ScheduledMergesTest@ does it.
+data MergingRunData t k v b =
+    CompletedMergeData t MR.NumRuns (RunData k v b)
+  | OngoingMergeData t [NonEmptyRunData k v b]  -- ^ at least 2 inputs
+  deriving stock (Show, Eq)
+
+mergingRunDataMergeType :: MergingRunData t k v b -> t
+mergingRunDataMergeType = \case
+    CompletedMergeData mt _ _ -> mt
+    OngoingMergeData mt _     -> mt
+
+-- | See @mergeInvariant@ in the prototype.
+mergingRunDataInvariant :: MergingRunData t k v b -> Either String ()
+mergingRunDataInvariant = \case
+    CompletedMergeData _ (MR.NumRuns n) _ ->
+      assertI "completed merges are non-trivial (at least two inputs)" $
+        n >= 2
+    OngoingMergeData _ rds -> do
+      assertI "ongoing merges are non-trivial (at least two inputs)" $
+        length rds >= 2
+  where
+    assertI msg False = Left msg
+    assertI _   True  = Right ()
+
+mapMergingRunData ::
+     Ord k'
+  => (k -> k') -> (v -> v') -> (b -> b')
+  -> MergingRunData t k v b -> MergingRunData t k' v' b'
+mapMergingRunData f g h = \case
+    CompletedMergeData t n r ->
+      CompletedMergeData t n $ mapRunData f g h r
+    OngoingMergeData t rs ->
+      OngoingMergeData t $ map (mapNonEmptyRunData f g h) rs
+
+type SerialisedMergingRunData t =
+    MergingRunData t SerialisedKey SerialisedValue SerialisedBlob
+
+serialiseMergingRunData ::
+     (SerialiseKey k, SerialiseValue v, SerialiseValue b)
+  => MergingRunData t k v b -> SerialisedMergingRunData t
+serialiseMergingRunData =
+    mapMergingRunData serialiseKey serialiseValue serialiseBlob
+
+{-------------------------------------------------------------------------------
+  QuickCheck
+-------------------------------------------------------------------------------}
+
+labelMergingRunData ::
+     Show t => SerialisedMergingRunData t -> Property -> Property
+labelMergingRunData (CompletedMergeData mt _ rd) =
+      tabulate "merging run state" ["CompletedMerge"]
+    . tabulate "merge type" [show mt]
+    . labelRunData rd
+labelMergingRunData (OngoingMergeData mt rds) =
+      tabulate "merging run state" ["OngoingMerge"]
+    . tabulate "merge type" [show mt]
+    . tabulate "merging run inputs" [showPowersOf 2 (length rds)]
+    . foldr ((.) . labelNonEmptyRunData) id rds
+
+instance ( Arbitrary t, Ord k, Arbitrary k, Arbitrary v, Arbitrary b
+         ) => Arbitrary (MergingRunData t k v b) where
+  arbitrary = genMergingRunData arbitrary arbitrary arbitrary arbitrary
+  shrink = shrinkMergingRunData shrink shrink shrink
+
+genMergingRunData ::
+     Ord k
+  => Gen t
+  -> Gen k
+  -> Gen v
+  -> Gen b
+  -> Gen (MergingRunData t k v b)
+genMergingRunData genMergeType genKey genVal genBlob =
+    QC.oneof
+      [ do
+          mt <- genMergeType
+          numRuns <- MR.NumRuns <$> QC.chooseInt (2, 8)
+          rd <- genRunData genKey genVal genBlob
+          pure (CompletedMergeData mt numRuns rd)
+      , do
+          s  <- QC.getSize
+          mt <- genMergeType
+          n  <- QC.chooseInt (2, max 2 (s * 8 `div` 100))  -- 2 to 8
+          rs <- QC.vectorOf n $
+            -- Scaled, so overall number of entries is similar to a completed
+            -- merge. However, the entries themselves should not be smaller.
+            QC.scale (`div` n) $
+              genNonEmptyRunData
+                (resize s genKey)
+                (resize s genVal)
+                (resize s genBlob)
+          pure (OngoingMergeData mt rs)
+      ]
+
+shrinkMergingRunData ::
+     Ord k
+  => (k -> [k])
+  -> (v -> [v])
+  -> (b -> [b])
+  -> MergingRunData t k v b
+  -> [MergingRunData t k v b]
+shrinkMergingRunData shrinkKey shrinkVal shrinkBlob = \case
+    CompletedMergeData mt numRuns rd ->
+      [ CompletedMergeData mt numRuns' rd'
+      | (numRuns', rd') <-
+          liftShrink2
+            (fmap MR.NumRuns . filter (>= 2) . shrink . MR.unNumRuns)
+            (shrinkRunData shrinkKey shrinkVal shrinkBlob)
+            (numRuns, rd)
+      ]
+    OngoingMergeData mt rds ->
+      [ OngoingMergeData mt rds'
+      | rds' <-
+          liftShrink
+            (shrinkNonEmptyRunData shrinkKey shrinkVal shrinkBlob)
+            rds
+      , length rds' >= 2
+      ]
--- a/src-extras/Database/LSMTree/Extras/MergingTreeData.hs
+++ b/src-extras/Database/LSMTree/Extras/MergingTreeData.hs
@@ -265,7 +265,8 @@ genMergingTreeData genKey genVal genBlob =
 --
 -- The size is measured by the number of MergingTreeData constructors.
 genMergingTreeDataOfSize ::
-     Ord k => Gen k -> Gen v -> Gen b -> Int -> Gen (MergingTreeData k v b)
+     forall k v b. Ord k
+  => Gen k -> Gen v -> Gen b -> Int -> Gen (MergingTreeData k v b)
 genMergingTreeDataOfSize genKey genVal genBlob = \n0 -> do
     tree <- genMergingTree n0
     assert (mergingTreeDataSize tree == n0) $
@@ -277,8 +278,8 @@ genMergingTreeDataOfSize genKey genVal genBlob = \n0 -> do
 
       | n == 1
       = QC.oneof
-          [ CompletedTreeMergeData <$> genRunData genKey genVal genBlob
-          , OngoingTreeMergeData <$> genMergingRunData arbitrary genKey genVal genBlob
+          [ CompletedTreeMergeData <$> genRun
+          , OngoingTreeMergeData <$> genMergingRun arbitrary
           , genPendingLevelMergeNoChild
           ]
 
@@ -290,12 +291,12 @@ genMergingTreeDataOfSize genKey genVal genBlob = \n0 -> do
 
     -- n == 1
     genPendingLevelMergeNoChild = do
-        numPreExisting <- chooseIntSkewed (0, 5)
+        numPreExisting <- chooseIntSkewed (0, 6)
         initPreExisting <- QC.vectorOf numPreExisting $
           -- these can't be last level. we generate the last input below.
-          genPreExistingRunData (pure MR.MergeMidLevel) genKey genVal genBlob
+          genPreExistingRun (pure MR.MergeMidLevel)
         -- there must be at least one (last) input to the pending merge.
-        lastPreExisting <- genPreExistingRunData arbitrary genKey genVal genBlob
+        lastPreExisting <- genPreExistingRun arbitrary
         let preExisting = initPreExisting ++ [lastPreExisting]
         return (PendingLevelMergeData preExisting Nothing)
 
@@ -304,7 +305,7 @@ genMergingTreeDataOfSize genKey genVal genBlob = \n0 -> do
         numPreExisting <- chooseIntSkewed (0, 6)
         preExisting <- QC.vectorOf numPreExisting $
           -- there can't be a last level merge, child is last
-          genPreExistingRunData (pure MR.MergeMidLevel) genKey genVal genBlob
+          genPreExistingRun (pure MR.MergeMidLevel)
         tree <- genMergingTree (n - 1)
         return (PendingLevelMergeData preExisting (Just tree))
 
@@ -312,6 +313,18 @@ genMergingTreeDataOfSize genKey genVal genBlob = \n0 -> do
     genPendingUnionMerge n = do
         ns <- QC.shuffle =<< arbitraryPartition2 (n - 1)
         PendingUnionMergeData <$> traverse genMergingTree ns
+
+    genRun                    = genScaled genRunData
+    genMergingRun genType     = genScaled (genMergingRunData genType)
+    genPreExistingRun genType = genScaled (genPreExistingRunData genType)
+
+    -- To avoid generating too large test cases, we reduce the number of
+    -- entries for each run. The size of the individual entries is unaffected.
+    genScaled :: forall r. (Gen k -> Gen v -> Gen b -> Gen r) -> Gen r
+    genScaled gen =
+        QC.sized $ \s ->
+          QC.scale (`div` 2) $
+            gen (QC.resize s genKey) (QC.resize s genVal) (QC.resize s genBlob)
 
     -- skewed towards smaller values
     chooseIntSkewed (lb, ub) = do
--- a/src-extras/Database/LSMTree/Extras/MergingTreeData.hs
+++ b/src-extras/Database/LSMTree/Extras/MergingTreeData.hs
@@ -133,29 +133,38 @@ data PreExistingRunData k v b =
   | PreExistingMergingRunData (MergingRunData MR.LevelMergeType k v b)
   deriving stock (Show, Eq)
 
+mergingTreeIsStructurallyEmpty :: MergingTreeData k v b -> Bool
+mergingTreeIsStructurallyEmpty = \case
+    CompletedTreeMergeData _   -> False  -- could be, but we match MT
+    OngoingTreeMergeData _     -> False
+    PendingLevelMergeData ps t -> null ps && null t
+    PendingUnionMergeData ts   -> null ts
+
 -- | See @treeInvariant@ in prototype.
 mergingTreeDataInvariant :: MergingTreeData k v b -> Either String ()
-mergingTreeDataInvariant = \case
-    CompletedTreeMergeData _rd ->
-      Right ()
-    OngoingTreeMergeData mr ->
-      mergingRunDataInvariant mr
-    PendingLevelMergeData prs t -> do
-      assertI "pending level merges have at least one input" $
-        length prs + length t > 0
-      for_ prs $ \case
-        PreExistingRunData        _r -> Right ()
-        PreExistingMergingRunData mr -> mergingRunDataInvariant mr
-      for_ (drop 1 (reverse prs)) $ \case
-        PreExistingRunData        _r -> Right ()
-        PreExistingMergingRunData mr ->
-          assertI "only the last pre-existing run can be a last level merge" $
-            mergingRunDataMergeType mr == MR.MergeMidLevel
-      for_ t mergingTreeDataInvariant
-    PendingUnionMergeData ts -> do
-      assertI "pending union merges are non-trivial (at least two inputs)" $
-        length ts >= 2
-      for_ ts mergingTreeDataInvariant
+mergingTreeDataInvariant mtd
+  | mergingTreeIsStructurallyEmpty mtd = Right ()
+  | otherwise = case mtd of
+      CompletedTreeMergeData _rd ->
+        Right ()
+      OngoingTreeMergeData mr ->
+        mergingRunDataInvariant mr
+      PendingLevelMergeData prs t -> do
+        assertI "pending level merges have at least one input" $
+          length prs + length t > 0
+        for_ prs $ \case
+          PreExistingRunData        _r -> Right ()
+          PreExistingMergingRunData mr -> mergingRunDataInvariant mr
+        for_ (drop 1 (reverse prs)) $ \case
+          PreExistingRunData        _r -> Right ()
+          PreExistingMergingRunData mr ->
+            assertI "only the last pre-existing run can be a last level merge" $
+              mergingRunDataMergeType mr == MR.MergeMidLevel
+        for_ t mergingTreeDataInvariant
+      PendingUnionMergeData ts -> do
+        assertI "pending union merges are non-trivial (at least two inputs)" $
+          length ts >= 2
+        for_ ts mergingTreeDataInvariant
   where
     assertI msg False = Left msg
     assertI _   True  = Right ()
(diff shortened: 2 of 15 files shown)


jorisdral:
> The tests are quite slow ...

"the tests" meaning the generator invariant tests, I take it?

src-extras/Database/LSMTree/Extras/MergingRunData.hs
```diff
@@ -0,0 +1,217 @@
...
+import qualified System.FS.API as FS
+import           System.FS.API (HasFS)
+import           System.FS.BlockIO.API (HasBlockIO)
+import           Test.QuickCheck as QC
+
+{-------------------------------------------------------------------------------
+  Flush runs
+-------------------------------------------------------------------------------}
+
+-- | Create a temporary 'Run' using 'unsafeFlushMergingRun'.
```
jorisdral, line 46: Outdated comment

src-extras/Database/LSMTree/Extras/MergingRunData.hs
```diff
@@ -0,0 +1,217 @@
...
+  -> UniqCounter IO
+  -> SerialisedMergingRunData t
+  -> (Ref (MergingRun t IO h) -> IO a)
+  -> IO a
+withMergingRun hfs hbio resolve indexType path counter mrd = do
+    bracket
+      (unsafeFlushMergingRun hfs hbio resolve indexType path counter mrd)
+      releaseRef
+
+-- | Flush serialised run data to disk as if it were a write buffer.
```
jorisdral, line 63: Outdated comment

src-extras/Database/LSMTree/Extras/MergingRunData.hs
```diff
@@ -0,0 +1,217 @@
...
+--
+-- This might leak resources if not run with asynchronous exceptions masked.
+-- Use helper functions like 'withMergingRun' or 'withMergingRuns' instead.
```
jorisdral, line 66: There is no `withMergingRuns`

src-extras/Database/LSMTree/Extras/MergingRunData.hs
```diff
@@ -0,0 +1,217 @@
...
+--
+-- Use of this function should be paired with a 'releaseRef'.
+unsafeFlushMergingRun ::
```
jorisdral, line 69: "Flush" in my mind means flushing of the write buffer to disk. Maybe we should just call it `unsafeAsMergingRun`?

src-extras/Database/LSMTree/Extras/MergingRunData.hs
```diff
@@ -0,0 +1,217 @@
...
+  -> UniqCounter IO
+  -> SerialisedMergingRunData t
+  -> IO (Ref (MergingRun t IO h))
+unsafeFlushMergingRun hfs hbio resolve indexType path counter = \case
+    CompletedMergeData _ numRuns rd -> do
+      run <- unsafeFlushRun hfs hbio indexType path counter rd
+      let totalDebt = Run.size run  -- slightly hacky, generally it's larger
+      mr <- MR.newCompleted numRuns totalDebt run
+      releaseRef run
+      return mr
```
jorisdral, line 85: Use a `withRun` equivalent, because this pairing of `unsafeFlushRun` and `releaseRef` is not exception safe

src-extras/Database/LSMTree/Extras/MergingRunData.hs
```diff
@@ -0,0 +1,217 @@
...
+
+    OngoingMergeData mergeType rds -> do
+      runs <- fmap V.fromList $ forM rds $ \rd -> do
+        unsafeFlushRun hfs hbio indexType path counter (unNonEmptyRunData rd)
+      n <- incrUniqCounter counter
+      let fsPaths = RunFsPaths path (RunNumber (uniqueToInt n))
+      mr <- MR.new hfs hbio resolve CacheRunData (RunAllocFixed 10) indexType
+              mergeType fsPaths runs
+      mapM_ releaseRef runs
+      return mr
```
jorisdral, line 95: Similarly, use a `withRuns` equivalent

src-extras/Database/LSMTree/Extras/MergingRunData.hs
```diff
@@ -0,0 +1,217 @@
...
+     Show t => SerialisedMergingRunData t -> Property -> Property
+labelMergingRunData (CompletedMergeData mt _ rd) =
+      tabulate "merging run state" ["CompletedMerge"]
+    . tabulate "merge type" [show mt]
+    . labelRunData rd
+labelMergingRunData (OngoingMergeData mt rds) =
+      tabulate "merging run state" ["OngoingMerge"]
+    . tabulate "merge type" [show mt]
+    . tabulate "merging run inputs" [showPowersOf 2 (length rds)]
+    . foldr ((.) . labelRunData . unNonEmptyRunData) id rds
```
jorisdral, line 164: Maybe tabulate "merging run inputs" for the first one as well?

src-extras/Database/LSMTree/Extras/MergingRunData.hs
```diff
@@ -0,0 +1,217 @@
...
+    CompletedMergeData mt numRuns rd ->
+      [ CompletedMergeData mt numRuns' rd'
+      | (numRuns', rd') <-
+          liftShrink2
+            (fmap MR.NumRuns . filter (> 1) . shrink . MR.unNumRuns)
+            (shrinkRunData shrinkKey shrinkVal shrinkBlob)
+            (numRuns, rd)
+      ]
+    OngoingMergeData mt rds ->
+      -- [ CompletedMergeData mt md (completeM m) ]
```
jorisdral, line 210: Remove

src-extras/Database/LSMTree/Extras/MergingRunData.hs
```diff
@@ -0,0 +1,217 @@
...
+genMergingRunData genKey genVal genBlob =
+    QC.oneof
+      [ do
+          mt <- arbitrary
+          numRuns <- MR.NumRuns <$> QC.chooseInt (2, 32)
+          rd <- genRunData genKey genVal genBlob
+          pure (CompletedMergeData mt numRuns rd)
+      , do
+          mt <- arbitrary
+          n  <- QC.chooseInt (2, 8)
```
jorisdral, line 186: Is it important for tests that there are more than 2 or 3? I feel like we test all the interesting cases with 2 or 3 input runs.

src-extras/Database/LSMTree/Extras/MergingRunData.hs
```diff
@@ -0,0 +1,217 @@
...
+          rs <- QC.vectorOf n $ QC.scale (`div` n) $
+            -- scaled, so overall size is similar to a single RunData
+            genNonEmptyRunData genKey genVal genBlob
```
jorisdral, line 189: We should check, but I think this not only scales the number of key/value pairs, but also the size of the keys/values themselves. Is that problematic?

src-extras/Database/LSMTree/Extras/MergingRunData.hs
```diff
@@ -0,0 +1,217 @@
...
+     forall t k v b. (Arbitrary t, Ord k)
+  => Gen k
+  -> Gen v
+  -> Gen b
+  -> Gen (MergingRunData t k v b)
+genMergingRunData genKey genVal genBlob =
+    QC.oneof
+      [ do
+          mt <- arbitrary
+          numRuns <- MR.NumRuns <$> QC.chooseInt (2, 32)
```
jorisdral, line 181: Similarly, would using `QC.chooseInt (2, 3)` be sufficient?

src-extras/Database/LSMTree/Extras/MergingTreeData.hs
```diff
@@ -0,0 +1,358 @@
...
+import qualified System.FS.API as FS
+import           System.FS.API (HasFS)
+import           System.FS.BlockIO.API (HasBlockIO)
+import           Test.QuickCheck as QC
+
+{-------------------------------------------------------------------------------
+  Flush runs
+-------------------------------------------------------------------------------}
+
+-- | Create a temporary 'Run' using 'unsafeFlushAsWriteBuffer'.
```
jorisdral, line 44: Outdated comment

src-extras/Database/LSMTree/Extras/MergingTreeData.hs
```diff
@@ -0,0 +1,358 @@
...
+  -> UniqCounter IO
+  -> SerialisedMergingTreeData
+  -> (Ref (MergingTree IO h) -> IO a)
+  -> IO a
+withMergingTree hfs hbio resolve indexType path counter mrd = do
+    bracket
+      (unsafeFlushMergingTree hfs hbio resolve indexType path counter mrd)
+      releaseRef
+
+-- | Flush serialised run data to disk as if it were a write buffer.
```
jorisdral, line 60: Outdated comment

src-extras/Database/LSMTree/Extras/MergingTreeData.hs
```diff
@@ -0,0 +1,358 @@
...
+        MT.PreExistingRun r         -> releaseRef r
+        MT.PreExistingMergingRun mr -> releaseRef mr
+      forM_ mt $ releaseRef
+      return tree
+    PendingUnionMergeData mtds -> do
+      mts <- forM mtds $
+        unsafeFlushMergingTree hfs hbio resolve indexType path counter
+      tree <- MT.newPendingUnionMerge mts
+      forM_ mts $ releaseRef
+      return tree
```
jorisdral, line 105: Use `with*`-style combinators here

src-extras/Database/LSMTree/Extras/MergingTreeData.hs
```diff
@@ -0,0 +1,358 @@
...
+    PendingLevelMergeData prs t -> do
+      assertI "pending level merges have at least one input" $
+        length prs + length t > 0
+      for_ prs $ \case
+        PreExistingRunData        _r -> Right ()
+        PreExistingMergingRunData mr -> mergingRunDataInvariant mr
+      for_ t mergingTreeDataInvariant
+    PendingUnionMergeData ts -> do
+      assertI "pending union merges are non-trivial (at least two inputs)" $
+        length ts > 1
```
jorisdral, line 144: Minor nit: I find `>=2` clearer, because that's also how you describe the invariant in the text

src-extras/Database/LSMTree/Extras/MergingRunData.hs
```diff
@@ -0,0 +1,217 @@
...
+    CompletedMergeData t MR.NumRuns (RunData k v b)
+  | OngoingMergeData t [NonEmptyRunData k v b]  -- ^ at least 2 inputs
+  deriving stock (Show, Eq)
+
+-- | See @mergeInvariant@ in the prototype.
+mergingRunDataInvariant :: MergingRunData t k v b -> Either String ()
+mergingRunDataInvariant = \case
+    CompletedMergeData _ (MR.NumRuns n) _ ->
+      assertI "completed merges are non-trivial (at least two inputs)" $
+        n > 1
```
jorisdral, line 120: Same minor nit about `>` vs. `>=`

src-extras/Database/LSMTree/Extras/MergingRunData.hs
```diff
@@ -0,0 +1,217 @@
...
+    OngoingMergeData _ rds -> do
+      assertI "inputs to ongoing merges aren't empty" $
+        all nonEmptyRunDataInvariant rds
+      assertI "ongoing merges are non-trivial (at least two inputs)" $
+        length rds > 1
```
jorisdral, line 125: Same minor nit about `>` vs. `>=`

src-extras/Database/LSMTree/Extras/MergingTreeData.hs
```diff
@@ -0,0 +1,358 @@
...
+-- | See @treeInvariant@ in prototype.
+mergingTreeDataInvariant :: MergingTreeData k v b -> Either String ()
+mergingTreeDataInvariant = \case
+    CompletedTreeMergeData _rd ->
+      Right ()
+    OngoingTreeMergeData mr ->
+      mergingRunDataInvariant mr
+    PendingLevelMergeData prs t -> do
+      assertI "pending level merges have at least one input" $
+        length prs + length t > 0
```
jorisdral, line 137: Same minor nit about `>` vs. `>=`

src-extras/Database/LSMTree/Extras/MergingTreeData.hs
```diff
@@ -0,0 +1,358 @@
...
+    assertI msg False = Left msg
+    assertI _   True  = Right ()
+
+mapMergingTreeData ::
+     Ord k'
+  => (k -> k') -> (v -> v') -> (b -> b')
+  -> MergingTreeData k v b -> MergingTreeData k' v' b'
+mapMergingTreeData f g h = go
+  where
+    go = \case
```
jorisdral, line 156: You can inline the `go`

src-extras/Database/LSMTree/Extras/MergingTreeData.hs
```diff
@@ -0,0 +1,358 @@
...
+      (unsafeFlushMergingTree hfs hbio resolve indexType path counter mrd)
+      releaseRef
+
+-- | Flush serialised run data to disk as if it were a write buffer.
+--
+-- This might leak resources if not run with asynchronous exceptions masked.
+-- Use helper functions like 'withMergingTree' instead.
+--
+-- Use of this function should be paired with a 'releaseRef'.
+unsafeFlushMergingTree ::
```
jorisdral, line 66: Same comment about the use of the term "flush" here

src-extras/Database/LSMTree/Extras/MergingTreeData.hs
```diff
@@ -0,0 +1,358 @@
...
+              numPreExisting <- QC.chooseInt (0, min 20 (n - 2))
+              preExisting <- QC.vectorOf numPreExisting $
+                genPreExistingRunData genKey genVal genBlob
+              tree <- go (n - numPreExisting - 1)
+              return (PendingLevelMergeData preExisting (Just tree)))
+          , (2, do
+              -- pending union merge
+              ns <- QC.shuffle =<< arbitraryPartition2 n
+              PendingUnionMergeData <$> traverse go ns)
+          ]
```
jorisdral, line 269: Same comment about the number of children: on't we test the most interesting cases with `2` or `3` children?

mheinzel:
> "the tests" meaning the generator invariant tests, I take it?

Mainly all the shrinker invariant tests, but for `MergingTreeData` in particular, even just generating them is kind of slow (see the test output in the PR description).

Instead of decreasing the size of these particular tests, I should probably change the instance to generate smaller trees, since any property on trees will otherwise be quite slow.

src-extras/Database/LSMTree/Extras/MergingTreeData.hs
```diff
@@ -0,0 +1,358 @@
...
+    PreExistingRunData SerialisedKey SerialisedValue SerialisedBlob
+
+serialiseMergingTreeData ::
+     (SerialiseKey k, SerialiseValue v, SerialiseValue b)
+  => MergingTreeData k v b -> SerialisedMergingTreeData
+serialiseMergingTreeData =
+    mapMergingTreeData serialiseKey serialiseValue serialiseBlob
+
+{-------------------------------------------------------------------------------
+  QuickCheck
```
jorisdral, line 191: There is quite a bit of duplication here with the `ScheduledMergesTest` module. Not saying that's a problem, but we could put a TODO somewhere to see if we can reduce duplication, maybe by reusing the `T` type from the prototypes tests?

test/Test/Database/LSMTree/Generators.hs
```diff
@@ -62,6 +65,11 @@ tests = testGroup "Test.Database.LSMTree.Generators" [
         prop_arbitraryAndShrinkPreserveInvariant
           labelRunData
           (const True)  -- no invariant
+     ++ [ testProperty "withRun doesn't leak resources" $ \rd ->
+            QC.ioProperty $
+              withSimHasBlockIO propNoOpenHandles MockFS.empty $ \hfs hbio _ ->
+                prop_withRunDoesntLeak hfs hbio rd
+        ]
```
jorisdral, line 72: I'd use the name of the property as the test name, i.e., `prop_withRunDoesntLeak`, because it makes it easier to run them from the command line. Similarly for other properties below

test/Test/Database/LSMTree/Generators.hs
```diff
@@ -106,3 +124,42 @@ prop_distributionKOps kops' =
...
+  -> IO Property
+prop_withMergingRunDoesntLeak hfs hbio mrd = do
+    let index = Index.Ordinary
+    let path = FS.mkFsPath []
+    counter <- newUniqCounter 0
+    withMergingRun hfs hbio resolveVal index path counter mrd $ \_mr -> do
+      return (QC.property True)
+
+prop_withMergingTreeDoesntLeak ::
+     forall h.
```
jorisdral, line 152:
```suggestion
```

test/Test/Database/LSMTree/Generators.hs
```diff
@@ -106,3 +124,42 @@ prop_distributionKOps kops' =
...
+  -> FS.HasBlockIO IO h
+  -> SerialisedRunData
+  -> IO Property
+prop_withRunDoesntLeak hfs hbio rd = do
+    let index = Index.Ordinary
+    withRun hfs hbio index (simplePath 0) rd $ \_run -> do
+      return (QC.property True)
+
+prop_withMergingRunDoesntLeak ::
+     forall h.
```
jorisdral, line 139:
```suggestion

```

src-extras/Database/LSMTree/Extras/MergingRunData.hs
jorisdral: Related to https://github.com/IntersectMBO/lsm-tree/pull/592#issuecomment-2687564622

jorisdral (changes requested):
> The tests are quite slow even after halving the maximum size. The main issue is that RunData has a lot of possible shrinks (sometimes over 100k), which gets even worse when there are dozens of them in the tree.

Normally with shrinking test cases it's not extremely important to not generate too many shrinks. Rather, the ordering of shrunk values is more important, so that we don't evaluate too many before we find the next, smaller, failing test case. But yeah, it would be nice to reduce the number of shrink values for these "arbitrary/shrinking preserve invariant" tests. I've commented elsewhere that we could maybe reduce the tree size by generating fewer children in the merging tree

test/Test/Util/Arbitrary.hs
```diff
@@ -1,34 +1,41 @@
...
 import           Test.QuickCheck
 import           Test.Tasty (TestTree)
 import           Test.Tasty.QuickCheck (testProperty)
 
 prop_arbitraryAndShrinkPreserveInvariant ::
-    forall a. (Arbitrary a, Show a) => (a -> Bool) -> [TestTree]
-prop_arbitraryAndShrinkPreserveInvariant =
-    prop_forAllArbitraryAndShrinkPreserveInvariant arbitrary shrink
+     forall a prop. (Arbitrary a, Show a, Testable prop)
+  => (a -> Property -> Property) -> (a -> prop) -> [TestTree]
```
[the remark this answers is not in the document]
mheinzel: Allowing the caller to produce any `prop` is indeed more convenient for them (which part of the reason why I generalised `(a -> Bool)` to `(a -> prop)`. But for the other function, that doesn't really apply, since `prop` appears in contravariant position there. 

We could make the `prop` there the same as in the predicate: `(a -> prop1 -> Property) -> (a -> prop) -> [TestTree]`. But if the caller e.g. wants to use a `a -> Bool` predicate, they are now forced to provide a `a -> Bool -> Property`, which doesn't seem more useful and in practice just means they'll have to call `QC.property` themselves somewhere.

And if we make it a separate `prop1`, then this function becomes impossible to implement, since we get a `prop` out of the predicate and then have to apply the `prop1 -> Property` tagging function to it, with no way of converting between these two opaque types.

test/Test/Util/Arbitrary.hs
```diff
@@ -1,34 +1,41 @@
...
     [ testProperty "Arbitrary satisfies invariant" $
-            property $ forAllShrink gen shr inv
+        forAllShrink gen shr $ \x ->
+          tag x $ property $ inv x
     , testProperty "Shrinking satisfies invariant" $
-            property $ forAll gen $ \x ->
-                         case shr x of
-                           [] -> label "no shrinks" $ property True
-                           xs -> forAll (growingElements xs) inv
+        -- if we find out that the shrinker is broken, we don't want to use it!
```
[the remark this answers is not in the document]
mheinzel: Yes. I can clarify that it is about `forAll` vs. `forAllShrink`.

src-extras/Database/LSMTree/Extras/RunData.hs
```diff
@@ -151,54 +186,85 @@ withSerialisedWriteBuffer hfs hbio wbPaths wb wbb =
...
--- | Like 'simplePath', but for a list.
-simplePaths :: [Int] -> [RunFsPaths]
-simplePaths ns = fmap simplePath ns
+-- | A collection of arbitrary key\/value pairs that are suitable for creating
+-- 'Run's.
+--
+-- Note: 'b ~ Void' should rule out blobs.
+newtype NonEmptyRunData k v b =
+    NonEmptyRunData { unNonEmptyRunData :: RunData k v b }
+  deriving stock (Eq, Show)
```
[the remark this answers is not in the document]
mheinzel: I can try that, but it adds a dependency and we will basically always turn this into a regular `RunData` immediately after it was generated, so just having a newtype is kind of nice.

test/Test/Database/LSMTree/Generators.hs
```diff
@@ -55,6 +58,24 @@ tests = testGroup "Test.Database.LSMTree.Generators" [
         [ testProperty "prop_distributionKOps" $
             prop_distributionKOps
         ]
+    , testGroup "RunData" $
+        prop_arbitraryAndShrinkPreserveInvariant
+          labelRunData
+          (const True)  -- no invariant
```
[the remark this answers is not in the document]
mheinzel: I'll add `noInvariant`, similar to `noTags`.

src-extras/Database/LSMTree/Extras/MergingRunData.hs
mheinzel: Ah, thanks. I even replaced the `unsafeFlushRun`, but missed the `Run`.

src-extras/Database/LSMTree/Extras/MergingRunData.hs
mheinzel: or `unsafeCreateMergingRun`?

src-extras/Database/LSMTree/Extras/MergingRunData.hs
mheinzel: Since the ones above are just a number without real impact, I'm hesitant to throw them into the same counter, since they can mask issues with the distribution of the "real" input runs. I can add a separate "completed merging run inputs" one, but I'm not sure it matters enough to pull its weight.

src-extras/Database/LSMTree/Extras/MergingRunData.hs
mheinzel: I would not make it impossible to have more than 3 runs, it might matter somewhere. But we can make it skew towards smaller values, and on my branch I had just tweaked it to `QC.chooseInt (2, max 2 (s * 8 `div` 100))` to still allow up to 8 runs, but have `57.1% 2 <= n < 4`.  However, it doesn't significantly impact the test runtime, since we `QC.scale (`div` n)` below, so on average we still get the same number of entries.

test/Test/Database/LSMTree/Generators.hs
mheinzel: I followed `prop_arbitraryAndShrinkPreserveInvariant`, so the test output looks consistent. We can change it there, too, of course.

src-extras/Database/LSMTree/Extras/RunData.hs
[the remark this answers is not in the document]
jorisdral: It's a test suite so an extra dependency is not a problem

src-extras/Database/LSMTree/Extras/RunData.hs
```diff
@@ -103,14 +114,38 @@ unsafeFlushAsWriteBuffer ::
...
     wbblobs <- WBB.new fs blobpath
     wb <- WB.fromMap <$> traverse (traverse (WBB.addBlob fs wbblobs)) m
     run <- Run.fromWriteBuffer fs hbio CacheRunData (RunAllocFixed 10) indexType
                                fsPaths wb wbblobs
     releaseRef wbblobs
     return run
 
+-- | Like 'unsafeFlushAsWriteBuffer', but uses a 'UniqCounter' to determine
+-- the 'RunFsPaths', based on some file path.
+unsafeFlushRun ::
```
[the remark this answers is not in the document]
mheinzel: > a version of `unsafeFlushRun` that creates its own local `UniqCounter`, for tests where the file names won't conflict if we do that

Do you mean something like `unsafeFlushAsWriteBuffer _ _ _ simplePath`? I think then the main difference would then be which version is implemented in terms of which.

src-extras/Database/LSMTree/Extras/MergingTreeData.hs
mheinzel: Here it does indeed matter and I landed on distribution between 0 and 6 pre-existing run, but skewing towards smaller numbers.

src-extras/Database/LSMTree/Extras/MergingRunData.hs
mheinzel: It's not as critical here, since it's just for testing (and we already have the same issue for `WBB.new` in `RunData`), but it's easy to do, so I think it's a good suggestion. :+1:

src-extras/Database/LSMTree/Extras/RunData.hs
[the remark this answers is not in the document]
mheinzel: The `...At` versions with `simplePath` should be convenient enough for tests where conflicts aren't an issue.

test/Test/Database/LSMTree/Generators.hs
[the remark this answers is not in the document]
mheinzel: Done.

src-extras/Database/LSMTree/Extras/MergingRunData.hs
mheinzel: I went with that.

src-extras/Database/LSMTree/Extras/MergingRunData.hs
mheinzel: Done, also fixed the WBB issue in `RunData`.

src-extras/Database/LSMTree/Extras/MergingTreeData.hs
mheinzel: How do you feel about the `chooseIntSkewed` I went with?

src-extras/Database/LSMTree/Extras/MergingTreeData.hs
mheinzel: I added a TODO.

test/Test/Database/LSMTree/Generators.hs
mheinzel: I probably should have done it in #604, won't do it here.

test/Test/Util/Arbitrary.hs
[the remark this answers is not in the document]
jorisdral: You can change `(a -> Property -> Property)` to `(forall prop'. Testable prop' => a -> prop' -> Property)`, and that would work, but then all the functions you pass to it have to be `prop' -> Property` instead of `Property -> Property` and that's annoying too. Let's keep it as is

src-extras/Database/LSMTree/Extras/MergingRunData.hs
jorisdral: *has large k/op* is not the only affected metric, in general everything just gets smaller, even keys

Something like this would work:
```haskell
          rs <- QC.vectorOf n $ do
            s <- getSize
            QC.scale (`div` n) $
            -- scaled, so overall size is similar to a single RunData
            genNonEmptyRunData (resize s genKey) (resize s genVal) (resize s genBlob)
```
Not saying this is definitely necessarily, but it's relatively simple, no?

src-extras/Database/LSMTree/Extras/RunData.hs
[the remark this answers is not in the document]
jorisdral: My point is that maybe we don't need the function where we pass in a `RunFsPaths`, but it looks okay as is

test/Test/Database/LSMTree/Generators.hs
```diff
@@ -119,3 +137,40 @@ labelTestKOps kops' =
...
+     FS.HasFS IO h
+  -> FS.HasBlockIO IO h
+  -> SerialisedMergingTreeData
+  -> IO Property
+prop_withMergingTreeDoesntLeak hfs hbio mrd = do
+    let index = Index.Ordinary
+    let path = FS.mkFsPath []
+    counter <- newUniqCounter 0
+    withMergingTree hfs hbio resolveVal index path counter mrd $ \_tree -> do
+      return (QC.property True)
```
jorisdral, line 173: Maybe add a note/TODO that this only test that nothing is leaked on the happy path, and that we'd have to use `simErrorHasBlockIO` to test nothing is leaked on the unhappy path

src-extras/Database/LSMTree/Extras/RunData.hs
```diff
@@ -221,6 +311,29 @@ shrinkRunData shrinkKey shrinkVal shrinkBlob =
...
+  => (k -> [k])
+  -> (v -> [v])
+  -> (b -> [b])
+  -> NonEmptyRunData k v b
+  -> [NonEmptyRunData k v b]
+shrinkNonEmptyRunData shrinkKey shrinkVal shrinkBlob =
+      map (NonEmptyRunData . NEMap.fromList)
+    . liftShrink (liftShrink2 shrinkKey (liftShrink2 shrinkVal shrinkBlob))
+    . NEMap.toList
+    . unNonEmptyRunData
```
jorisdral, line 335:
```suggestion
genNonEmptyRunData ::
     Ord k
  => Gen k
  -> Gen v
  -> Gen b
  -> Gen (NonEmptyRunData k v b)
genNonEmptyRunData genKey genVal genBlob =
    NonEmptyRunData <$> liftArbitrary2NEMap genKey (liftArbitrary2 genVal genBlob)

shrinkNonEmptyRunData ::
     Ord k
  => (k -> [k])
  -> (v -> [v])
  -> (b -> [b])
  -> NonEmptyRunData k v b
  -> [NonEmptyRunData k v b]
shrinkNonEmptyRunData shrinkKey shrinkVal shrinkBlob =
      fmap NonEmptyRunData
    . liftShrink2NEMap shrinkKey (liftShrink2 shrinkVal shrinkBlob)
    . unNonEmptyRunData

-- | We cannot implement 'Arbitrary2' since we have constraints on @k@.
liftArbitrary2NEMap :: Ord k => Gen k -> Gen v -> Gen (NEMap k v)
liftArbitrary2NEMap genk genv = NEMap.fromList <$>
    liftArbitrary (liftArbitrary2 genk genv)

-- | We cannot implement 'Arbitrary2' since we have constraints @k@.
liftShrink2NEMap :: Ord k => (k -> [k]) -> (v -> [v]) -> NEMap k v -> [NEMap k v]
liftShrink2NEMap shrinkk shrinkv m = NEMap.fromList <$>
    liftShrink (liftShrink2 shrinkk shrinkv) (NEMap.toList m)
```

jorisdral (approved):
LGTM!

src-extras/Database/LSMTree/Extras/MergingRunData.hs
mheinzel: Ah, true, that's pretty straightforward.

src-extras/Database/LSMTree/Extras/MergingTreeData.hs
```diff
@@ -0,0 +1,426 @@
...
+  -> HasBlockIO IO h
+  -> ResolveSerialisedValue
+  -> IndexType
+  -> FS.FsPath
+  -> UniqCounter IO
+  -> SerialisedMergingTreeData
+  -> IO (Ref (MergingTree IO h))
+unsafeCreateMergingTree hfs hbio resolve indexType path counter = go
+  where
+    go = \case
```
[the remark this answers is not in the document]
mheinzel: I can inline it, but then I need to pass all the arguments to each recursive call, which is kind of noisy and often requires a line break.

src-extras/Database/LSMTree/Extras/MergingTreeData.hs
mheinzel: It has a noticeable, but relatively small effect. Changing the range to `(0, 3)` lead to about 25% faster tests. I decided to generate 50% shorter runs instead (see last commit), halving the runtime.

PR #598, opened 2025-02-27
Public API: add union debt/credits types and related functions

jorisdral:
The implementation is left open for now. I'll follow up this PR with a PR to add test support for union debt/supply.

### DIFF
--- a/prototypes/ScheduledMerges.hs
+++ b/prototypes/ScheduledMerges.hs
@@ -67,7 +67,9 @@ module ScheduledMerges (
     NominalDebt(..),
     Run,
     runSize,
+    UnionCredits (..),
     supplyCreditsMergingTree,
+    UnionDebt(..),
     remainingDebtMergingTree,
     mergek,
     mergeBatchSize,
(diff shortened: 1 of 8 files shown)


prototypes/ScheduledMerges.hs
```diff
@@ -840,25 +842,47 @@ unions lsms = do
...
+  deriving stock (Show, Eq, Ord)
+  deriving newtype Num
+
+-- | Supply union credits to reduce union debt.
+--
+-- Supplying union credits leads to union merging work being performed in
+-- batches. This reduces the union debt returned by 'remainingUnionDebt'. Union
+-- debt will be reduced by /at least/ the number of supplied union credits. It
+-- is therefore advisable to query 'remainingUnionDebt' every once in a while to
+-- see what the current debt is.
```
mheinzel, line 879: Nice clarification!

Oh, and there could be a short note explaining that the result are leftover credits and if they are non-zero, then the union is paid off. Both here and in the public API.

mheinzel (approved):
Nice!

PR #599, opened 2025-02-27
Add general index benchmarks

jeltsch:
This pull request adds index benchmarks that are independent of index type and instantiates them for compact and ordinary indexes.

### DIFF
--- a/bench/micro/Bench/Database/LSMTree/Internal/Index.hs
+++ b/bench/micro/Bench/Database/LSMTree/Internal/Index.hs
@@ -1,0 +1,123 @@
+{-# LANGUAGE CPP #-}
+
+module Bench.Database.LSMTree.Internal.Index (benchmarks) where
+
+import           Control.Category ((>>>))
+import           Control.DeepSeq (rnf)
+import           Control.Monad.ST.Strict (runST)
+import           Criterion.Main (Benchmark, Benchmarkable, bench, bgroup, env,
+                     whnf)
+#if __GLASGOW_HASKELL__ < 910
+import           Data.List (foldl')
+#endif
+import           Database.LSMTree.Extras.Generators (getKeyForIndexCompact,
+                     mkPages, toAppends)
+                     -- also for @Arbitrary@ instantiation of @SerialisedKey@
+import           Database.LSMTree.Extras.Index (Append, append)
+import           Database.LSMTree.Internal.Index (Index,
+                     IndexType (Compact, Ordinary), newWithDefaults, search,
+                     unsafeEnd)
+import           Database.LSMTree.Internal.Serialise
+                     (SerialisedKey (SerialisedKey))
+import           Test.QuickCheck (choose, vector)
+import           Test.QuickCheck.Gen (Gen (MkGen))
+import           Test.QuickCheck.Random (mkQCGen)
+
+-- * Benchmarks
+
+benchmarks :: Benchmark
+benchmarks = bgroup "Bench.Database.LSMTree.Internal.Index" $
+             map (uncurry benchmarksForSingleType)          $
+             [("Compact", Compact), ("Ordinary", Ordinary)]
+    where
+
+    benchmarksForSingleType :: String -> IndexType -> Benchmark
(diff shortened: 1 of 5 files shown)


bench/micro/Bench/Database/LSMTree/Internal/Index.hs
```diff
@@ -0,0 +1,119 @@
...
+{-|
+    Constructs append operations whose serialised keys conform to the key size
+    constraint of compact indexes.
+-}
+appendsForIndexCompact :: Int      -- ^ Number of keys used in the construction
+                       -> [Append] -- ^ Constructed append operations
+appendsForIndexCompact = keysForIndexCompact                >>>
+                         mkPages 0.03 (choose (0, 16)) 0.01 >>>
+                         generated                          >>>
+                         toAppends
```
jorisdral, line 71: I personally normally avoid using `QuickCheck` generators because they are often in flux, and because they are primarily tailored towards testing, not benchmarking. It can be more future proof to write these functions with `System.Random` and related functions

jorisdral (approved):

bench/micro/Bench/Database/LSMTree/Internal/Index.hs
jeltsch: I’m also not completely satisfied with using QuickCheck generators for benchmarks. That said, by using `System.Random` directly, we lose the ability to use existing utilities, like we use `mkPages` above. I’d like to leave the above code as it is for now. It might be worthwhile, though, to generally revisit the uses of QuickCheck generators in our benchmarks.

bench/micro/Bench/Database/LSMTree/Internal/Index.hs
```diff
@@ -0,0 +1,119 @@
...
+
+-- | The action to be performed by a search benchmark.
+searchBenchmarkable :: Index -> [SerialisedKey] -> Benchmarkable
+searchBenchmarkable index = whnf $ foldl' (\ _ key -> rnf (search key index)) ()
+
+-- ** Incremental construction
+
+-- | Constructs append operations to be used in index construction.
+incrementalConstructionAppends
+    :: Int      -- ^ Number of keys used in the construction
```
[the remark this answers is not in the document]
jorisdral: @jeltsch did you forget to resolve this?

bench/micro/Bench/Database/LSMTree/Internal/Index.hs
jorisdral: > That said, by using `System.Random` directly, we lose the ability to use existing utilities, like we use `mkPages` above. 

That's not necessarily true. You should run the `mkPages` `Gen` with a seed, yes, but you can generate the keys with `System.Random`. See the `Index.Compact` benchmarks

bench/micro/Bench/Database/LSMTree/Internal/Index.hs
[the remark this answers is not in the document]
jeltsch: I thought that this wouldn’t be so important, given that it was labeled a “suggested” change and the pull request was already approved as it were.

Generally, I think that having this sort of hanging indentation by not putting `::` above the arrows makes for an inferior layout. I know that your editor and GitHub seem to have problems with a layout like the present one when it comes to syntax highlighting, but I think we shouldn’t let current bugs of tools influence our layout.

Well, I can make the change. Should I?

bench/micro/Bench/Database/LSMTree/Internal/Index.hs
jeltsch: So you’re advocating a mixed-approach with some data being generated by directly using `System.Random` and other by using QuickCheck?

In the compact-index benchmark module, I can only see one use of `mkPages`, the one in `constructionEnv`. However, there the generator constructed by `mkPages` is run via `generate`, which means that it uses a seed produced by the global random number generator, not a seed specified in the source file. That said, I could still use my `generated` function for generators to work with fixed seeds.

Should I change the general index benchmarks to use that mixed approach where just the page data is generated using QuickCheck?

PR #604, opened 2025-03-03
Improvements to generators and shrinkers

mheinzel:
# Description

General improvements. Some things are split out of #592.

Best reviewed commit by commit. The commit messages include test output to show their changes' impact.

### DIFF
--- a/src-extras/Database/LSMTree/Extras/Generators.hs
+++ b/src-extras/Database/LSMTree/Extras/Generators.hs
@@ -489,6 +489,7 @@ deriving newtype instance Arbitrary SerialisedBlob
 
 newtype LargeRawBytes = LargeRawBytes RawBytes
   deriving stock Show
+  deriving newtype NFData
 
 instance Arbitrary LargeRawBytes where
   arbitrary = genRawBytesSized (4096*3) >>= fmap LargeRawBytes . genSlice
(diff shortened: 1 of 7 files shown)


src-extras/Database/LSMTree/Extras/Generators.hs
```diff
@@ -454,8 +457,31 @@ packRawBytesPinnedOrUnpinned True  = \ws ->
       return mba
 
 shrinkRawBytes :: RawBytes -> [RawBytes]
-shrinkRawBytes (RawBytes pvec) = [ RawBytes (VP.fromList ws)
-                                 | ws <- QC.shrink (VP.toList pvec) ]
+shrinkRawBytes (RawBytes pvec) =
+    [ RawBytes pvec'
+    | pvec' <- shrinkVec (take 1 . QC.shrink) pvec
+                          -- no need to try harder shrinking individual bytes
```
jorisdral, line 463: My first thought is that I disagree. It would be nice to make `prop_forAllArbitraryAndShrinkPreserveInvariant` faster, but I'd argue getting good shrinks should be the goal, and arguably you shouldn't restrict the shrink space too much to do that. Maybe you could give an idea about why `take 1 . QC.shrink` is sufficient in your opinion?

jorisdral (changes requested):
Requesting changes, but only because I think we should agree on https://github.com/IntersectMBO/lsm-tree/pull/604/files#r1977234085 first

src-extras/Database/LSMTree/Extras/Generators.hs
jorisdral: Arguably shrinking the bytes of serialised keys is more important than shrinking the bytes of serialised values. Maybe we should have different shrinkers for the two?

jorisdral (approved):

src-extras/Database/LSMTree/Extras/Generators.hs
mheinzel: We now try up to two shrinks (0 and ``x `div` 2``) to at least allow getting to smaller numbers via iterative shrinking. Also, if we really encounter a test case where the shrinking doesn't seem ideal, we can still go back to a more exhaustive one and re-run the test.

PR #611, opened 2025-03-07
QLS: add `SupplyPortionOfDebt`

jorisdral:
And some miscellaneous QLS tagger / generator improvements

### DIFF
--- a/test/Test/Database/LSMTree/StateMachine.hs
+++ b/test/Test/Database/LSMTree/StateMachine.hs
@@ -92,12 +92,12 @@ import           Data.List.NonEmpty (NonEmpty (..))
 import qualified Data.List.NonEmpty as NE
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
-import           Data.Maybe (catMaybes, fromJust, fromMaybe)
+import           Data.Maybe (catMaybes, fromJust, fromMaybe, isJust)
 import           Data.Monoid (First (..))
 import           Data.Primitive.MutVar
 import           Data.Set (Set)
 import qualified Data.Set as Set
-import           Data.Typeable (Proxy (..), Typeable, cast)
+import           Data.Typeable (Proxy (..), Typeable, cast, eqT)
 import qualified Data.Vector as V
 import qualified Database.LSMTree as R
 import           Database.LSMTree.Class (LookupResult (..), QueryResult (..))
@@ -1761,16 +1761,17 @@ arbitraryActionWithVars _ label ctx (ModelState st _stats) =
             Map.member (Model.tableID t) (Model.tables st)
       ]
 
-    genUnionedTableVar = QC.elements (unionedTableVars Model.IsUnionDescendant)
-    genNonUnionedTableVar = QC.elements (unionedTableVars Model.IsNotUnionDescendant)
+    genUnionDescendantTableVars = QC.elements unionDescendantTableVars
+    genNotUnionDescendantTableVars = QC.elements notUnionDescendantTableVars
 
-    -- | Variables for tables that are a (descendant of a) union table, or not.
-    unionedTableVars :: Model.IsUnionDescendant -> [Var h (WrapTable h IO k v b)]
-    unionedTableVars target =
-        [ v
+    unionDescendantTableVars, notUnionDescendantTableVars :: [Var h (WrapTable h IO k v b)]
+    (unionDescendantTableVars, notUnionDescendantTableVars) = partitionEithers $
+        [ case Model.isUnionDescendant t of
+            Model.IsUnionDescendant    -> Left v
+            Model.IsNotUnionDescendant -> Right v
         | v <- tableVars
-        , case lookupVar ctx v of
-            MTable t -> Model.isUnionDescendant t == target
+        , let t = case lookupVar ctx v of
+                MTable t' -> t'
         ]
 
     genCursorVar = QC.elements cursorVars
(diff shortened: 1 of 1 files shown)


test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -731,17 +731,31 @@ data Action' h a where
...
     -> Act' h R.UnionCredits
-  -- TODO: SupplyUnionCredits has no information about union debt, so the
-  -- credits we generate are arbitrary, and it would require precarious,
-  -- manual tuning to make sure the debt is ever paid off by an action
-  -- sequence. We can probably add a version of SupplyUnionCredits that
-  -- supplies a percentage (if not al) of the current or previously computed
-  -- debt. Note that the model has no accurate debt information, so we might
-  -- have to use variables.
+  -- | Alternative version of 'SupplyUnionCredits' that supplies a portion of
+  -- the a table's current uniondebt as union credits.
```
mheinzel, line 735:
```suggestion
  -- the table's current union debt as union credits.
```

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -731,17 +731,31 @@ data Action' h a where
...
     -> R.UnionCredits
     -> Act' h R.UnionCredits
-  -- TODO: SupplyUnionCredits has no information about union debt, so the
-  -- credits we generate are arbitrary, and it would require precarious,
-  -- manual tuning to make sure the debt is ever paid off by an action
-  -- sequence. We can probably add a version of SupplyUnionCredits that
-  -- supplies a percentage (if not al) of the current or previously computed
-  -- debt. Note that the model has no accurate debt information, so we might
-  -- have to use variables.
+  -- | Alternative version of 'SupplyUnionCredits' that supplies a portion of
```
mheinzel, line 734: Maybe it could even replace `SupplyUnionCredits`. Supplying absolute credits doesn't seem super meaningful without a frame of reference (how much debt there is). The distribution of portions seems easier to tune, as it automatically scales with union size.

Although absolute credits could sometimes allow for better shrinking if a specific number of supplied credits triggers an issue, but shrinking any of the union inputs would make `SupplyPortionOfDebt` result in a different number of credits. :shrug:

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -731,17 +731,31 @@ data Action' h a where
...
+  -- the a table's current uniondebt as union credits.
   --
+  -- 'SupplyUnionCredits' gets no information about union debt, so the union
+  -- credits we generate are arbitrary, and it would require precarious, manual
+  -- tuning to make sure the debt is ever paid off by an action sequence.
+  -- 'SupplyUnionCredits' supplies a portion (if not al) of the current debt, so
+  -- that unions are morelikely to finish during a sequence of actions.
```
mheinzel, line 741:
```suggestion
  -- that unions are more likely to finish during a sequence of actions.
```

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -1126,33 +1144,35 @@ instance ( Eq (Class.TableConfig h)
       Unions{}         -> OEither $ bimap OId (const OTable) result
       RemainingUnionDebt{} -> OEither $ bimap OId OId result
       SupplyUnionCredits{} -> OEither $ bimap OId OId result
+      SupplyPortionOfDebt{} -> OEither $ bimap OId OId result
```
mheinzel, line 1147: Ah, here we're doing some kind of semi-aligned style. I'm not fan of aligning for this reason, but if we do it in the rest of the module, maybe we should also do it here. Same in in the `IOSim` instance.

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -1900,6 +1983,9 @@ arbitraryActionWithVars _ label ctx (ModelState st _stats) =
           , (1, R.UnionCredits <$> QC.arbitrary)
           ]
 
+        -- TODO: tweak distribution once table unions are implemented
+        genPortion = Portion <$> QC.elements [1, 2, 3]
```
mheinzel, line 1987: Ah, nice, so no matter much much debt, we always have a chance to finish the union immediately.

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -2068,6 +2119,17 @@ shrinkAction'WithVars _ctx _st a = case a of
...
+    SupplyUnionCredits tableVar (R.UnionCredits x) -> [
+        Some $ SupplyUnionCredits tableVar (R.UnionCredits x')
+      | x' <- QC.shrink x
+      ]
+
+    SupplyPortionOfDebt tableVar (Portion x) -> [
+        Some $ SupplyPortionOfDebt tableVar (Portion x')
+      | x' <- QC.shrink x
+      , x' > 0
+      ]
```
mheinzel, line 2153: Interesting, this shrinks towards larger portions. E.g. supplying 1/3 of the debt shrinks to 1/2, which is counter-intuitive, but kind of makes sense. The other direction is probably trickier to get right.

We could also try shrinking to some absolute `SupplyUnionCredits`. The number of credits is kind of arbitrary, but it can't result in a loop and might be worth a try.

mheinzel (approved):
Looks good. Had some ideas, but looks completely reasonable.

test/Test/Database/LSMTree/StateMachine.hs
jorisdral: I think it's useful to keep `SupplyUnionCredits` because it broadens the "generator space" beyond supplying 1x 1/2x 1/3x the debt, and so we can shrink `SupplyPortionDebt` to `SupplyUnionCredits`. There are cases like you describe where shrinking to absolute credits might not work so well, but I also don't think it's entirely useless in cases where we are shrinking an independent action

test/Test/Database/LSMTree/StateMachine.hs
jorisdral: I did both! But I'm preferring shrinking to `SupplyUnionCredits` to shrinking towards larger portions. The former is more likely to actually provide us with a minimal counterexample, I thinks. Larger portions implies more work being done

PR #614, opened 2025-03-10
Enable unions in lockstep tests

mheinzel:
# Description

Should go on top of #608 and #609. We should also add some union-related labelling.

### DIFF
--- a/src/Database/LSMTree/Internal.hs
+++ b/src/Database/LSMTree/Internal.hs
@@ -1649,4 +1649,9 @@ supplyUnionCredits t credits = do
       RW.withWriteAccess (tableContent tEnv) $ \tableContent ->
         case tableUnionLevel tableContent of
           NoUnion -> pure (tableContent, credits) -- all leftovers
-          Union{} -> error "supplyUnionCredits: not yet implemented"
+          Union{}
+           | credits <= UnionCredits 0 -> pure (tableContent, UnionCredits 0)
+           --TODO: remove this 0 special case once the general case covers it.
+           -- We do not need to optimise the 0 case. It is just here to
+           -- simplify test coverage.
+           | otherwise -> error "supplyUnionCredits: not yet implemented"
--- a/src/Database/LSMTree/Internal.hs
+++ b/src/Database/LSMTree/Internal.hs
@@ -1630,8 +1630,10 @@ remainingUnionDebt :: (MonadSTM m, MonadThrow m) => Table m h -> m UnionDebt
 remainingUnionDebt t = do
     traceWith (tableTracer t) TraceRemainingUnionDebt
     withOpenTable t $ \tEnv -> do
-      RW.withReadAccess (tableContent tEnv) $ \_tableContent -> do
-        error "remainingUnionDebt: not yet implemented"
+      RW.withReadAccess (tableContent tEnv) $ \tableContent ->
+        case tableUnionLevel tableContent of
+          NoUnion -> pure (UnionDebt 0)
+          Union{} -> error "remainingUnionDebt: not yet implemented"
 
 -- | See 'Database.LSMTree.Normal.UnionCredits'.
 newtype UnionCredits = UnionCredits Int
@@ -1644,5 +1646,7 @@ supplyUnionCredits t credits = do
     traceWith (tableTracer t) $ TraceSupplyUnionCredits credits
     withOpenTable t $ \tEnv -> do
       -- TODO: should this be acquiring read or write access?
-      RW.withWriteAccess (tableContent tEnv) $ \_tableContent -> do
-        error "supplyUnionCredits: not yet implemented"
+      RW.withWriteAccess (tableContent tEnv) $ \tableContent ->
+        case tableUnionLevel tableContent of
+          NoUnion -> pure (tableContent, credits) -- all leftovers
+          Union{} -> error "supplyUnionCredits: not yet implemented"
--- a/test/Database/LSMTree/Model/Session.hs
+++ b/test/Database/LSMTree/Model/Session.hs
@@ -846,7 +846,7 @@ supplyUnionCredits ::
 supplyUnionCredits t@Table{..} c@(UnionCredits credits)
   | credits <= 0 = do
       _ <- guardTableIsOpen t
-      pure c
+      pure (UnionCredits 0) -- always 0, not negative
   | otherwise = do
       (updc, table) <- guardTableIsOpen t
       when (isUnionDescendant == IsUnionDescendant) $
--- a/test/Database/LSMTree/Model/Session.hs
+++ b/test/Database/LSMTree/Model/Session.hs
@@ -844,7 +844,9 @@ supplyUnionCredits ::
   -> UnionCredits
   -> m UnionCredits
 supplyUnionCredits t@Table{..} c@(UnionCredits credits)
-  | credits <= 0 = pure c
+  | credits <= 0 = do
+      _ <- guardTableIsOpen t
+      pure c
   | otherwise = do
       (updc, table) <- guardTableIsOpen t
       when (isUnionDescendant == IsUnionDescendant) $
--- a/test/Test/Database/LSMTree/StateMachine.hs
+++ b/test/Test/Database/LSMTree/StateMachine.hs
@@ -2221,7 +2221,7 @@ instance InterpretOp Op (ModelValue (ModelState h)) where
 data Stats = Stats {
     -- === Tags
     -- | Names for which snapshots exist
-    snapshotted        :: Set R.SnapshotName
+    snapshotted        :: !(Set R.SnapshotName)
     -- === Final tags (per action sequence, across all tables)
     -- | Number of succesful lookups and their results
   , numLookupsResults  :: {-# UNPACK #-} !(Int, Int, Int)
@@ -2242,13 +2242,13 @@ data Stats = Stats {
   , closedTables       :: !(Map Model.TableID Model.SomeTable)
     -- | The ultimate parents for each table. These are the 'TableId's of tables
     -- created using 'new' or 'open'.
-  , parentTable        :: Map Model.TableID [Model.TableID]
+  , parentTable        :: !(Map Model.TableID [Model.TableID])
     -- | Track the interleavings of operations via different but related tables.
     -- This is a map from each ultimate parent table to a summary log of which
     -- tables (derived from that parent table via duplicate or union) have had
     -- \"interesting\" actions performed on them. We record only the
     -- interleavings of different tables not multiple actions on the same table.
-  , dupTableActionLog  :: Map Model.TableID [Model.TableID]
+  , dupTableActionLog  :: !(Map Model.TableID [Model.TableID])
     -- | The subset of tables (open or closed) that were created as a result
     -- of a union operation. This can be used for example to select subsets of
     -- the other per-table tracking maps above, or the state from the model.
--- a/test/Test/Database/LSMTree/StateMachine.hs
+++ b/test/Test/Database/LSMTree/StateMachine.hs
@@ -1755,7 +1755,7 @@ arbitraryActionWithVars _ label ctx (ModelState st _stats) =
       concat
         [ genActionsSession
         , genActionsTables
-        , genUnionActions
+        , genActionsUnion
         , genActionsCursor
         , genActionsBlobRef
         ]
@@ -1896,8 +1896,10 @@ arbitraryActionWithVars _ label ctx (ModelState st _stats) =
         | let genErrors = pure Nothing -- TODO: generate errors
         ]
      ++ [ (5,  fmap Some $ (Action <$> genErrors <*>) $
-            RangeLookup <$> genRange <*> genTableVar)
+            RangeLookup <$> genRange <*> genNotUnionDescendantTableVar)
+            -- TODO: enable range lookups on tables with unions
         | let genErrors = pure Nothing -- TODO: generate errors
+        , not (null notUnionDescendantTableVars) -- TODO: enable range lookups on tables with unions
         ]
      ++ [ (10, fmap Some $ (Action <$> genErrors <*>) $
             Updates <$> genUpdates <*> genTableVar)
@@ -1916,9 +1918,11 @@ arbitraryActionWithVars _ label ctx (ModelState st _stats) =
         | let genErrors = pure Nothing -- TODO: generate errors
         ]
      ++ [ (3,  fmap Some $ (Action <$> genErrors <*>) $
-            NewCursor <$> QC.arbitrary <*> genTableVar)
+            NewCursor <$> QC.arbitrary <*> genNotUnionDescendantTableVar)
+            -- TODO: cursors for tables with unions
         | length cursorVars <= 5 -- no more than 5 cursors at once
         , let genErrors = pure Nothing -- TODO: generate errors
+        , not (null notUnionDescendantTableVars) -- TODO: cursors for tables with unions
         ]
      ++ [ (2,  fmap Some $ (Action <$> genErrors <*>) $
             CreateSnapshot <$> genCorruption <*> pure label <*> genUnusedSnapshotName <*> genTableVar)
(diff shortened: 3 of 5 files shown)


test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -1970,8 +1956,14 @@ arbitraryActionWithVars _ label ctx (ModelState st _stats) =
...
-     ++ [ (2,  fmap Some $ (Action <$> genErrors <*>) $
-            Unions <$> gen2or3TableVars)
+     ++ [ (2,  fmap Some $ (Action <$> genErrors <*>) $ do
+            -- Generate at least a 2-way union, and at most a 3-way union.
+            --
+            -- Tests for 0-way and 1-way unions are included in the UnitTests
+            -- module. n-way unions for n>3 lead to large unions, which are less
+            -- likely to be finished before the end of an action sequence.
+            n <- QC.chooseInt (2, 3)
+            Unions . NE.fromList <$> QC.vectorOf n genTableVar)
```
jorisdral, line 1966: Yeah this is an improvement over what `gen2or3TableVars` was doing :stuck_out_tongue:

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -2005,35 +1997,17 @@ arbitraryActionWithVars _ label ctx (ModelState st _stats) =
...
-            -- TODO: replace union actions on non-union tables with a few unit
-            -- tests?
-         ++ [ (1, genNotUnionDescendantTableVars)
-            | not (null notUnionDescendantTableVars)
-            ]
+        -- TODO: replace union actions on non-union tables with unit tests?
+        genUnionTableVar = QC.frequency [
+            (9 * length unionDescendantTableVars,    genUnionDescendantTableVar)
+          , (1 * length notUnionDescendantTableVars, genNotUnionDescendantTableVar)
+          ]
```
jorisdral, line 2006: So the idea is to skew more towards union descendants?

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -1901,8 +1901,10 @@ arbitraryActionWithVars _ label ctx (ModelState st _stats) =
         | let genErrors = pure Nothing -- TODO: generate errors
         ]
      ++ [ (5,  fmap Some $ (Action <$> genErrors <*>) $
-            RangeLookup <$> genRange <*> genTableVar)
+            RangeLookup <$> genRange <*> genNotUnionDescendantTableVar)
+            -- TODO: enable range lookups on tables with unions
         | let genErrors = pure Nothing -- TODO: generate errors
+        , not (null notUnionDescendantTableVars)
```
jorisdral, line 1907: Minor nit: I'd probably reiterate the TODO on the `not (null ...)` case just to be sure we don't forget to remove it

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -1921,9 +1923,11 @@ arbitraryActionWithVars _ label ctx (ModelState st _stats) =
         | let genErrors = pure Nothing -- TODO: generate errors
         ]
      ++ [ (3,  fmap Some $ (Action <$> genErrors <*>) $
-            NewCursor <$> QC.arbitrary <*> genTableVar)
+            NewCursor <$> QC.arbitrary <*> genNotUnionDescendantTableVar)
+            -- TODO: cursors for tables with unions
         | length cursorVars <= 5 -- no more than 5 cursors at once
         , let genErrors = pure Nothing -- TODO: generate errors
+        , not (null notUnionDescendantTableVars)
```
jorisdral, line 1930: Same here

jorisdral (approved):
Nice nice!

dcoutts:
I see my changes raced with Joris's review, but I think I happened to address most of his comments.

test/Test/Database/LSMTree/StateMachine.hs
dcoutts: Yes, though in a follow-up patch I made this even more extreme and moved the trivial cases of non-union tables to a unit test, and replaced `genUnionTableVar` with `genUnionDescendantTableVar` so the QLS test only covers the non-trivial cases of union-derived tables.

test/Test/Database/LSMTree/StateMachine.hs
dcoutts: yes fair enough, done.

test/Test/Database/LSMTree/StateMachine.hs
dcoutts: done

test/Test/Database/LSMTree/StateMachine.hs
dcoutts: BTW, the original code does not do what it intends to do. It intends to have 90% : 10% skew toward union tables, but in fact the result is 40% : 60% skew, i.e. 60% are trivial non-union tables.

I fix this in a later patch (by removing the trivial case entirely), but it could also be fixed by moving the two cases up into the parent `QC.frequency`, each with their corresponding guard. That way we don't get forced into picking the trivial case when there are no non-trivial cases available. It's very common that there are no union tables available.

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -1958,8 +1958,8 @@ arbitraryActionWithVars _ label ctx (ModelState st _stats) =
      ++ [ (2,  fmap Some $ (Action <$> genErrors <*>) $ do
             -- Generate at least a 2-way union, and at most a 3-way union.
             --
-            -- Tests for 0-way and 1-way unions are included in the UnitTests
-            -- module. n-way unions for n>3 lead to large unions, which are less
+            -- Tests for 1-way unions are included in the UnitTests module.
```
dcoutts, line 1963: The point being that there are no 0-way tests (the type forbids it), so it's not actually covered in the unit tests.

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -1759,7 +1759,7 @@ arbitraryActionWithVars _ label ctx (ModelState st _stats) =
       concat
         [ genActionsSession
         , genActionsTables
-        , genUnionActions
+        , genActionsUnion
```
dcoutts, line 1764: just for consistency, while we're at it

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -2248,10 +2248,10 @@ data Stats = Stats {
...
     -- | Number of actions per table (succesful or failing)
   , numActionsPerTable :: !(Map Model.TableID Int)
-    -- | The size of tables that were closed. This is used to augment the table
-    -- sizes from the final model state (which of course has only tables still
-    -- open in the final state).
-  , closedTableSizes   :: !(Map Model.TableID Int)
+    -- | The state of model tables at the point they were closed. This is used
+    -- to augment the tables from the final model state (which of course has
+    -- only tables still open in the final state).
+  , closedTables       :: !(Map Model.TableID Model.SomeTable)
```
dcoutts, line 2244: It turned out that I never needed this change (as I had expected I would) but I've left it here anyway since it's a cheap and sane generalisation, and may indeed be useful later.

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -2261,6 +2261,12 @@ data Stats = Stats {
     -- \"interesting\" actions performed on them. We record only the
     -- interleavings of different tables not multiple actions on the same table.
   , dupTableActionLog  :: Map Model.TableID [Model.TableID]
+    -- | The subset of tables (open or closed) that were created as a result
+    -- of a union operation. This can be used for example to select subsets of
+    -- the other per-table tracking maps above, or the state from the model.
+    -- The map value is the size of the union table at the point it was created,
+    -- so we can distinguish trivial empty unions from non-trivial.
+  , unionTables        :: !(Map Model.TableID Int)
```
dcoutts, line 2259: BTW, the reason we need this is that the model tables do _not_ in fact tell us that they're derived from a union, despite the existence of `isUnionDescendant :: Model.Table k v b -> IsUnionDescendant`. That's because `Model.Table k v b` and `Model.Table k v b` are not the same type. One is `Database.LSMTree.Model.Session as Model` and the other is `Database.LSMTree.Model.Table as Model`. The latter is the one we get from the model state, and does _not_ tell us about unions. The former is really a table handle, while the latter is a table state.

Otherwise I would have used the model final state + the closed tables (generalised in the previous patch) and selected a subset using `isUnionDescendant`.

But as it is, we have to track union tables again in the stats.

test/Test/Database/LSMTree/UnitTests.hs
```diff
@@ -155,18 +150,17 @@ unit_snapshots =
...
     withSession nullTracer hfs hbio (FS.mkFsPath []) $ \sess ->
     withTable @_ @Key1 @Value1 @Blob1 sess defaultTableConfig $ \table -> do
       inserts table [(Key1 17, Value1 42, Nothing)]
 
       bracket (unions $ table :| []) close $ \table' ->
         bracket (duplicate table) close $ \table'' -> do
-          inserts table [(Key1 17, Value1 43, Nothing)]
-          inserts table [(Key1 17, Value1 44, Nothing)]
+          inserts table' [(Key1 17, Value1 43, Nothing)]
+          inserts table'' [(Key1 17, Value1 44, Nothing)]
```
dcoutts, line 165: Oops! The untested unit test was buggy!

test/Database/LSMTree/Model/Session.hs
```diff
@@ -844,7 +844,9 @@ supplyUnionCredits ::
   -> UnionCredits
   -> m UnionCredits
 supplyUnionCredits t@Table{..} c@(UnionCredits credits)
-  | credits <= 0 = pure c
+  | credits <= 0 = do
+      _ <- guardTableIsOpen t
+      pure c
```
dcoutts: It's a bit more consistent/regular to specify it this way. We don't _really_ expect the impl to allow supplying 0 credits to closed tables without an error. All operations on closed tables are an error, trivial or not.

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -1968,42 +1968,36 @@ arbitraryActionWithVars _ label ctx (ModelState st _stats) =
...
-        -- interesting cases to test for are when tables are union tables. For
-        -- non-union tables, these operations should just be no-ops, so we
-        -- generate them only rarely.
-        --
-        -- TODO: tweak distribution once table unions are implemented
-        -- TODO: replace union actions on non-union tables with unit tests?
-        genUnionTableVar = QC.frequency [
-            (9 * length unionDescendantTableVars,    genUnionDescendantTableVar)
-          , (1 * length notUnionDescendantTableVars, genNotUnionDescendantTableVar)
-          ]
```
dcoutts: This nesting of QC.frequency in an outer QC.frequency doesn't do what one might expect. The decisions are dependent rather than independent.

src/Database/LSMTree/Internal.hs
```diff
@@ -1634,4 +1634,9 @@ supplyUnionCredits t credits = do
       RW.withWriteAccess (tableContent tEnv) $ \tableContent ->
         case tableUnionLevel tableContent of
           NoUnion -> pure (tableContent, credits) -- all leftovers
-          Union{} -> error "supplyUnionCredits: not yet implemented"
+          Union{}
+           | credits <= UnionCredits 0 -> pure (tableContent, UnionCredits 0)
+           --TODO: remove this 0 special case once the general case covers it.
+           -- We do not need to optimise the 0 case. It is just here to
+           -- simplify test coverage.
+           | otherwise -> error "supplyUnionCredits: not yet implemented"
```
dcoutts, line 1642: I figured we would more reliably notice this TODO and resolve it when we implement the general case, rather than notice a TODO on an out-of-the-way unit test to say to enable it once `supplyUnionCredits` was implemented. So better to have the unit test running, and temporarily implement the special case in the real impl.

dcoutts (approved):
Obviously I'm also reviewing my own changes here, so I think @mheinzel should review this one too now.

test/Test/Database/LSMTree/UnitTests.hs
dcoutts: Which no doubt git blame will tell me was my fault!

dcoutts:
If @jorisdral and @mheinzel are happy, we can squash the fixes and merge.

test/Test/Database/LSMTree/UnitTests.hs
```diff
@@ -191,6 +192,27 @@ unit_union_credits =
...
+-- | Supplying zero or negative credits to union tables works, but does nothing.
+unit_union_credit_0 :: Assertion
+unit_union_credit_0 =
+    withTempIOHasBlockIO "test" $ \hfs hbio ->
+    withSession nullTracer hfs hbio (FS.mkFsPath []) $ \sess ->
+    withTable @_ @Key1 @Value1 @Blob1 sess defaultTableConfig $ \table -> do
+      inserts table [(Key1 17, Value1 42, Nothing)]
+
+      bracket (union table table) close $ \table' -> do
+        -- Suppling 0 credits works and returns 0 leftovers.
```
jorisdral, line 204: "suppling"

test/Test/Database/LSMTree/UnitTests.hs
```diff
@@ -191,6 +192,27 @@ unit_union_credits =
...
+        UnionCredits leftover <- supplyUnionCredits table' (UnionCredits 0)
+        leftover @?= 0
+
+        -- Suppling negative credits also works and returns 0 leftovers.
```
jorisdral, line 208: "suppling"

test/Test/Database/LSMTree/StateMachine.hs
jorisdral: :+1:  but even in the parent `QC.frequency` you'd get a different distribution than the integers you supply, depending on what actions are available and which guards are true/false. But I agree it's still better

dcoutts: Yes you would, but that should then work as intended.

test/Test/Database/LSMTree/StateMachine.hs
dcoutts: It is a confusing sentence reflecting the confusion I encountered in debugging it :-)

In each individual module the imports are clear, but it goes like this:
```
module Test.Database.LSMTree.StateMachine where
import qualified Database.LSMTree.Model.Session as Model

module Database.LSMTree.Model.Session where
import qualified Database.LSMTree.Model.Table as Model
```
so for code like
```
        , let unionTables =
                Map.filter (Model.withSomeTable (\t -> Model.isUnionDescendant t == Model.IsUnionDescendant))
                           (snd <$> Model.tables finalState)
```
we get
```
    • Couldn't match expected type: Model.Table k40 v0 b0
                  with actual type: Database.LSMTree.Model.Table.Table k v b
      NB: ‘Model.Table’ is defined in ‘Database.LSMTree.Model.Session’
          ‘Database.LSMTree.Model.Table.Table’
            is defined in ‘Database.LSMTree.Model.Table’
    • In the first argument of ‘Model.isUnionDescendant’, namely ‘t’
```
which is sort-of clear actually. GHC does tell us the definition sites.

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -2548,6 +2556,22 @@ updateStats action@(Action _merrs action') lookUp modelBefore _modelAfter result
...
 
+    updUnionTables stats = case action' of
+        Union{}  -> insertUnionTable
+        Unions{} -> insertUnionTable
+        _        -> stats
+      where
+        insertUnionTable
+          | MEither (Right (MTable t)) <- result
+          , let tid = Model.tableID t
+          , Just (_,tbl) <- Map.lookup tid (Model.tables _modelAfter)
```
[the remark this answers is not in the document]
dcoutts: oh yes, oops committed an intermediate state of hacking

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -2253,10 +2254,10 @@ data Stats = Stats {
   , closedTableSizes   :: !(Map Model.TableID Int)
     -- | The ultimate parent for each table. This is the 'TableId' of a table
     -- created using 'new' or 'open'.
-  , parentTable        :: Map Model.TableID Model.TableID
+  , parentTable        :: Map Model.TableID [Model.TableID]
     -- | Track the interleavings of operations via different but related tables.
     -- This is a map from the ultimate parent table to a summary log of which
```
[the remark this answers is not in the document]
dcoutts: applied as "each"

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -2601,6 +2601,8 @@ data Tag =
     -- | A snapshot failed to open because we detected that the snapshot was
     -- corrupt
   | OpenSnapshotDetectsCorruption R.SnapshotName
+    -- | Opened a snapshot for a table involving a union
```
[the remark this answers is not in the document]
dcoutts: done

jorisdral (approved):

PR #621, opened 2025-03-13
doc: add package description, add documented Simple API, and various other minor rewrites

wenkokke:
I'll soon be marking this PR as ready to merge.

I won't make any changes that require changing files in ways that are likely to result in conflict as a part of this PR—it's big enough as it is—so I won't be, e.g., changing the documentation for functions and types re-exported from internal modules, rather than defined in the Simple module. I also won't be removing the Normal and Monoidal modules as a part of this PR, nor will I be rewriting the documentation for the top-level module.

Due to their "soon to be destroyed" status, I could revert 50c3b393, since it adds docs to Normal and Monoidal.


jeltsch:
I read through the Haddock-generated page for `Database.LSMTree.Simple` in the 49277444d1ab8d6d664514bd2f53aba86bca4965 version to familiarize myself with the API. While reading, I already spotted some issues, which I’d like to list here. A thorough review is supposed to come at another time.

Here’s what I found:

* You capitalize all words in titles. Personally, I’m a fan of capitalizing words in titles (except for certain “minor” words, which don’t exist in the titles in this module). However, I think that sections and subsections of Haskell modules are too small to warrant such capitalization. Also, from a quick look at some other packages, I got the impression that capitalizing only the first word of a title is what is done almost universally in Haddock documentation.

* There are several sections with subsections where the titles of the subsections repeat or almost repeat the titles of the respective sections: *Tables* has the subsections *Table lookups*, etc., *Table configuration* has the subsection *Table configuration overrides*, and *Key/value serialisation* has the subsection *Key/value serialisation property tests*. Since the sectioning structure makes it clear what general topics the various subsections refer to, I would just leave out that duplication and name those subsections just *Lookups*, etc., *Overrides*, and *Property tests*.

* There are several occurrences of “key/value”. I think that this should be written “key–value” instead, with the dash being the en dash (U+2013).

* The table in the subsection *Usage notes* » *Resource management* should probably also list `withUnion`/`union` and `withUnions`/`unions` in the `Table` line.

* The subsection *Usage notes* » *Concurrency* has a sentence that begins as follows:

  > It is possible to read obtain read a stable view of a table while concurrently writing to it

  I suggest to change this to the following:

  > It is possible to read from a stable view of a table while concurrently writing to the table

  This would not only fix the wrong phrasing at the beginning but also, by replacing “it” with “the table”, make it clear that it’s the table that is written to, not the stable view of it.

* The last paragraph of the subsection *Usage notes* » *Concurrency* twice contains the phrase “an cursor”, which, of course, should be “a cursor”.

* The documentation of several particular operations repeats information from the *Usage notes* section, for example the information that certain newly created tables must be explicitly closed and that certain operations can be performed concurrently. However, this repeated information doesn’t seem to be comprehensive, and it might be difficult to not only make it but also keep it so. Maybe we should just leave out such repetition. If there is reason to believe that too many users would miss the *Usage notes* section, we could add references to this section to the documentation of the individual operations.

* The documentation refers to runs several times, without explaining what a run is.

* There are two occurrences of “emphemeral”, which should be “ephemeral”.

* The documentation of `lookup` starts with “Look up a the value”, which should be “Look up the value”.

* I don’t understand why `update` and `updates` use values of type `Maybe v`. Wouldn’t it be simpler to use just `v` and in those cases where `Nothing` would have been used just not invoke `update` or invoke `updates` with a thinned-out vector, respectively?

* The documentation of `withUnion` and `union` says that `TableNotCompatibleError` is thrown if both tables aren’t from the same session. What happens if the tables don’t have the same configuration? There should be an error thrown as well, but which one is it?

wenkokke:
> * There are several sections with subsections where the titles of the subsections repeat or almost repeat the titles of the respective sections: _Tables_ has the subsections _Table lookups_, etc., _Table configuration_ has the subsection _Table configuration overrides_, and _Key/value serialisation_ has the subsection _Key/value serialisation property tests_. Since the sectioning structure makes it clear what general topics the various subsections refer to, I would just leave out that duplication and name those subsections just _Lookups_, etc., _Overrides_, and _Property tests_.

This is intentional. It looks verbose and redundant in the table of contents, but when you're browsing the actual documentation it's much more informative to see "Table Configuration Overrides" instead of just "Overrides".

wenkokke:
@jeltsch  Did you have any feedback on the exported API itself?

jeltsch:
> @jeltsch Did you have any feedback on the exported API itself?

So far I have no comments on the API beyond the last two of the above points: the use of `Maybe v` in the types of `update` and `updates` and the unclarity regarding what `withUnion` and `union` do when table configurations don’t match. That said, this was only my first read for familiarizing myself with the API; I’ll have a deeper look at another point. Generally, the interface looks good to me: simple, clean, and intuitive.

wenkokke:
> I don’t understand why update and updates use values of type Maybe v. Wouldn’t it be simpler to use just v and in those cases where Nothing would have been used just not invoke update or invoke updates with a thinned-out vector, respectively?

The purpose is that using Maybe you can batch inserts (Just) and deletions (Nothing).

jeltsch:
> > I don’t understand why `update` and `updates` use values of type `Maybe v`. Wouldn’t it be simpler to use just `v` and in those cases where `Nothing` would have been used just not invoke `update` or invoke `updates` with a thinned-out vector, respectively?
> 
> The purpose is that using `Maybe` you can batch inserts (`Just`) and deletions (`Nothing`).

Oh, I thought that in case of `Nothing` the respective key would just be ignored. The behavior you described makes sense of course. I think the documentation should mention those meanings of `Just` and `Nothing` values.

wenkokke:
> > > I don’t understand why `update` and `updates` use values of type `Maybe v`. Wouldn’t it be simpler to use just `v` and in those cases where `Nothing` would have been used just not invoke `update` or invoke `updates` with a thinned-out vector, respectively?
> 
> > 
> 
> > The purpose is that using `Maybe` you can batch inserts (`Just`) and deletions (`Nothing`).
> 
> 
> 
> Oh, I thought that in case of `Nothing` the respective key would just be ignored. The behavior you described makes sense of course. I think the documentation should mention those meanings of `Just` and `Nothing` values.
> 
> 

Yes, I haven't written the documentation for these functions yet.

recursion-ninja:
> This is intentional. It looks verbose and redundant in the table of contents, but when you're browsing the actual documentation it's much more informative to see "Table Configuration Overrides" instead of just "Overrides".

Maybe the parts which are different could be italicized to stand out more int he table of contents? I am not strongly attached to this suggestion, just an idea.

wenkokke:
> > This is intentional. It looks verbose and redundant in the table of contents, but when you're browsing the actual documentation it's much more informative to see "Table Configuration Overrides" instead of just "Overrides".
> 
> Maybe the parts which are different could be italicized to stand out more int he table of contents? I am not strongly attached to this suggestion, just an idea.

I don't think this would be necessary?

wenkokke:
> - [ ] You capitalize all words in titles. Personally, I’m a fan of capitalizing words in titles (except for certain “minor” words, which don’t exist in the titles in this module). However, I think that sections and subsections of Haskell modules are too small to warrant such capitalization. Also, from a quick look at some other packages, I got the impression that capitalizing only the first word of a title is what is done almost universally in Haddock documentation.
> 
> - [ ] There are several occurrences of “key/value”. I think that this should be written “key–value” instead, with the dash being the en dash (U+2013).
> 
> - [X] The table in the subsection *Usage notes* » *Resource management* should probably also list `withUnion`/`union` and `withUnions`/`unions` in the `Table` line.
> 
> - [ ] The subsection *Usage notes* » *Concurrency* has a sentence that begins as follows:
> 
>   > It is possible to read obtain read a stable view of a table while concurrently writing to it
> 
>   I suggest to change this to the following:
> 
>   > It is possible to read from a stable view of a table while concurrently writing to the table
> 
>     This would not only fix the wrong phrasing at the beginning but also, by replacing “it” with “the table”, make it clear that it’s the table that is written to, not the stable view of it.
> 
> - [ ] The last paragraph of the subsection *Usage notes* » *Concurrency* twice contains the phrase “an cursor”, which, of course, should be “a cursor”.
> 
> - [ ] The documentation of several particular operations repeats information from the *Usage notes* section, for example the information that certain newly created tables must be explicitly closed and that certain operations can be performed concurrently. However, this repeated information doesn’t seem to be comprehensive, and it might be difficult to not only make it but also keep it so. Maybe we should just leave out such repetition. If there is reason to believe that too many users would miss the *Usage notes* section, we could add references to this section to the documentation of the individual operations.
> 
> - [ ] The documentation refers to runs several times, without explaining what a run is.
> 
> - [ ] There are two occurrences of “emphemeral”, which should be “ephemeral”.
> 
> - [ ] The documentation of `lookup` starts with “Look up a the value”, which should be “Look up the value”.
> 
> - [ ] The documentation of `withUnion` and `union` says that `TableNotCompatibleError` is thrown if both tables aren’t from the same session. What happens if the tables don’t have the same configuration? There should be an error thrown as well, but which one is it?

wenkokke:
> > Maybe the parts which are different could be italicized to stand out more in the table of contents? I am not strongly attached to this suggestion, just an idea.
> 
> 
> 
> It would be untypical to do so and may reduce clarity therefore. Therefore, I wouldn’t use italics but either drop the information that is already in the respective section titles or just leave the subsection titles as they are.

I'll leave them as they are.

jeltsch:
Hi, @wen!

I’ve now quite thoroughly reviewed the actual API, in the 1988979dae8af1b2a3e012b518b4d9cbd13d84ec version. I’ve not checked the information about which operations may raise which exceptions: I plan to do this when I thoroughly check the documentation.

Here is my feedback on the API:

* I’m a bit skeptical about the addition of `fromList` and `toList`.

  `fromList` is not just about converting a list into some sort of pure table. Rather it is about creating a new mutable table and just initializing it with the data from the list, which can then be followed by modifications to the table. Is it sensible to bundle the first two steps, creation and initialization, but have the third one, modification, separate? I’d argue that it would be better to just have users create tables with the dedicated operations and then use `inserts` for initializing them. By the way, `fromList` is another operation that requires manual table closing; so its addition carries with it the incentive to add a corresponding bracketing operation, which may in the end lead to overcomplication of the simple(!) interface.

  `toList`, on the other hand, seems more sensible to me, as it is merely about fetching the contents of a table. That said, is this an operation likely to be used? LSM-Tree tables should tend to be big, so that turning their whole contents into ordinary data structures doesn’t seem to be something that is usually done. Also, we already have range lookups.

* What is the `Int` argument of `takeWhile` for? In any case, the existence of this argument makes `takeWhile` for cursors notably different from traditional `takeWhile` operations, which means that it should perhaps be renamed?.

* `openTableFromSnapshotWith` should have its `TableConfigOverride` argument first, like `withTableFromSnapshotWith`.

* Most of the exception types have names that end with `Error` but constructors whose names start with `Err`. I think it would be better to have the constructor names end with `Error` as well.

* Several exception types with single constructors have name parts before `Error` that are different from the parts after `Err` in the names of the corresponding constructors. I think it would be better to name those constructors like their types.

* Is `ByteCountDiscrepancy` an error as well? It doesn’t have `Error` in its name but is an instance of the `Exception` class.

* Also `DeserialiseFailure` doesn’t have `Error` in its name but is an instance of `Exception`. That said, I understand that it is not *only* used as an exception but also as a function result and it comes from `cborg`.

* Why is `SnapshotTableType` listed in the *Errors* section?

* The constructor names of `SnapshotTableType` begin with just `Snap`. I think it would be better for them to begin with `Snapshot` instead.

* I find the repetition of fields for index information in `TableNotCompatibleError` unfortunate. Also, including such information for `union` (as opposed to `unions`) doesn’t make much sense in the first place. I guess it would be good to come up with a nicer structure for this kind of exceptions.

* Unlike the other exception types, `InvalidSnapshotNameError` is not an instance of `Eq`.

wenkokke:
> > The table in the subsection _Usage notes_ » _Resource management_ should probably also list `withUnion`/`union` and `withUnions`/`unions` in the `Table` line.
> 
> 
> 
> I see that you added `withUnion` and `union` but not `withUnions` and `unions`. Was this intentional? One may argue that it’s sort of clear that what applies to the former also applies to the latter, but it might be better to be explicit.

The plural versions are explicitly documented as variants, so I think it's appropriate.

wenkokke:
> * I’m a bit skeptical about the addition of `fromList` and `toList`.
> 
> 
> 
>   `fromList` is not just about converting a list into some sort of pure table. Rather it is about creating a new mutable table and just initializing it with the data from the list, which can then be followed by modifications to the table. Is it sensible to bundle the first two steps, creation and initialization, but have the third one, modification, separate? I’d argue that it would be better to just have users create tables with the dedicated operations and then use `inserts` for initializing them. By the way, `fromList` is another operation that requires manual table closing; so its addition carries with it the incentive to add a corresponding bracketing operation, which may in the end lead to overcomplication of the simple(!) interface.
> 
> 
> 
>   `toList`, on the other hand, seems more sensible to me, as it is merely about fetching the contents of a table. That said, is this an operation likely to be used? LSM-Tree tables should tend to be big, so that turning their whole contents into ordinary data structures doesn’t seem to be something that is usually done. Also, we already have range lookups.

These are less "important additions to the API" and more "stuff I threw in there just now so I could actually write some examples."

wenkokke:
> * What is the `Int` argument of `takeWhile` for? In any case, the existence of this argument makes `takeWhile` for cursors notably different from traditional `takeWhile` operations, which means that it should perhaps be renamed?.

The upper bound on takeWhile is inherited from the internal API, so the question of whether a version of this function without the upper bound makes sense to include is something we'd have to discuss with whomever implemented cursors.

wenkokke:
> Most of the exception types have names that end with Error but constructors whose names start with Err. I think it would be better to have the constructor names end with Error as well.

I agree, though this was simply the naming scheme that all errors in the library used, so changing it would require some degree of discussion.

wenkokke:
> Several exception types with single constructors have name parts before Error that are different from the parts after Err in the names of the corresponding constructors. I think it would be better to name those constructors like their types.

I disagree. The type names denote the class of error, and the constructor names indicate the specific cause. The fact that these differ, eg, in TableCorruptedError, implies that table corruption is currently only detected in lookup operations, but leaves room to have other specific causes in the future.

wenkokke:
> Is ByteCountDiscrepancy an error as well? It doesn’t have Error in its name but is an instance of the Exception class.

This is simply part of the underlying API that I'm reexporting. I'm fixing this in my split error PR.

wenkokke:
> Why is SnapshotTableType listed in the Errors section?
> 
> The constructor names of SnapshotTableType begin with just Snap. I think it would be better for them to begin with Snapshot instead.

This is simply part of the underlying API that I'm reexporting. Joris is about to remove this type.

jeltsch:
> > I’m a bit skeptical about the addition of `fromList` and `toList`.
> >
> > `fromList` is not just about converting a list into some sort of pure table. Rather it is about creating a new mutable table and just initializing it with the data from the list, which can then be followed by modifications to the table. Is it sensible to bundle the first two steps, creation and initialization, but have the third one, modification, separate? I’d argue that it would be better to just have users create tables with the dedicated operations and then use `inserts` for initializing them. By the way, `fromList` is another operation that requires manual table closing; so its addition carries with it the incentive to add a corresponding bracketing operation, which may in the end lead to overcomplication of the simple(!) interface.
> >
> > `toList`, on the other hand, seems more sensible to me, as it is merely about fetching the contents of a table. That said, is this an operation likely to be used? LSM-Tree tables should tend to be big, so that turning their whole contents into ordinary data structures doesn’t seem to be something that is usually done. Also, we already have range lookups.
> 
> These are less “important additions to the API” and more “stuff I threw in there just now so I could actually write some examples”.

If they are less important, then I’d remove them. 🙂 How difficult would it be for your examples to be rephrased in terms of operations like `inserts`? I’d guess their vector-based interfaces should make them quite convenient to use.

jeltsch:
> > What is the `Int` argument of `takeWhile` for? In any case, the existence of this argument makes `takeWhile` for cursors notably different from traditional `takeWhile` operations, which means that it should perhaps be renamed?.
> 
> The upper bound on `takeWhile` is inherited from the internal API, so the question of whether a version of this function without the upper bound makes sense to include is something we’d have to discuss with whomever implemented cursors.

I understand. @jorisdral, @mheinzel, any opinions on that?

jeltsch:
> > Most of the exception types have names that end with `Error` but constructors whose names start with `Err`. I think it would be better to have the constructor names end with `Error` as well.
> 
> I agree, though this was simply the naming scheme that all errors in the library used, so changing it would require some degree of discussion.

@jorisdral, @mheinzel, would you agree to the naming change suggested above?

jeltsch:
> > Several exception types with single constructors have name parts before `Error` that are different from the parts after `Err` in the names of the corresponding constructors. I think it would be better to name those constructors like their types.
> 
> I disagree. The type names denote the class of error, and the constructor names indicate the specific cause. The fact that these differ, eg, in `TableCorruptedError`, implies that table corruption is currently only detected in lookup operations, but leaves room to have other specific causes in the future.

Right, in a case like this such a name “mismatch” is totally understandable, and I agree with you. That said, there are also cases where it’s hard to see how the specific cause of an error could be different from the single one that is currently mentioned: for example, how could a session not exist other than by its directory not existing?

jeltsch:
> > I find the repetition of fields for index information in `TableNotCompatibleError` unfortunate. Also, including such information for `union` (as opposed to `unions`) doesn’t make much sense in the first place. I guess it would be good to come up with a nicer structure for this kind of exceptions.
> 
> This is simply part of the underlying API that I’m reexporting. I don’t know what these indices means, so when I get to clarifying these error types, I’ll require some input from whomever designed these.

My guess is that these indices refer to the list of tables and that a concrete pair of indices specifies an example of two tables that are not compatible. This is also why I thought that these indices don’t make much sense for `union` (as opposed to `unions`), as with `union` there are always only two tables.

wenkokke:
> > > Most of the exception types have names that end with `Error` but constructors whose names start with `Err`. I think it would be better to have the constructor names end with `Error` as well.
> > 
> > 
> > I agree, though this was simply the naming scheme that all errors in the library used, so changing it would require some degree of discussion.
> 
> @jorisdral, @mheinzel, would you agree to the naming change suggested above?

For the record, I oppose this naming change, as I believe the type and constructor should have different names.

PR #623, opened 2025-03-14
Run `Class`-based and `StateMachine` tests using both index types

jorisdral:
Enabling both types of tests for both types of indexes unearthed some `NoThunks` failures and a subtle assertion failure. Both are fixed now, though the assertion failure currently has a workaround instead of a proper fix.


dcoutts (approved):
LGTM!

PR #624, opened 2025-03-14
fix: LSMTreeError docs

wenkokke:
The previous docs were misaligned with their respective constructors.

Screenshot 2025-03-14 at 16 17 08

### DIFF
--- a/src/Database/LSMTree/Internal.hs
+++ b/src/Database/LSMTree/Internal.hs
@@ -191,23 +191,24 @@ instance NFData (MonoidalCursor m k v) where
 
 -- TODO: give this a nicer Show instance.
 data LSMTreeError =
+    -- | The session directory does not exist.
     SessionDirDoesNotExist FsErrorPath
-    -- | The session directory is already locked
-  | SessionDirLocked FsErrorPath
-    -- | The session directory is malformed: the layout of the session directory
+  | -- | The session directory is already locked
+    SessionDirLocked FsErrorPath
+  | -- | The session directory is malformed: the layout of the session directory
     -- contains unexpected files and/or directories.
-  | SessionDirMalformed FsErrorPath
-    -- | All operations on a closed session as well as tables or cursors within
+    SessionDirMalformed FsErrorPath
+  | -- | All operations on a closed session as well as tables or cursors within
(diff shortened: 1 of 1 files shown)


jorisdral (commented):
Nice find!

> The previous docs were misaligned with their respective constructors.

Huh, this almost looks like a `haddock` bug to me, specially since the [haddock documentation](https://haskell-haddock.readthedocs.io/latest/markup.html#constructors-and-record-fields) prescribes the approach that we are taking on `main` currently. 

I'm not opposed this change but a minor nit is that my `vscode`'s comment rewrapping command does not work on
```haskell
data Datatype =
  | -- | comment
    Constructor
```
while it does work on
```haskell
data Datatype =
    -- | comment
  | Constructor 
```

FYI, putting a newline between constructors solves the problem as well, or alternatively we can write the documentation like this:
```haskell
data Datatype =
    Constructor
      -- ^ constructor comment
      T1 -- ^ first argument
      T2 -- ^ second argument
```

Thoughts?

wenkokke:
> Nice find!
> 
> > The previous docs were misaligned with their respective constructors.
> 
> Huh, this almost looks like a `haddock` bug to me, specially since the [haddock documentation](https://haskell-haddock.readthedocs.io/latest/markup.html#constructors-and-record-fields) prescribes the approach that we are taking on `main` currently.
> 
> I'm not opposed this change but a minor nit is that my `vscode`'s comment rewrapping command does not work on
> 
> ```haskell
> data Datatype =
>   | -- | comment
>     Constructor
> ```
> 
> while it does work on
> 
> ```haskell
> data Datatype =
>     -- | comment
>   | Constructor 
> ```
> 
> FYI, putting a newline between constructors solves the problem as well, or alternatively we can write the documentation like this:
> 
> ```haskell
> data Datatype =
>     Constructor
>       -- ^ constructor comment
>       T1 -- ^ first argument
>       T2 -- ^ second argument
> ```
> 
> Thoughts?

I do prefer the prefix style (and unrelatedly prefer no comment rewrapping, which I think is a bad idea in general since it creates bad diffs).

jorisdral (approved):

PR #625, opened 2025-03-14
feat: export doesSnapshotExist

wenkokke:
This PR adds `doesSnapshotExist` so the user doesn't have to do ``(snap `elem`) <$> listSnapshots session``.

### DIFF
--- a/src/Database/LSMTree/Internal.hs
+++ b/src/Database/LSMTree/Internal.hs
@@ -63,6 +63,7 @@ module Database.LSMTree.Internal (
   , createSnapshot
   , openSnapshot
   , deleteSnapshot
+  , doesSnapshotExist
   , listSnapshots
     -- * Multiple writable tables
   , duplicate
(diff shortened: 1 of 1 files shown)


src/Database/LSMTree/Internal.hs
```diff
@@ -63,6 +63,7 @@ module Database.LSMTree.Internal (
   , createSnapshot
   , openSnapshot
   , deleteSnapshot
+  , doesSnapshotExist
```
jorisdral, line 66: Are you intentionally not adding it to the public API yet?

src/Database/LSMTree/Internal.hs
```diff
@@ -1209,9 +1210,8 @@ createSnapshot snap label tableType t = do
 
         -- Guard that the snapshot does not exist already
         let snapDir = Paths.namedSnapshotDir (tableSessionRoot tEnv) snap
-        doesSnapshotExist <-
-          FS.doesDirectoryExist (tableHasFS tEnv) (Paths.getNamedSnapshotDir snapDir)
-        if doesSnapshotExist then
+        snapshotExists <- doesSnapshotExist (tableSession t) snap
```
jorisdral, line 1213: I like the reuse, but it's not great that it takes a second read lock on the session, though it does not matter much for performance or correctness reasons. It's just a minor nit that we don't need to take the lock twice. We can extract a function on `SessionEnv` (or just the components that it needs) that we then reuse instead

src/Database/LSMTree/Internal.hs
```diff
@@ -1342,6 +1342,21 @@ openSnapshot sesh label tableType override snap resolve = do
           , tableUnionLevel = unionLevel
           }
 
+{-# SPECIALISE doesSnapshotExist ::
+     Session IO h
+  -> SnapshotName
+  -> IO Bool #-}
+-- |  See 'Database.LSMTree.Common.doesSnapshotExist'.
+doesSnapshotExist ::
```
jorisdral, line 1350: :+1:

jorisdral (approved):

PR #619, opened 2025-03-12
Make ordinary indexes the default

jeltsch:
Up to now, `defaultTableConfig` declared compact indexes to be used by default. This pull request changes this, making ordinary indexes the default.

Furthermore, this pull request changes the following files such that they now override the `confFencePointerIndex` field of `defaultTableConfig` with `CompactIndex` in order to continue using compact indexes:

* `bench/macro/lsm-tree-bench-wp8.hs`
* `bench/micro/Bench/Database/LSMTree/Monoidal.hs`
* `bench/micro/Bench/Database/LSMTree/Normal.hs`
* `test/Test/Database/LSMTree/Internal/Snapshot/Codec/Golden.hs`
* `test/Test/Database/LSMTree/Internal/Snapshot/FS.hs`
* `test/Test/Database/LSMTree/UnitTests.hs`

The remaining source code files that use `defaultTableConfig` are the following ones:

* `test/Test/Database/LSMTree/Class.hs`
* `test/Test/Database/LSMTree/Internal.hs`

These are not changed, so that they now operate with ordinary indexes.

PR #628, opened 2025-03-15
Finish switching of index-using tests to ordinary indexes

jeltsch:
This pull request changes the tests of various index-using modules such that these tests now use ordinary indexes instead of compact ones. #619 already made such changes to certain test modules that refer to the default table configuration. The present pull request changes test modules that specify the index type in isolation, that is, not as part of a table configuration. It modifies all such modules except for the snapshot-related ones and the main one of the state-machine-related ones.


jorisdral (approved):
LGTM. Let's merge once #638 is done

jorisdral (approved):

PR #629, opened 2025-03-17
Use `ghc-9.12.2`

jorisdral:


.github/workflows/haskell.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["8.10.7", "9.2.8", "9.4.8", "9.6.4", "9.8.2", "9.10.1", "9.12.1"]
+        ghc: ["8.10.7", "9.2.8", "9.4.8", "9.6.4", "9.8.2", "9.10.1", "9.12.12]
```
wenkokke, line 24: That says "9.12.12"

wenkokke (changes requested):

.github/workflows/haskell.yml
jorisdral: Whoops. Fixed it!

dcoutts (approved):

wenkokke:
@jorisdral We need to update the branch protection rules to no longer expect the tests with GHC 9.12.1, merge, and then add the tests with GHC 9.12.2.

PR #631, opened 2025-03-18
Update bloom filter FPR tests

jorisdral:
These tests were missing a link with the real implementation, since we were testing bloom filter construction functions that were not used in the real implementation. The updated tests test the `RunAcc`'s method of bloom filter allocation directly.


dcoutts (approved):
LGTM!

PR #632, opened 2025-03-18
fix: replace mkSnapshotName that returns Maybe with toSnapshotName that errors

wenkokke:
Perhaps tellingly, every single use save one used either `fromJust` or `maybe (error ...)`. The only use that handled the error simply didn't run any tests if the function returned maybe.

### DIFF
--- a/src/Database/LSMTree/Internal.hs
+++ b/src/Database/LSMTree/Internal.hs
@@ -1394,15 +1394,13 @@ listSnapshots sesh = do
       snaps <- mapM (checkSnapshot hfs root) $ Set.toList contents
       pure $ catMaybes snaps
   where
-    checkSnapshot hfs root s =
-      case Paths.mkSnapshotName s of
-        Nothing   -> pure Nothing
-        Just snap -> do
-          -- check that it is a directory
-          b <- FS.doesDirectoryExist hfs
-                (Paths.getNamedSnapshotDir $ Paths.namedSnapshotDir root snap)
-          if b then pure $ Just snap
-               else pure $ Nothing
+    checkSnapshot hfs root s = do
+      let snap = Paths.toSnapshotName s
+      -- check that it is a directory
+      b <- FS.doesDirectoryExist hfs
+            (Paths.getNamedSnapshotDir $ Paths.namedSnapshotDir root snap)
+      if b then pure $ Just snap
+            else pure $ Nothing
 
 {-------------------------------------------------------------------------------
   Multiple writable tables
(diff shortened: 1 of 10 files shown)


src/Database/LSMTree/Internal.hs
```diff
@@ -1394,15 +1394,13 @@ listSnapshots sesh = do
...
-                (Paths.getNamedSnapshotDir $ Paths.namedSnapshotDir root snap)
-          if b then pure $ Just snap
-               else pure $ Nothing
+    checkSnapshot hfs root s = do
+      let snap = Paths.toSnapshotName s
+      -- check that it is a directory
+      b <- FS.doesDirectoryExist hfs
+            (Paths.getNamedSnapshotDir $ Paths.namedSnapshotDir root snap)
+      if b then pure $ Just snap
+            else pure $ Nothing
```
jorisdral, line 1403: This is not doing the same thing as before, is that okay? Should `listSnapshots` throw an error if there is a file/directory that violates `isValidSnapshotname`, or should it ignore the file/directory?

src/Database/LSMTree/Internal/Paths.hs
```diff
@@ -111,47 +112,90 @@ snapshotMetaDataChecksumFile (NamedSnapshotDir dir) =
...
 instance IsString SnapshotName where
-  fromString s = fromMaybe bad (mkSnapshotName s)
-    where
-      bad = error ("SnapshotName.fromString: invalid name " ++ show s)
+  fromString :: String -> SnapshotName
+  fromString = toSnapshotName
 
--- | Create snapshot name.
+data InvalidSnapshotNameError
+  = ErrSnapshotNameInvalid !String
```
jorisdral, line 128: This should be exposed from the public API

src/Database/LSMTree/Internal/Paths.hs
```diff
@@ -111,47 +112,90 @@ snapshotMetaDataChecksumFile (NamedSnapshotDir dir) =
...
-  , len < 65
-  , System.FilePath.Posix.isValid s
-  , System.FilePath.Windows.isValid s
-  = Just (MkSnapshotName s)
-
-  | otherwise
-  = Nothing
+-- See, e.g., [the VBA docs for the "Bad file name or number" error](https://learn.microsoft.com/en-us/office/vba/language/reference/user-interface-help/bad-file-name-or-number-error-52).
+isValidSnapshotName :: String -> Bool
+isValidSnapshotName str =
```
jorisdral, line 173: This should be exposed from the public API

PR #637, opened 2025-03-18
refactor: organize error types

wenkokke:


src/Database/LSMTree/Internal/Errors.hs
jorisdral, line 1: I think I'd prefer aggregating the errors in a module/re-export them in the public API. The current `Errors` module enforces structure because of import cycles that I don't think is great, like having a `Snapshot.Core` module, or having exceptions away from modules where they are used

test/Database/LSMTree/Model/Session.hs
```diff
@@ -246,59 +250,60 @@ runModelMWithInjectedErrors ::
...
 -- Errors
 --
 
-data Err =
-    ErrTableClosed
-  | ErrSnapshotExists
-  | ErrSnapshotDoesNotExist
-  | ErrSnapshotWrongType
-  | ErrBlobRefInvalidated
+data Err
```
jorisdral, line 263: I wonder if we can leave out the errors that the model never throws, like `ErrTableTypeMismatch`. If we don't model certain behaviour, then maybe we don't need to represent the errors. Then, SUT errors that have no model equivalent can be mapped using `ErrOther . show`

jorisdral (commented):
Not a full review yet, but here are some initial thoughts:

test/Database/LSMTree/Model/Session.hs
wenkokke: I'm taking `ErrOther` to only be used for errors not thrown by the library API, so that it's easier to see if something unusual is going on.

src/Database/LSMTree/Internal/Errors.hs
wenkokke: That makes sense.

wenkokke:
I've moved all the error types to be defined next to the function that throws them and I've unmade the `Snapshot.Core` module.

test/Database/LSMTree/Model/Session.hs
wenkokke: Let's leave `ErrTableTypeMismatch` for a future PR to sort out, because it looks like the _real_ implementation doesn't throw it either, so that might require some input from the folks working on unions.

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -465,97 +469,98 @@ createSystemTempDirectory prefix = do
...
-fileFormatErrorHandler :: Monad m => Handler m (Maybe Model.Err)
-fileFormatErrorHandler = Handler $ pure . handler'
-  where
-    handler' :: FileFormatError -> Maybe Model.Err
-    handler' e = Just (Model.ErrSnapshotCorrupted (displayException e))
+handleAbortActionRegistryError :: AbortActionRegistryError -> Model.Err
+handleAbortActionRegistryError = \case
+  AbortActionRegistryError abortReason actionErrors ->
+    Model.ErrAbortActionRegistry $
+      handleSomeException <$> maybe id NE.cons (getReasonExitCaseException abortReason) (getActionError <$> actionErrors)
```
jorisdral, line 506: I think `ErrAbortActionRegistry` should keep the exit case reason separate. This is going to be a type that users can see in counter examples, so it's useful to see the distinction  between the exit case reason and any exceptions that happen while commiting / aborting the action registry

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -465,97 +469,98 @@ createSystemTempDirectory prefix = do
...
+handleBlobRefInvalidError :: BlobRefInvalidError -> Model.Err
+handleBlobRefInvalidError = \case
+  ErrBlobRefInvalid _ -> Model.ErrBlobRefInvalid
+
+handleCursorClosedError :: CursorClosedError -> Model.Err
+handleCursorClosedError = \case
+  ErrCursorClosed -> Model.ErrCursorClosed
+
+handleFsError :: FsError -> Model.Err
+handleFsError = Model.ErrDiskFault . displayException
```
jorisdral, line 563: `ErrDiskFault` was originally representing a supserset of errors. Maybe we can map `FsError` to `ErrFsError`, and then reserve `ErrDiskFault` for the catch-all "something went wrong" error that the model throws

test/Database/LSMTree/Model/Session.hs
```diff
@@ -246,59 +250,60 @@ runModelMWithInjectedErrors ::
...
+
+isOther :: Err -> Bool
+isOther (ErrOther _)                 = True
+isOther (ErrCommitActionRegistry es) = all isOther es
+isOther (ErrAbortActionRegistry es)  = all isOther es
+isOther _                            = False
+
+firstNotOtherSatisfies :: (Err -> Bool) -> NonEmpty Err -> Bool
+firstNotOtherSatisfies p =
+  maybe False (p . fst) . L.uncons . NE.dropWhile isOther
```
jorisdral, line 306: I wonder if we shouldn't ignore `ErrOther`. The way I see it, `ErrOther` is always an unexpected exception, so maybe it should always lead to a test failure, which would be contingent on not "swallowing" the `ErrOther` in `isDiskFault` and `isSnapshotCorrupted`

src/Database/LSMTree/Internal.hs
```diff
@@ -1344,6 +1394,57 @@ openSnapshot sesh label tableType override snap resolve = do
...
+        wrapSomeException :: SomeException -> SomeException
+        wrapSomeException e =
+            fromMaybe e . getFirst . mconcat . fmap First $
+                [ toException . wrapFileCorruptedError <$> fromException e
+                , toException . wrapAbortActionRegistryError <$> fromException e
+                , toException . wrapCommitActionRegistryError <$> fromException e
+                ]
+
+        wrapFileCorruptedError :: FileCorruptedError -> SnapshotCorruptedError
+        wrapFileCorruptedError = ErrSnapshotCorrupted snapshotName
```
jorisdral, line 1431: So the user would see an `AbortActionRegistryError` or `CommitActionRegistryError` that contains an `ErrSnapshotCorrupted`?

I think the nicer thing would be if we flip this: the user gets `ErrSnapshotCorrupted` containing a `FileCorruptedError`/`AbortActionRegistryError`/`CommitActionRegistryError`

test/Database/LSMTree/Model/Session.hs
jorisdral: Similarly, I think `isOther` should hold when `isOther` holds for *any* of exceptions recursively

src/Database/LSMTree/Internal/Paths.hs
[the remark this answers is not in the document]
jorisdral: It's not something we have to address in this PR necessarily, I just thought I'd include it here for future PRs

src/Database/LSMTree/Internal.hs
```diff
@@ -949,6 +938,21 @@ updates resolve es t = do
...
 
+{- | A 'BlobRef' used with 'retrieveBlobs' was invalid.
+
+'BlobRef's are obtained from lookups in a 'Table', but they may be
+invalidated by subsequent changes in that 'Table'. In general the
+reliable way to retrieve blobs is not to change the 'Table' before
+retrieving the blobs. To allow later retrievals, duplicate the table
+before making modifications and keep the table open until all blob
+retrievals are complete.
+-}
```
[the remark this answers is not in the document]
wenkokke: I'd prefer not to. The main use of block comments is that there's no admin around line breaks. Indenting would make them as difficult to maintain as line comments.

src/Database/LSMTree/Internal.hs
```diff
@@ -1437,6 +1538,25 @@ duplicate t@Table{..} = do
...
+-- | An operation was called with two tables that are not compatible.
+data TableNotCompatibleError
+    = -- | An operation was called with two tables that are not of the same type.
+      --
+      --   TODO: This error is no longer used by 'unions'.
+      ErrTableTypeMismatch
+        -- | Vector index of table @t1@ involved in the mismatch
+        Int
+        -- | Vector index of table @t2@ involved in the mismatch
+        Int
```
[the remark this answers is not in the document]
wenkokke: It's unused, but should be used.

src/Database/LSMTree/Internal.hs
```diff
@@ -1437,6 +1538,25 @@ duplicate t@Table{..} = do
    Table union
 -------------------------------------------------------------------------------}
 
+-- | An operation was called with two tables that are not compatible.
+data TableNotCompatibleError
```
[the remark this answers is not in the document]
wenkokke: But it won't necessarily always only come up in unions. It comes up for all operations with two or more arguments.

The fact that it's only unions will only be true for as long as unions are our only operations with two or more arguments.

src/Database/LSMTree/Internal/Paths.hs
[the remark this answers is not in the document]
wenkokke: This is an artifact of you asking me to put snapshot names back there.

src/Database/LSMTree/Internal/Snapshot/Codec.hs
```diff
@@ -106,48 +106,41 @@ writeFileSnapshotMetaData hfs contentPath checksumPath snapMetaData = do
...
     -> FsPath
     -> FsPath
-    -> IO (Either DeserialiseFailure SnapshotMetaData)
+    -> IO SnapshotMetaData
   #-}
 -- | Read from 'SnapshotMetaDataFile' and attempt to decode it to
 -- 'SnapshotMetaData'.
 readFileSnapshotMetaData ::
-     MonadThrow m
+     (MonadThrow m)
```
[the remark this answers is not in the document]
wenkokke: Yes.

test/Test/Database/LSMTree/StateMachine.hs
wenkokke: So map FsError to ErrFsError and then match ErrFsError and the model's ErrDiakFault?

test/Database/LSMTree/Model/Session.hs
wenkokke: I could make that change but I need the current behavior, so it'd be kinda pointless?

src/Database/LSMTree/Internal.hs
wenkokke: The point is only to wrap the file corruption error in a snapshot corruption error, not to restructure anything.

src/Database/LSMTree/Internal.hs
```diff
@@ -949,6 +938,21 @@ updates resolve es t = do
...
+'BlobRef's are obtained from lookups in a 'Table', but they may be
+invalidated by subsequent changes in that 'Table'. In general the
+reliable way to retrieve blobs is not to change the 'Table' before
+retrieving the blobs. To allow later retrievals, duplicate the table
+before making modifications and keep the table open until all blob
+retrievals are complete.
+-}
+data BlobRefInvalidError
+    = -- | The 'Int' index indicates the first invalid 'BlobRef'.
+      ErrBlobRefInvalid !Int
```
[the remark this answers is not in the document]
wenkokke: This is a comment I'm preserving, rather than something I wrote. I'll fix the documentation in a separate PR.

src/Database/LSMTree/Internal.hs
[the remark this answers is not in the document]
jorisdral: Hmm could it be because of https://github.com/IntersectMBO/lsm-tree/pull/637#discussion_r2011005471 that they are unused?

src/Database/LSMTree/Internal/Paths.hs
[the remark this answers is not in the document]
jorisdral: Ah, my bad

test/Database/LSMTree/Model/Session.hs
jorisdral: If I change it to:

```haskell
isSnapshotCorrupted :: Err -> Bool
isSnapshotCorrupted (ErrDiskFault _)             = True
isSnapshotCorrupted (ErrSnapshotCorrupted _)     = True
isSnapshotCorrupted e@(ErrCommitActionRegistry es) = not (isOther e) && (isSnapshotCorrupted $ fst $ NE.uncons es)
isSnapshotCorrupted e@(ErrAbortActionRegistry es)  = not (isOther e) && (isSnapshotCorrupted $ fst $ NE.uncons es)
isSnapshotCorrupted _                            = False

isDiskFault :: Err -> Bool
isDiskFault (ErrDiskFault _)             = True
isDiskFault e@(ErrCommitActionRegistry es) = not (isOther e) && (isDiskFault $ fst $ NE.uncons es)
isDiskFault e@(ErrAbortActionRegistry es)  = not (isOther e) && (isDiskFault $ fst $ NE.uncons es)
isDiskFault _                            = False

isOther :: Err -> Bool
isOther (ErrOther _)                 = True
isOther (ErrCommitActionRegistry es) = any isOther es
isOther (ErrAbortActionRegistry es)  = any isOther es
isOther _                            = False
```

then the tests pass for me. Do you need the previous behaviour for something else?

src/Database/LSMTree/Internal.hs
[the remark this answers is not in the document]
jorisdral: Are we planning on adding other operations?

src/Database/LSMTree/Internal/CRC32C.hs
```diff
@@ -412,21 +420,51 @@ expectChecksum ::
...
 -------------------------------------------------------------------------------}
 
-data FileFormatError = FileFormatError FsPath String
-  deriving stock Show
-  deriving anyclass Exception
+data FileFormat
+    = FormatChecksumsFile
+    | FormatBloomFilterFile
+    | FormatIndexFile
+    | FormatRunFile
```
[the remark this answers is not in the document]
wenkokke: Which one?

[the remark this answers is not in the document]
jorisdral: `FormatRunFile`

test/Test/Database/LSMTree/StateMachine.hs
```diff
@@ -1035,18 +1040,17 @@ instance Eq (Obs h a) where
       -- 'runRealWithInjectedErrors'.
       (OEither (Left (OId lhs)), OEither (Left (OId rhs)))
         | Just (e :: Model.Err) <- cast lhs
-        , case e of
-            Model.ErrOther _ -> False
-            _                -> True
-        , Just Model.DefaultErrDiskFault <- cast rhs
+        , not $ Model.isOther e
+        , Just (Model.ErrDiskFault _) <- cast rhs
```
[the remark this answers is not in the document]
wenkokke: This... code isn't there anymore?

test/Database/LSMTree/Model/Session.hs
```diff
@@ -280,33 +278,22 @@ data Err
   | ErrFsError !String
   | ErrCommitActionRegistry !(NonEmpty Err)
   | ErrAbortActionRegistry !(Maybe Err) !(NonEmpty Err)
-  | ErrOther !String
```
jorisdral: Any uncaught exceptions in the state machine tests will make the tests fail without printing counterexamples. I haven't really looked into why that is the case, but to combat it I added `ErrOther` as a catch-all for all unexpected exceptions. Then the tests will fail because of a pure inequality (like `ErrOther _ /= _`), which does print the counterexamples nicely.

So I think it would be nice to keep `ErrOther`, and probably we should add a comment to `ErrOther` to explain that it exists for nicer counterexamples

And I liked that previously we could check that exceptions did not recursively contain exceptions that we did not expect, using `isOther`. But I guess now those exceptions would not be mapped to a model error?

So I suspect you removed it because then the `isOther` handling is kind of automatic. Is that right?

test/Test/Database/LSMTree/StateMachine.hs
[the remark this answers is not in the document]
jorisdral: Hmm, weird, it's here for me

```suggestion
        , Just Model.ModelErrDiskFault <- cast rhs
```

src/Database/LSMTree/Internal.hs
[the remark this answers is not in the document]
jorisdral: ^ bumping my question

test/Database/LSMTree/Model/Session.hs
jorisdral: I've suggested in an earlier comment to bring it back

src/Database/LSMTree/Internal.hs
jorisdral: I suppose I'm suggesting to restructure it. Would it be something you think makes sense to do? Currently, I wouldn't expect users to have to catch `ActionRegistry` exceptions, but maybe you had a design in mind?

test/Database/LSMTree/Model/Session.hs
wenkokke: Sorta, but the uncaught exceptions point is valid. However, should we fix that with a model error, or by catching all exceptions at the top and printing them?

wenkokke:
> I suppose I'm suggesting to restructure it. Would it be something you think makes sense to do? Currently, I wouldn't expect users to have to catch ActionRegistry exceptions, but maybe you had a design in mind?

The current PR preserves existing behaviour. If restructuring is something you want, I'd suggest that should be a follow-up PR.

wenkokke:
> @jorisdral [wrote](https://github.com/IntersectMBO/lsm-tree/pull/637#discussion_r2013880491): bumping my question

I can no longer see the question? GitHub is having a morning.

wenkokke:
@jorisdral Do you want me to revert 17e8de75f5e2cf2c55a95b639b7f2efaaae9219d?

test/Test/Database/LSMTree/StateMachine.hs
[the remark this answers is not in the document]
wenkokke: Screenshot 2025-03-26 at 11 42 02

src/Database/LSMTree/Internal.hs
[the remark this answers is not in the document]
wenkokke: I don't know, but it would be awkward to restructure the whole error mechanism if we do? I'll update the constructors to reflect the fact that they originate from unions, though.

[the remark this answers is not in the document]
wenkokke: We can always have the Haddock clarify this.

test/Database/LSMTree/Model/Session.hs
wenkokke: Do you want me to revert 17e8de75f5e2cf2c55a95b639b7f2efaaae9219d?

src/Database/LSMTree/Internal.hs
wenkokke: I suggest that this change should be considered out-of-scope for this PR, and be the subject of a future PR.

It's entirely non-obvious to me how to decide which of the various exceptions under an ActionRegistry exception should take precedence, e.g., what do you do if there's two `FileCorruptedError`s?

test/Test/Database/LSMTree/StateMachine.hs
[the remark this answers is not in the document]
wenkokke: Either way matching on `ErrDiskFault` is fine, since there's no reason to verify via pattern match that the model error string contains the only string that the model ever puts there.

src/Database/LSMTree/Internal.hs
jorisdral: Okay, let's do it in a follow-up PR

> It's entirely non-obvious to me how to decide which of the various exceptions under an ActionRegistry exception should take precedence, e.g., what do you do if there's two `FileCorruptedError`s?

Probably in chronological order

test/Database/LSMTree/Model/Session.hs
wenkokke: > If I change it to:
> 
> ```haskell
> isSnapshotCorrupted :: Err -> Bool
> isSnapshotCorrupted (ErrDiskFault _)             = True
> isSnapshotCorrupted (ErrSnapshotCorrupted _)     = True
> isSnapshotCorrupted e@(ErrCommitActionRegistry es) = not (isOther e) && (isSnapshotCorrupted $ fst $ NE.uncons es)
> isSnapshotCorrupted e@(ErrAbortActionRegistry es)  = not (isOther e) && (isSnapshotCorrupted $ fst $ NE.uncons es)
> isSnapshotCorrupted _                            = False
> 
> isDiskFault :: Err -> Bool
> isDiskFault (ErrDiskFault _)             = True
> isDiskFault e@(ErrCommitActionRegistry es) = not (isOther e) && (isDiskFault $ fst $ NE.uncons es)
> isDiskFault e@(ErrAbortActionRegistry es)  = not (isOther e) && (isDiskFault $ fst $ NE.uncons es)
> isDiskFault _                            = False
> 
> isOther :: Err -> Bool
> isOther (ErrOther _)                 = True
> isOther (ErrCommitActionRegistry es) = any isOther es
> isOther (ErrAbortActionRegistry es)  = any isOther es
> isOther _                            = False
> ```
> 
> then the tests pass for me. Do you need the previous behaviour for something else?

This has rather different semantics. The point is to find the first cause that isn't an `ErrOther` and classify that. This handles the scenario where the causes contain both corruption and disk faults.

The semantics of `isOther` are also incorrect here. `ErrOther` is the bottom element of a lattice, so any non-other error takes precedence, which explains the need for `all`. Corruption and disk fault on the other hand take precedence (in order) over any other error.

test/Test/Database/LSMTree/StateMachine.hs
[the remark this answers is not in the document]
jorisdral: It doesn't hurt to make the pattern match more precise

test/Database/LSMTree/Model/Session.hs
jorisdral: Probably, yes

wenkokke: > Probably, yes

W.r.t. reverting or w.r.t. agreeing that these predicates should indeed use the causes of action registry errors differently?

src/Database/LSMTree/Internal.hs
wenkokke: Makes sense, but might be better to present the user with the whole picture rather than selectively hide stuff and possibly get that wrong.

test/Database/LSMTree/Model/Session.hs
jorisdral: Reverting

src/Database/LSMTree/Internal.hs
wenkokke: Might be a good change, but let's push that for a future PR.

wenkokke: See #646 for follow-up discussion.

PR #639, opened 2025-03-20
Remove some obsolete TODOs and unused declarations

jorisdral:
I perused comments about TODOs and found some things that we can probably remove or simplify


dcoutts (approved):
LGTM

PR #648, opened 2025-03-27
Extract a `blockio` package from `lsm-tree`

jorisdral:
Both the `HasBlockIO` API and simulation will have to be exposed in some manner to users of the `lsm-tree` library. The former so that users can open a session, and the latter to make sure that users can also run `lsm-tree` with a file system simulation. Recall `openSession`:

https://github.com/IntersectMBO/lsm-tree/blob/5e72f1f2b93ecbbcb44b8d0f807a1cb4b35a29aa/src/Database/LSMTree/Common.hs#L182-L189

^ the answer to the TODO here is: I don't think that in general we can prevent the user from having to pass in `HasBlockIO`. Initialising a `HasFS` is linked to initialising a `HasBlockIO`, but the way in which we do this for the real file system and simulated file system is different:

https://github.com/IntersectMBO/lsm-tree/blob/5e72f1f2b93ecbbcb44b8d0f807a1cb4b35a29aa/blockio-api/src/System/FS/BlockIO/IO.hs#L12-L17

https://github.com/IntersectMBO/lsm-tree/blob/5e72f1f2b93ecbbcb44b8d0f807a1cb4b35a29aa/blockio-sim/src/System/FS/BlockIO/Sim.hs#L148-L155

So we can't hide `HasBlockIO` from the user, it seems. And, as @wenkokke suggested, we should probably expose the API and simulation in a separate package.

I had initially hoped we could re-export the necessary `blockio` definitions from the main `lsm-tree` library while keeping the sub-libraries hidden. However, since we shouldn't export the simulation (with dependencies on test libraries) from the public API, we would have to create a public sub-library with just the blockio simulation. At this point, it's probably better to have a separate package.

BWT, I also made `blockio-sim` a public sub-library of `blockio-api` (i.e., now the component is identified by `blockio:sim`). The reasoning here is that there are very few definitions in `blockio-sim`, and one would have to depend on `blockio-api` anyway to do anything with the simulation. As such, you could argue it's better not to separately publish `blockio-sim`. As far as I understand, `Hackage` has also gotten better at support for sub-libraries, see https://hackage.haskell.org/package/io-classes for example.


blockio/README.md
jorisdral, line 1: I should put some text here if we want to proceed with this PR

wenkokke (approved):
LGTM!

That README does probably need some content, but I'm happy to merge this with is containing no more than the package name.

PR #654, opened 2025-03-27
doc: add package description

wenkokke:
For the package description to make sense, this depends on #653, which is still marked WIP due to a few small pending fixes to the error types. Furthermore, this PR leaves a few TODOs that do not have clear answers yet. However, I believe it can be merged in its current state.

### DIFF
--- a/lsm-tree.cabal
+++ b/lsm-tree.cabal
@@ -1,8 +1,117 @@
 cabal-version:      3.4
 name:               lsm-tree
 version:            0.1.0.0
-synopsis:           Log-structured merge-tree
-description:        Log-structured merge-tree.
+synopsis:           Log-structured merge-trees
+description:
+  This package contains an efficient implementation of on-disk key–value storage, implemented as a log-structured merge-tree or LSM-tree.
+  An LSM-tree is a data structure for key–value mappings, similar to "Data.Map", but optimized for large tables with a high insertion volume.
+  It has support for:
+
+  *   Basic key–value operations, such as lookup, insert, and delete.
+  *   Range lookups, which efficiently retrieve the values for all keys in a given range.
+  *   Monoidal upserts (or \"mupserts\") which combine the stored and new values.
+  *   BLOB storage which assocates a large auxiliary BLOB with a key.
+  *   Durable on-disk persistence and rollback via named snapshots.
+  *   Cheap table duplication where all duplicates can be independently accessed and modified.
+  *   High-performance lookups on SSDs using I\/O batching and parallelism.
+
+  This package exports two modules:
+
+  *   "Database.LSMTree.Simple"
+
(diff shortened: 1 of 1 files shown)


lsm-tree.cabal
```diff
@@ -1,8 +1,117 @@
...
+      \"Purely Functional Data Structures\"
+      [doi:10.1017/CBO9780511530104](https://doi.org/10.1017/CBO9780511530104)
+  *   Niv Dayan, Manos Athanassoulis, and Stratos Idreos.
+      2017.
+      \"Monkey: Optimal Navigable Key-Value Store.\"
+      [doi:10.1145/3035918.3064054](https://doi.org/10.1145/3035918.3064054)
+  *   Subhadeep Sarkar, Dimitris Staratzis, Ziehen Zhu, and Manos Athanassoulis.
+      2021.
+      \"Constructing and analyzing the LSM compaction design space.\"
+      [doi:10.14778/3476249.3476274](https://doi.org/10.14778/3476249.3476274)
```
jorisdral, line 113: My preference is normally to put less in the `description`, and more into the `README`, so that the cabal file stays more concise and mostly deals with build information. The `README` is shipped with the package to users and is a natural place to look for more information. And it is rendered by Hackage too. What do you think?

lsm-tree.cabal
```diff
@@ -172,6 +281,7 @@ library
     Database.LSMTree.Internal.WriteBufferWriter
     Database.LSMTree.Monoidal
     Database.LSMTree.Normal
+    Database.LSMTree.Simple
```
jorisdral, line 284:
```suggestion
```

lsm-tree.cabal
wenkokke: I'd written a script to extract the package description into the README. If the README is rendered by Hackage, then the alternative you're suggesting is fine. Do you have a link to the documentation for Hackage's rendering of Markdown?

wenkokke: Since this feature isn't supported by Haddock, that means the documentation will not be rendered on any non-Hackage server. It also doesn't appear to be documented? Using rendering of Markdown from the README.md file also introduces a lot of difficulty in consistently generating links into the modules. Therefore, I think I'm opposed to this change, and would prefer to keep the documentation in the `description` field.

jorisdral: There is no documentation for the rendering, but It's rendered in both Hackage and Stackage. No hyperrefs to modules is unfortunate, but even then the documentation would still be useful IMO

Maybe others can weigh in if they have preferences, but I won't cling too hard to putting the documentation in the package description

wenkokke: For reference, I've added #661 which adds a script to generate the README from the package description.

jorisdral (approved):
The masses have spoken

PR #655, opened 2025-03-27
fix: move Data.Arena to Database.LSMTree.Internal.Arena

wenkokke:


jorisdral (approved):
LGTM, let's squash the commits and merge

jorisdral (approved):

PR #658, opened 2025-03-27
Make comparison of `Unsliced` values more efficient

jeltsch:
This pull requests makes comparisons of `Unsliced` values more efficient by avoiding conversions of byte arrays to vectors and instead using conversions to short byte strings. This also opens the opportunity to have comparisons of `Unsliced` values for underlying data types other than `SerialisedKey`, and this pull request introduces a corresponding generalization. Note, however, that this latter change implies that comparison now works by directly comparing the underlying byte arrays using the methods of `ShortByteString`; any `Ord` instantiation of the underlying data type is ignored. As of now, this doesn’t make a difference for `SerialisedKey`, as `SerialisedKey` values are compared using the methods of `RawBytes`, which work like the ones for `ShortByteString`.


src/Database/LSMTree/Internal/Unsliced.hs
```diff
@@ -74,8 +75,14 @@ fromUnslicedKey x = SerialisedKey (fromUnsliced (to x))
...
+  Unsliced ba1 == Unsliced ba2 = SBS ba1' == SBS ba2'
+    where
+      !(ByteArray ba1') = ba1
+      !(ByteArray ba2') = ba2
+
+instance Ord (Unsliced a) where
+  compare (Unsliced ba1) (Unsliced ba2) = compare (SBS ba1') (SBS ba2')
+    where
+      !(ByteArray ba1') = ba1
+      !(ByteArray ba2') = ba2
```
jorisdral: Nice, but I don't think we need the generalisation to any `a`. Any `a` might have its own `Ord`/`Eq` instance that deviates from `Ord`/`Eq` on short bytestrings. And we're currently only using it at `SerialisedKey` anyway

jorisdral (commented):
Nice! I added two commits with suggestions, let me know what you think.

jeltsch:
> Let me squash and rebase the branch.

Done.

jorisdral (approved):

PR #661, opened 2025-03-28
feat: generate README from package description

wenkokke:
This PR adds a script that generates the README from the package description.

This depends on #654.


scripts/generate-readme.hs
jorisdral, line 1: If we decide to use this script, then we should probably run it on CI

jorisdral (commented):
It's very cool that we can generate a README from the package description, but it does not help convince me that we should put the documentation in the package description (like we talked about in #661). If anything, duplicating the documentation between the package description and README (even if it can mechanically be kept in sync) makes me think that it would be simpler to put all the text in the README directly so that we don't need the script

scripts/generate-readme.hs
wenkokke: The main argument is that our the list of our internal modules is incredibly overwhelming, and so we really, really need the manually listed links to the public modules.

wenkokke: I can easily write a CI script that generates the README and checks if there are any changes. Running the script on CI and committing it to main is a bad idea.

wenkokke: But it's a good idea to merge running this script into the `pre-commit` script.

jorisdral: > I can easily write a CI script that generates the README and checks if there are any changes. Running the script on CI and committing it to main _from CI_ is a bad idea.

Running the script and checking if there is a non-empty git diff (like we do in the `format-stylish` job) would be my suggestion, not change/commit the README automatically

scripts/generate-readme.hs
```diff
@@ -0,0 +1,67 @@
...
+        , "[![Haskell CI](https://img.shields.io/github/actions/workflow/status/IntersectMBO/lsm-tree/haskell.yml?label=Build)](https://github.com/IntersectMBO/lsm-tree/actions/workflows/haskell.yml)"
+        , "[![Documentation CI](https://img.shields.io/github/actions/workflow/status/IntersectMBO/lsm-tree/documentation.yml?label=Documentation%20build)](https://github.com/IntersectMBO/lsm-tree/actions/workflows/documentation.yml)"
+        , "[![Haddocks](https://img.shields.io/badge/documentation-Haddocks-purple)](https://IntersectMBO.github.io/lsm-tree/)"
+        , ""
+        , "> :warning: **This library is in active development**: there is currently no release schedule!"
+        , ""
+        , "This package is developed by Well-Typed LLP on behalf of Input Output Global, Inc. (IOG) and INTERSECT."
+        , "The main contributors are Duncan Coutts, Joris Dral, Matthias Heinzel, Wolfgang Jeltsch, Wen Kokke, and Alex Washburn."
+        , ""
+        , "## Description"
```
jorisdral, line 49: Alternatively, you could put this in an existing markdown file and `cat` it together with the package description to produce the output file. That would also make it easier to modify the "prelude"

scripts/generate-readme-header.md
```diff
@@ -0,0 +1,15 @@
+<!-- WARNING: Do not edit this file. This file is generated by `scripts/build-readme.hs`. -->
+
+# lsm-tree
+
+[![handbook](https://img.shields.io/badge/policy-Cardano%20Engineering%20Handbook-informational)](https://input-output-hk.github.io/cardano-engineering-handbook)
+[![Haskell CI](https://img.shields.io/github/actions/workflow/status/IntersectMBO/lsm-tree/haskell.yml?label=Build)](https://github.com/IntersectMBO/lsm-tree/actions/workflows/haskell.yml)
+[![Documentation CI](https://img.shields.io/github/actions/workflow/status/IntersectMBO/lsm-tree/documentation.yml?label=Documentation%20build)](https://github.com/IntersectMBO/lsm-tree/actions/workflows/documentation.yml)
```
jorisdral, line 7: BTW, since #665 these links are now dead

jorisdral (approved):
LGTM

PR #664, opened 2025-03-31
fix: do not fail fast in pre-commit.sh

wenkokke:
Superseded by #665.

### DIFF
--- a/scripts/format-cabal.sh
+++ b/scripts/format-cabal.sh
@@ -15,7 +15,7 @@ if [ ! "${cabal_fmt_installed_version}" = "${cabal_fmt_required_version}" ]; the
     exit 1
 fi
 
-# Check Cabal files with cabal-fmt
+# Format Cabal files with cabal-fmt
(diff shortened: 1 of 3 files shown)


jorisdral:
I'd rather have reviewed and merge this PR separately from #665 -- #665 is already quite chunky and the change to he `pre-commit-sh` script seems independent enough that it could be merged separately

wenkokke:
> I'd rather have reviewed and merge this PR separately from #665 -- #665 is already quite chunky and the change to he `pre-commit-sh` script seems independent enough that it could be merged separately

That's fair, but they got rather entwined, so I did this to avoid merge conflicts.

jorisdral:
> > I'd rather have reviewed and merge this PR separately from #665 -- #665 is already quite chunky and the change to he `pre-commit-sh` script seems independent enough that it could be merged separately
> 
> That's fair, but they got rather entwined, so I did this to avoid merge conflicts.

That's also fair, but then it having it in a separate commit (with an explanation of what it's changing) might be useful

PR #665, opened 2025-03-31
 ci: refactor workflows

wenkokke:
This PR constitutes a fairly massive overhaul of the GitHub Actions workflows.

- It ensures that `scripts/pre-commit.sh` behaves the same as the CI linting jobs.
- It refactors steps that form logical units into actions.
- It introduces global defaults for the GHC and Cabal versions and the Cabal project file.
- It removes a number of idiosyncrasies—such as setting `TASTY_TIMEOUT` globally when passing it via the local env key would suffice, or checking the prefix of `matrix.os` instead of the `runner.os` variable.
- It reduces the complexity of the linting jobs by removing all single-entry matrixes.
- It reduces the complexity of the main testing matrix by factoring out shared defaults.
- It enables running the CI workflows locally by introducing guards on `env.ACT` whenever appropriate.
- It creates a `lint-hlint` script that matches the behavior of the `lint-hlint` CI job.
- It creates a `lint-actionlint` CI job and matching script to lint the workflow files.
- It improves robustness by setting a global default for the shell.
- It improves readability of the workflow file by adding headers between the various jobs.
- It improves readability of the workflow logs by adding emoji to the start of each step.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,328 @@
...
+      url: ${{ steps.publish-pages.outputs.page_url }}
+
+    steps:
+      - name: 🛠️ Setup Pages
+        uses: actions/configure-pages@v5
+
+      - name: 📦 Download documentation
+        uses: actions/download-artifact@v4
+        with:
+          name: haddocks-Linux-ghc-${{ env.DEFAULT_GHC_VERSION }}-cabal-${{ env.DEFAULT_CABAL_VERSION }}
```
jorisdral, line 318: I suppose this would break if we ever decide to build the documentation using a non-default ghc and cabal. Might be useful to add a comment about it?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,328 @@
...
+      
+        exclude:
+          # GHC 8.10 is not supported on macOS wit ARM CPU
+          - ghc-version: "8.10"
+            os: "macOS-latest"
+
+        include:
+          # Include builds for various special cases
+          - name: "Build on macOS with Intel CPU"
+            os: "macOS-13"
```
jorisdral, line 171: Testing `ghc-8.10` on MacOS-latest is annoying because it uses an ARM CPU, so that's what the `macOS-13` job was for: testing on Mac with `ghc-8.10`. It's now using the default `ghc-9.6`

jorisdral (commented):
LGTM (almost). I have a few comments but then we can merge

scripts/lint-hlint.sh
```diff
@@ -0,0 +1,23 @@
...
+hlint_actual_version=$(${hlint} --version | head -n 1 | cut -d' ' -f 2 | sed -E 's/v(.*),/\1/')
+if [ ! "${hlint_actual_version}" = "${hlint_expect_version}" ]; then
+    echo "Requires hlint ${hlint_expect_version}; version ${hlint_actual_version} found"
+    exit 1
+fi
+
+# Lint Haskell files with HLint
+echo "Linting Haskell files with HLint..."
+# shellcheck disable=SC2086
+git ls-files --exclude-standard --no-deleted --deduplicate 'src/**/*.hs' 'test/**/*.hs' 'src-extras/**/*.hs' 'src-fcntl-nocache/**/*.hs' 'src-kmerge/**/*.hs' 'src-mcg/**/*.hs' 'src-rocksdb/**/*.hs' | xargs -L50 ${hlint}
```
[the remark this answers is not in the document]
wenkokke: I believe that's the default behaviour, and, as such, yes?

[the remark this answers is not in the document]
wenkokke: However, the interpretation of `fail-on` runs code on top of HLint instead of directly setting an HLint flag, and as such it doesn't have a direct analogy.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,328 @@
...
+       }}
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
+
+    steps:
+    - name: 📥 Checkout repository
+      uses: actions/checkout@v4
+
+    - name: 📥 Checkout repository
+      uses: actions/checkout@v4
```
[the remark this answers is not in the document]
wenkokke: Whoopsie.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,328 @@
...
+      uses: actions/upload-artifact@v4
+      with:
+        name: haddocks-${{ runner.os }}-ghc-${{ matrix.ghc-version || env.DEFAULT_GHC_VERSION }}-cabal-${{ matrix.cabal-version || env.DEFAULT_CABAL_VERSION }}
+        path: haddocks
+        if-no-files-found: error
+        retention-days: 1
+
+    strategy:
+      fail-fast: false
+      matrix:
```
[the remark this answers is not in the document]
wenkokke: Do you mean the individual instantiations or the full matrix?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,328 @@
...
+
+# Set the default shell on all platforms.
+defaults:
+  run:
+    shell: sh
+
+# Set the default GHC and Cabal versions.
+env:
+  DEFAULT_GHC_VERSION: "9.6"
+  DEFAULT_CABAL_VERSION: "3.10"
```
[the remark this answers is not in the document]
wenkokke: Let's leave that for a follow-up PR.

.github/workflows/ci.yml
wenkokke: Oh, I thought it was just general Intel CPU testing. Let me add that back.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,327 @@
...
+        exclude:
+          # GHC 8.10 is not supported on macOS wit ARM CPU
+          - ghc-version: "8.10"
+            os: "macOS-latest"
+
+        include:
+          # Include builds for various special cases
+          - name: "Build on macOS with Intel CPU and GHC 8.10"
+            os: "macOS-13"
+            ghc-version: "8.10"
```
jorisdral, line 169: But here I'm seeing
```
{
  os: macOS-13,
  ghc-version: 8.10,
  name: Build on macOS with Intel CPU and GHC 8.10
}
```
in the output, so `cabal-tests`, `cabal-benchmarks` and `cabal-documentation` are unexpectedly empty

.github/workflows/ci.yml
```diff
@@ -0,0 +1,327 @@
...
+          - name: "Build with +serialblockio"
+            os: "ubuntu-latest"
+            cabal-flags: "+serialblockio"
+          - name: "Build with cabal.project.release"
+            os: "ubuntu-latest"
+            cabal-project-file: "cabal.project.release"
```
jorisdral, line 175: And here
```
 {
  os: ubuntu-latest,
  cabal-project-file: cabal.project.release,
  name: Build with cabal.project.release
}
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,327 @@
...
+            os: "macOS-latest"
+
+        include:
+          # Include builds for various special cases
+          - name: "Build on macOS with Intel CPU and GHC 8.10"
+            os: "macOS-13"
+            ghc-version: "8.10"
+          - name: "Build with +serialblockio"
+            os: "ubuntu-latest"
+            cabal-flags: "+serialblockio"
```
jorisdral, line 172: And here
```
{
  os: macOS-13,
  ghc-version: 8.10,
  name: Build on macOS with Intel CPU and GHC 8.10
}
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,327 @@
...
+      uses: actions/upload-artifact@v4
+      with:
+        name: haddocks-${{ runner.os }}-ghc-${{ matrix.ghc-version || env.DEFAULT_GHC_VERSION }}-cabal-${{ matrix.cabal-version || env.DEFAULT_CABAL_VERSION }}
+        path: haddocks
+        if-no-files-found: error
+        retention-days: 1
+
+    strategy:
+      fail-fast: false
+      matrix:
```
jorisdral, line 150: This looks okay, I think?
```
{
  os: ubuntu-latest,
  ghc-version: 8.10,
  cabal-version: 3.10,
  cabal-flags: ,
  cabal-project-file: ,
  cabal-documentation: false,
  cabal-tests: true,
  cabal-benchmarks: true
}
```

.github/workflows/ci.yml
wenkokke: I misremembered the semantics of `include`, which only modifies original entries. I've fixed this by relying on the absence, and renaming `cabal-tests` to `cabal-skip-tests` and likewise for the benchmarks.

jorisdral (approved):

PR #667, opened 2025-04-02
Use `hlint` on every Haskell file

jorisdral:
Also, clean up HLINT pragmas by replacing them with `ignore` entries in the `.hlint.yaml`.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -27,6 +27,15 @@
 - ignore: {name: "Use shows"}
 - ignore: {name: "Use fmap"}
 - ignore: {name: "Use <=<"}
+- ignore: {name: "Use void"}
+- ignore: {name: "Use zipWith"}
+- ignore: {name: "Evaluate"}
+- ignore: {name: "Redundant if"}
+- ignore: {name: "Use catMaybes"}
+- ignore: {name: "Replace case with maybe"}
+- ignore: {name: "Redundant =="}
+- ignore: {name: "Hoist not"}
+- ignore: {name: "Use /="}
 
 # Specify additional command line arguments
 #
--- a/scripts/lint-hlint.sh
+++ b/scripts/lint-hlint.sh
@@ -20,4 +20,4 @@ fi
 # Lint Haskell files with HLint
 echo "Linting Haskell files with HLint..."
 # shellcheck disable=SC2086
-git ls-files --exclude-standard --no-deleted --deduplicate 'src/**/*.hs' 'test/**/*.hs' 'src-extras/**/*.hs' 'src-fcntl-nocache/**/*.hs' 'src-kmerge/**/*.hs' 'src-mcg/**/*.hs' 'src-rocksdb/**/*.hs' | xargs -L50 ${hlint}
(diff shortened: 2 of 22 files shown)


.hlint.yaml
```diff
@@ -27,6 +27,16 @@
...
 - ignore: {name: "Use fmap"}
 - ignore: {name: "Use <=<"}
+- ignore: {name: "Use void"}
+- ignore: {name: "Use zipWith"}
+- ignore: {name: "Evaluate"}
+- ignore: {name: "Redundant if"}
+- ignore: {name: "Use catMaybes"}
+- ignore: {name: "Replace case with maybe"}
+- ignore: {name: "Redundant =="}
+- ignore: {name: "Use fewer imports"}
```
wenkokke, line 37: I don't think we should be ignoring this, especially globally.

scripts/lint-hlint.sh
```diff
@@ -20,4 +20,4 @@ fi
 # Lint Haskell files with HLint
 echo "Linting Haskell files with HLint..."
 # shellcheck disable=SC2086
-git ls-files --exclude-standard --no-deleted --deduplicate 'src/**/*.hs' 'test/**/*.hs' 'src-extras/**/*.hs' 'src-fcntl-nocache/**/*.hs' 'src-kmerge/**/*.hs' 'src-mcg/**/*.hs' 'src-rocksdb/**/*.hs' | xargs -L50 ${hlint}
+git ls-files --exclude-standard --no-deleted --deduplicate '**.hs' | xargs -L50 ${hlint}
```
wenkokke, line 23: Just `'*.hs'` is sufficient.

src-extras/Database/LSMTree/Extras/MergingTreeData.hs
```diff
@@ -374,7 +374,6 @@ shrinkMergingTreeData shrinkKey shrinkVal shrinkBlob = \case
     | mr' <- shrinkMergingRunData shrinkKey shrinkVal shrinkBlob mr
     ]
   PendingLevelMergeData prs t ->
-    {- HLINT ignore "Use catMaybes" -}
```
wenkokke, line 377: I mean... HLint is right. Using `catMaybes` is equally performant and total.

test/Test/Database/LSMTree/Internal/RawBytes.hs
```diff
@@ -1,11 +1,8 @@
-{- HLINT ignore "Avoid restricted alias" -}
```
wenkokke, line 1: We shouldn't ignore this.

jorisdral: It's removed now, so I'm not sure what you mean

.hlint.yaml
wenkokke: "Use fewer imports"

test/Test/Database/LSMTree/Internal/RawBytes.hs
wenkokke: Just in case you were moving it to the global file.

wenkokke (approved):

PR #672, opened 2025-04-07
fix(simple): finalise documentation

wenkokke:


lsm-tree.cabal
```diff
@@ -52,45 +52,103 @@ description:
...
-  As such, the worst-case complexity of basic operations refer to the number of disk I\/O actions.
-
-  TODO: Describe the time complexity of the basic operations.
+  The worst-case behaviour of the library is described using [big-O notation](http://en.wikipedia.org/wiki/Big_O_notation).
+  The documentation provides two measures of complexity:
+
+  *   The time complexity of operations is described in terms of the number of disk I\/O operations and referred to as the disk I\/O complexity.
+      In practice, the time of the operations on LSM-trees is dominated by the number of disk I\/O actions.
+  *   The space complexity of tables is described in terms of the in-memory size of an LSM-tree table.
+      Both the in-memory and on-disk size of an LSM-tree table scales linearly.
```
jorisdral, line 61:
```suggestion
      Both the in-memory and on-disk size of an LSM-tree table scale linearly.
```

jorisdral, line 61: Scale linearly with what?

lsm-tree.cabal
```diff
@@ -52,45 +52,103 @@ description:
...
+      This is discussed in detail below, under [In-memory size of tables](#performance_size).
+
+  The complexities are described in terms of the following variables and constants:
+
+  *   The variable \(n\) refers to the number of /physical/ table entries.
+      A /physical/ table entry is any key–operation pair, e.g., @Insert k v@ or @Delete k@, whereas a /logical/ table entry is determined by all physical entries with the same key.
+      If the variable \(n\) is used to describe the complexity of an operation that involves multiple tables, it refers to the sum of all table entries.
+  *   The constant \(T\) refers to the size ratio of the table, which is a configuration parameter.
+
+  === Disk I\/O cost of operations #performance_time#
```
jorisdral, line 72: Small thing: we're summarising the complexity here but it is also documented in the `Simple` module, so it seems likely that it would get out of sync in the future if we're not careful

lsm-tree.cabal
```diff
@@ -52,45 +52,103 @@ description:
...
+  +----------+------------------------+----------------------------+------------------------------+
+  | Resource | Operation              | Merge policy               | Cost in disk I\/O operations |
+  +==========+========================+============================+==============================+
+  | Session  | Create\/Open           | N\/A                       | \(O(1)\)                     |
+  +----------+------------------------+----------------------------+------------------------------+
+  |          | Close                  | N\/A                       | \(O(t \: \log_T \: n)\)*     |
+  +----------+------------------------+----------------------------+------------------------------+
+  | Table    | Create                 | N\/A                       | \(O(1)\)                     |
+  +----------+------------------------+----------------------------+------------------------------+
+  |          | Close                  | N\/A                       | \(O(\log_T \: n)\)           |
```
jorisdral, line 88: Functions like this where the complexity depends on the number of runs in a table, should also depends on the merge policy: if $\log_T n$ approximates the number of levels, then with tiering there are at most $T-1$ runs per level, and with levelling at most $1$ per level

src/Database/LSMTree/Simple.hs
```diff
@@ -539,6 +624,8 @@ Throws the following exceptions:
 ['TableCorruptedError']:
     If the table data is corrupted.
 -}
+-- The worst-case time complexity is \(O(b \log_T n)\).
+-- The amortised time complexity is \(\Theta(b \log\log_T n)\).
```
jorisdral, line 628: Are these comments about time complexity supposed to stay in the code?

lsm-tree.cabal
```diff
@@ -52,45 +52,103 @@ description:
...
+
+  === In-memory size of tables #performance_size#
 
   The in-memory size of an LSM-tree is described in terms of the variable \(n\), which refers to the number of /physical/ database entries.
   A /physical/ database entry is any key–operation pair, e.g., @Insert k v@ or @Delete k@, whereas a /logical/ database entry is determined by all physical entries with the same key.
 
   The worst-case in-memory size of an LSM-tree is \(O(n)\).
 
-  *   The worst-case size of the write buffer is \(O(1)\).
+  *   The worst-case in-memory size of the write buffer is \(O(1)\).
```
jorisdral, line 124: I wonder if we can really consider the write buffer size a constant. It's a parameter to the table, no? What are your thoughts about this?

lsm-tree.cabal
```diff
@@ -52,45 +52,103 @@ description:
...
+  +----------+------------------------+----------------------------+------------------------------+
+  | Table    | Create                 | N\/A                       | \(O(1)\)                     |
+  +----------+------------------------+----------------------------+------------------------------+
+  |          | Close                  | N\/A                       | \(O(\log_T \: n)\)           |
+  +----------+------------------------+----------------------------+------------------------------+
+  |          | Lookup                 | @MergePolicyLazyLevelling@ | \(O(\log_T \: n)\)           |
+  +----------+------------------------+----------------------------+------------------------------+
+  |          | Range Lookup           | N\/A                       | \(O(\log_T \: n + b)\)**     |
+  +----------+------------------------+----------------------------+------------------------------+
+  |          | Insert\/Delete\/Update | @MergePolicyLazyLevelling@ | \(O(\log_T \: n)\)           |
```
jorisdral, line 94: I'm not sure if simplification has done the complexities justice. We're omitting bloom filter false positive rates, write buffer size, number of entries per page, page size, distribution of inserted keys, etc., and that makes the bounds very loose. 

 I'm currently leaning towards suggesting to make these complexity bounds detailed rather than simple, following the Monkey paper more closely. What are your thoughts about this?

lsm-tree.cabal
```diff
@@ -52,45 +52,103 @@ description:
...
+  +----------+------------------------+----------------------------+------------------------------+
+  |          | Duplicate              | N\/A                       | \(O(1)\)                     |
+  +----------+------------------------+----------------------------+------------------------------+
+  |          | Union                  | N\/A                       | \(O(n)\)                     |
```
jorisdral, line 98: Are we missing `incrementalUnion` in this list?

jorisdral (commented):
Not a full review, but here are some initial thoughts

lsm-tree.cabal
wenkokke: Regardless, the number of runs is proportional to the number of levels, so the big-O complexity is the same, right?

jorisdral: I'm a little bit hazy on the specific rules for big-O complexity, but I tend to consider $T$ to be a parameter to the algorithm (even if it is currently fixed to 4), in which case it shouldn't be considered a constant value I think

If $T$ is considered constant, shouldn't $\log_T$ then equivalent under big-O to $\log_2$?

lsm-tree.cabal
wenkokke: It's a constant, because it's a parameter. We can use `N` to represent it, but in most cases it gets absorbed by the O.

lsm-tree.cabal
wenkokke: No, because putting it there would be confusing. Incremental unions are still `O(n)`, except you do them in `n` steps of `O(1)`

src/Database/LSMTree/Simple.hs
wenkokke: Yes.

lsm-tree.cabal
wenkokke: True.

lsm-tree.cabal
```diff
@@ -52,45 +52,108 @@ description:
...
+  +----------+------------------------+----------------------------+------------------------------------------------+
+
+  (*The variable \(b\) refers to the number of entries retrieved by the range lookup.)
+
+  === In-memory size of tables #performance_size#
 
   The in-memory size of an LSM-tree is described in terms of the variable \(n\), which refers to the number of /physical/ database entries.
   A /physical/ database entry is any key–operation pair, e.g., @Insert k v@ or @Delete k@, whereas a /logical/ database entry is determined by all physical entries with the same key.
 
   The worst-case in-memory size of an LSM-tree is \(O(n)\).
```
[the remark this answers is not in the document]
wenkokke: No, it doesn't matter. It scales linearly whether its bytes or terabytes.

lsm-tree.cabal
```diff
@@ -52,45 +52,111 @@ description:
...
+
+  === Disk I\/O cost of operations #performance_time#
+
+  The following table summarises the cost of the operations on LSM-trees
+  measured in the number of disk I\/O operations.
+  If the cost depends on the merge policy, the table contains one entry for each merge policy.
+  Otherwise, the merge policy is listed as N\/A.
+
+  +----------+------------------------+----------------------------+------------------------------------------------+
+  | Resource | Operation              | Merge policy               | Cost in disk I\/O operations                   |
```
jorisdral, line 86: Range lookup, union, and read next entry also rely on $P$

jorisdral (approved):
LGTM!

PR #674, opened 2025-04-07
fix: update pre-commit script

wenkokke:
* fix: remove cabal-docspec from pre-commit
* fix: add generate-readme to pre-commit
* fix: add  echo to generate-readme


wenkokke:
> Why did `test-cabal-docspec.sh` have to go?

Tests don't belong in the pre-commit script. It should never have been there. It takes way too long.

PR #677, opened 2025-04-09
GHA: use `cabal-3.12`

jorisdral:
`cabal-3.10` has a bug in `cabal haddock-project`, which leads to bad hyperlinks in our published haddocks

Also: my editor automatically removed trailing whitespace in the workflow file

### DIFF
--- a/lsm-tree.cabal
+++ b/lsm-tree.cabal
@@ -77,7 +77,7 @@ description:
 
       [@AllocFixed bitsPerPhysicalEntry@]:
           The total size of all Bloom filters is the number of bits per physical entry multiplied by the number of physical entries.
-      [@AllocRequestFPR requestedFPR@]:      
+      [@AllocRequestFPR requestedFPR@]:
           TODO: How does one determine the bloom filter size using @AllocRequestFPR@?
 
   *   The worst-case size of the indexes is \(O(n)\).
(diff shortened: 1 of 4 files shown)


lsm-tree.cabal
```diff
@@ -337,7 +341,7 @@ library xxhash
   other-modules:   FFI
   hs-source-dirs:  xxhash/src
   build-depends:
-    , base
+    , base        <5
```
wenkokke, line 344: Do we need the constraint in base everywhere? I'm guessing one on lsm-tree and one on blockio-api suffices?

jorisdral: Unfortunately, this is what `cabal check` wants

wenkokke: Boo

wenkokke:
@jorisdral This will fail for GHC 8.10.7, so you'd have to use the previous version if `ghc --numeric-version | cut -d'.' -f1` returned `8`.

wenkokke:
> @jorisdral This will fail for GHC 8.10.7, so you'd have to use the previous version if `ghc --numeric-version | cut -d'.' -f1` returned `8`.

Never mind, it appears to work!

wenkokke (approved):
We should setup proper pre-commit hooks so we don't get this whitespace issue.

PR #683, opened 2025-04-14
Add script for finding missing `IO` specializations

jeltsch:
This pull requests adds a script that helps in finding missing `IO` specializations, along with some example input and output.

When running the command `scripts/find-missing-io-specialisations.sh src/**/*.hs` on my Ubuntu 24.04 machine while being at the current `main` tip, 78a32313952faa79b2ee7f799244133f62c073eb, I get the following output:

    Database.LSMTree.Internal.Arena.newArenaManager
    Database.LSMTree.Internal.Arena.withArena
    Database.LSMTree.Internal.Arena.scrambleArena
    Database.LSMTree.Internal.Arena.scrambleBlock
    Database.LSMTree.Internal.Arena.withUnmanagedArena
    Database.LSMTree.Internal.Arena.allocateFromArena
    Database.LSMTree.Internal.Arena.newBlockWithFree
    Database.LSMTree.Internal.BlobFile.readBlob
    Database.LSMTree.Internal.BlobRef.releaseBlobRef
    Database.LSMTree.Internal.BlobRef.readRawBlobRef
    Database.LSMTree.Internal.ChecksumHandle.dropCache
    Database.LSMTree.Internal.ChecksumHandle.closeHandle
    Database.LSMTree.Internal.CRC32C.expectValidFile
    Database.LSMTree.Internal.doesSnapshotDirExist
    Database.LSMTree.Internal.IncomingRun.newIncomingSingleRun
    Database.LSMTree.Internal.IncomingRun.newIncomingMergingRun
    Database.LSMTree.Internal.IncomingRun.depositNominalCredits
    Database.LSMTree.Internal.Merge.writeReaderEntry
    Database.LSMTree.Internal.Merge.writeSerialisedEntry
    Database.LSMTree.Internal.MergeSchedule.mkUnionCache
    Database.LSMTree.Internal.MergeSchedule.duplicateUnionCache
    Database.LSMTree.Internal.MergeSchedule.releaseUnionCache
    Database.LSMTree.Internal.MergingRun.unsafeNew
    Database.LSMTree.Internal.MergingRun.mergeType
    Database.LSMTree.Internal.MergingRun.atomicReadCredits
    Database.LSMTree.Internal.MergingRun.atomicModifyInt
    Database.LSMTree.Internal.MergingRun.supplyChecked
    Database.LSMTree.Internal.MergingRun.supplyCreditsRelative
    Database.LSMTree.Internal.MergingRun.supplyCreditsAbsolute
    Database.LSMTree.Internal.MergingTree.expectCompleted
    Database.LSMTree.Internal.MergingTree.Lookup.mapMStrict
    Database.LSMTree.Internal.RunReader.seekToDiskPage
    Database.LSMTree.Internal.UniqCounter.newUniqCounter
    Database.LSMTree.Internal.UniqCounter.incrUniqCounter
    Database.LSMTree.Internal.Vector.mapMStrict
    Database.LSMTree.Internal.Vector.imapMStrict
    Database.LSMTree.Internal.Vector.forMStrict
    Database.LSMTree.Internal.Vector.unsafeInsertWithMStrict

It should be checked that the same output arises on other systems, in particular such that do not use GNU sed.


wenkokke:
As a preliminary note, it would be good to change the script to require no parameters (ie inline the source glob) and perhaps rename it to start with lint, for consistency.

wenkokke:
> What about keeping the script as it is and adding a wrapper script that passes the main library’s modules and have its name start with `lint-`?

Sounds great! Another option would be to default to all Haskell files (or just `src/**/*.hs`) but allow the first argument to the script to override this.

scripts/find-missing-io-specialisations.sh
```diff
@@ -0,0 +1,91 @@
...
+#         specialisation has been found yet in the current module or the
+#         most recently found `IO` specialisation is considered to not
+#         be relevant for the remainder of the module.
+
+hic='[[:alnum:]_#]' # Haskell identifier character
+
+LC_COLLATE=C LC_CTYPE=C sed -En '
+    :start
+    /^module / {
+        s/module +([^ ]*).*/\1 /
```
wenkokke, line 43: Might be good to explicitly match ` |$` after the capture group for clarity, to not depend on greediness?

wenkokke:
OTOH, it's sufficiently clear the script isn't malicious, so once there's a CI job that runs it on macOS and Windows, I'm happy to merge it under the understanding that it's not a complete guarantee on the absence of missing specialisations... which is true anyway, given the assumptions it makes.

jorisdral:
Are there cases where the script returns false positives, and if not now, is it conceivable that it could do so in the future? If so, it might be good to think about how such false positives could be addressed, specially if it's a CI job we run that can hold up PRs

wenkokke:
> Are there cases where the script returns false positives, and if not now, is it conceivable that it could do so in the future? If so, it might be good to think about how such false positives could be addressed, specially if it's a CI job we run that can hold up PRs 

@jeltsch already floated the idea of an allowlist for false positives.

jeltsch:
f0f80d6bc61c41b69539977e15c429ec48bfa1d2 disables reporting of operations with inline directives. The expected result of running `scripts/find-missing-io-specialisations.sh src/**/*.hs` is now as follows:
```
Database.LSMTree.Internal.Arena.newArenaManager
Database.LSMTree.Internal.Arena.scrambleArena
Database.LSMTree.Internal.Arena.scrambleBlock
Database.LSMTree.Internal.Arena.withUnmanagedArena
Database.LSMTree.Internal.Arena.allocateFromArena
Database.LSMTree.Internal.Arena.newBlockWithFree
Database.LSMTree.Internal.ChecksumHandle.dropCache
Database.LSMTree.Internal.ChecksumHandle.closeHandle
Database.LSMTree.Internal.CRC32C.expectValidFile
Database.LSMTree.Internal.doesSnapshotDirExist
Database.LSMTree.Internal.IncomingRun.depositNominalCredits
Database.LSMTree.Internal.MergeSchedule.mkUnionCache
Database.LSMTree.Internal.MergeSchedule.duplicateUnionCache
Database.LSMTree.Internal.MergeSchedule.releaseUnionCache
Database.LSMTree.Internal.MergingTree.expectCompleted
Database.LSMTree.Internal.MergingTree.Lookup.mapMStrict
Database.LSMTree.Internal.RunReader.seekToDiskPage
```

scripts/find-missing-io-specialisations.sh
jeltsch: I used to put such extra information about delimiters into regular expressions, but now think that it’s fine to rely on greediness. After all, we’re relying on it in other situations already, where it’s hard or even impossible not to do so.

jeltsch:
> OTOH, it's sufficiently clear the script isn't malicious, so once there's a CI job that runs it on macOS and Windows, I'm happy to merge it under the understanding that it's not a complete guarantee on the absence of missing specialisations... which is true anyway, given the assumptions it makes.

Are you thinking here of a CI job that applies the script to the codebase to ensure that there are no missing specializations or one that tests the script using the data in `scripts/find-missing-io-specialisations.test`?

Beside that, do you think it is necessary to have CI run this script on Windows? After all, it’s not part of the software we want to ship in the end but a utility intended to help *us* (and perhaps future contributors). I mean, if supporting Windows, using some Unix compatibility layer, is easy, then I’m not against it; I just don’t want to spend much effort on making this script Windows-compatible (note that initially I intended this script to just assist me in a particular task and not even make it into the repository 🙂)

jeltsch:
> I think it might be a good idea to add dinner comments to the sed script that explain the logic a bit, not per se on a command by command basis, but to explain the algorithm.

Done in 325c20e7c9e1fd13b5aa43ec4733a2d367bb006d.

jeltsch:
> > What about keeping the script as it is and adding a wrapper script that passes the main library’s modules and have its name start with `lint-`?
> 
> Sounds great!

I’ve realized this in 539f06ae923f419ca19402e06cb3fb57f71f8731. The wrapper script not only automatically picks the source files of the main library but also enables users to explicitly allow certain missing `IO` specializations.

wenkokke:
I've opened #697 which contains the changes in this PR, plus a few fixes.

- The top-level script uses `set -e` which is not a POSIX shell feature and only works on Ubuntu because it lies and symlinks `/bin/sh` to `/bin/bash`.
- The sed script does not work with macOS sed. Rather than fixing the sed script, I've altered the shell script to ask for `gsed`.

jeltsch:
I’ve added another three commits, which contain changes from #697 and variants thereof.

jeltsch:
> The sed script does not work with macOS sed. Rather than fixing the sed script, I've altered the shell script to ask for `gsed`.

This was apparently a good idea. As we discussed already in the meeting, there seems to be no easy way to compare strings with `sed` in a fully POSIX-compatible way, as POSIX doesn’t guarantee backreferences in regular expressions to work.

wenkokke (approved):
LGTM.

dcoutts (approved):
This all looks very impressive. I've never seen such a sophisticated sed script. All my sed scripts are one liners!

jeltsch:
> LGTM. Are you happy to merge this in the current state, @jeltsch?

Yes. 🙂 I’ll rebase it in a minute and then initiate the merging.

PR #685, opened 2025-04-16
Add missing `IO` and `ST` specializations

jeltsch:
This pull request adds missing `IO` specializations and, in `Database.LSMTree.Internal.Arena`, where `ST` specializations are generally present as well, also missing `ST` specializations.


wenkokke (approved):
LGTM!

PR #686, opened 2025-04-18
Include C header files and license files in the distributed package

jorisdral:
C header files are included in the source files distributed for the `lsm-tree`
package to prevent build failures in Haskell projects that depend on `lsm-tree`.
We also include the license files for the vendored `xxHash` and `bloomfilter`
packages, as is required by the license types.


lsm-tree.cabal
```diff
@@ -195,22 +195,30 @@ description:
...
+build-type:         Simple
 tested-with:
   GHC ==8.10 || ==9.2 || ==9.4 || ==9.6 || ==9.8 || ==9.10 || ==9.12
 
+extra-doc-files:    CHANGELOG.md
+extra-source-files:
+  xxhash/include/HsXXHash.h
+  xxhash/xxHash-0.8.2/xxhash.h
+
+license-files:
```
wenkokke, line 218: Oh, wow, there's a cabal key for this now? Very cool.

wenkokke (approved):
LGTM!

PR #687, opened 2025-04-21
Some clarifications in the public documentation for table unions

jorisdral:


src/Database/LSMTree.hs
```diff
@@ -1449,7 +1449,11 @@ withUnions tables =
   bracket (unions tables) closeTable
 
 {- |
-Create a table that contains the left-biased union of the entries of the given tables.
+Create a table that contains the union of the entries of the given tables.
+
+If the given key is a member of a single input table, then the same key and value occur in the output table.
+Otherwise, the values for duplicate keys are combined using 'resolve'.
```
wenkokke, line 1455: Maybe say "from left to right" or "in the same order as the tables" here?

src/Database/LSMTree.hs
```diff
@@ -1669,6 +1673,12 @@ remainingUnionDebt (Table table) =
 {- |
 Supply the given amount of union credits.
 
+This reduces the union debt by /at least/ the number of supplied union credits.
+It is therefore advisable to query 'remainingUnionDebt' every once in a while to see what the current debt is.
+
+This function returns any surplus of union credits as /leftover/ credits when a union has finished.
+In particular, if the returned number of credits is non-negative, then the union is finished.
```
wenkokke, line 1680: It returns the negative of remaining union debt otherwise?

wenkokke: Maybe add that as a property if so.

wenkokke (approved):

src/Database/LSMTree.hs
wenkokke: Zero is non-negative, so the previous line should read positive instead.

wenkokke: Also in the related  case below.

PR #692, opened 2025-04-22
Drop support for `ghc-8.10`

jorisdral:
Recently, `cardano-node` releases have started using `ghc-9.6` instead of `ghc-8.10`. Dropping support for `ghc-8.10` in `lsm-tree` makes our lives a bit easier, because we can use less conditional compilation in our codebase. We'll now support versions `ghc-9.2` and up, because we weren't explicitly supporting `ghc-9.0` anyway

### DIFF
--- a/src/Database/LSMTree/Internal/BloomFilterQuery1.hs
+++ b/src/Database/LSMTree/Internal/BloomFilterQuery1.hs
@@ -42,37 +42,13 @@ type RunIx = Int
 --
 newtype RunIxKeyIx = MkRunIxKeyIx Word32
   deriving stock Eq
-#if MIN_VERSION_GLASGOW_HASKELL(9,0,0,0)
   deriving newtype P.Prim
-#else
-instance P.Prim RunIxKeyIx where
-    sizeOfType# _ = 4#
-    alignmentOfType# _ = 4#
-
-    indexByteArray# ba i =
-      MkRunIxKeyIx (P.indexByteArray# ba i)
-    readByteArray# ba i s =
-      case P.readByteArray# ba i s of
-        (# s', w #) -> (# s', MkRunIxKeyIx w #)
-    writeByteArray# ba i (MkRunIxKeyIx w) s =
-      P.writeByteArray# ba i w s
-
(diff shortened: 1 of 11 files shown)


jorisdral:
> If it's all there, shouldn't we just do a first release with support for 8.10 and drop it in the second?

I don't think it's very useful to do so. I can understand it from the point of view that there is at least *some* released version that supports `ghc-8.10` for users that want it, but we won't commit to supporting it into the future. Not supporting it from the get-go is in the end more useful for us as developers (no future support, no backporting, less maintenance, no users asking for backports/support). Morever, users that really want to use `ghc-8.10` wouldn't be able to meaningfully upgrade their dependency on `lsm-tree` anyway if we drop it immediately in the next release

src/Database/LSMTree/Internal/BloomFilterQuery1.hs
```diff
@@ -42,37 +42,14 @@ type RunIx = Int
...
-
-    indexOffAddr# ba i =
-      MkRunIxKeyIx (P.indexOffAddr# ba i)
-    readOffAddr# ba i s =
-      case P.readOffAddr# ba i s of
-        (# s', w #) -> (# s', MkRunIxKeyIx w #)
-    writeOffAddr# ba i (MkRunIxKeyIx w) s =
-      P.writeOffAddr# ba i w s
-#endif
+
```
jeltsch, line 46: This extra empty line shouldn’t be there, I guess.

jeltsch (approved):
Looks pretty good to me. Just one comment further below and the following two recommendations:

* The following code should be removed: https://github.com/IntersectMBO/lsm-tree/blob/97f498af4b9d143d9608e5753bde3b63bc9fcdf2/cabal.project.release#L25-L29
* In the following code, the lower bound for the `base` version should be changed to `4.16`: https://github.com/IntersectMBO/lsm-tree/blob/97f498af4b9d143d9608e5753bde3b63bc9fcdf2/scripts/generate-readme.hs#L4

dcoutts (approved):
LGTM!

PR #695, opened 2025-04-29
Include the NOTICE file in package distr

jorisdral:


wenkokke (approved):

wenkokke:
The name of the PR may be a bit misleading, since you're also adding the source repair declaration.

PR #698, opened 2025-05-01
doc(full): add missing examples

wenkokke:


src/Database/LSMTree.hs
```diff
@@ -1656,10 +1722,28 @@ _withInternalTables (Table (table :: Internal.Table m h) :| tables) action =
     | otherwise = throwIO $ ErrTableUnionHandleTypeMismatch 0 (typeRep $ Proxy @h) i (typeRep $ Proxy @h')
 
 {- |
-Get the amount of remaining union debt.
+Get an /upper bound/ for the amount of remaining union debt.
```
jorisdral, line 1725: Union debt is already defined to be an upper bound: 

> The amount of union debt is an upper bound on the amount of computation that
needs to be performed before an incremental union is completed.

So you could now read this as "an upper bound on an upper bound"

Same is true in a few other places in this module

src/Database/LSMTree.hs
jorisdral, line 1: Do these new examples have to be ported to the `Simple` module as well?

src/Database/LSMTree.hs
```diff
@@ -1393,6 +1391,23 @@ duplicate (Table table) =
...
+    LSMT.withUnion table1 table2 $ \table3 -> do
+      print =<< LSMT.lookup table3 0
+      print =<< LSMT.lookup table3 1
+    print =<< LSMT.lookup table1 0
+    print =<< LSMT.lookup table2 0
+:}
+Found (Value "Hello World")
+Found (Value "Goodbye")
+Found (Value "Hello")
+Found (Value "World")
```
jorisdral, line 1409: This example is now duplicated: alternatively, you could only put it in on `withUnion` or `union`

Same is true for other functions that have a `with*` version

src/Database/LSMTree.hs
wenkokke: Fair, so we should redefine union debt, because debt as a concept is a number not an upper bound.

src/Database/LSMTree.hs
wenkokke: Most of them don't, since the only interesting edge cases arise in the full module, but some could be.

src/Database/LSMTree.hs
wenkokke: They should be different in that one version uses bracket.

wenkokke: Either way, whichever function you look at should have an example.

src/Database/LSMTree.hs
wenkokke: Fixed in 4c07e5d16d0a7072e292c2ed11bed6487838eed5

src/Database/LSMTree/Internal/Unsafe.hs
```diff
@@ -1685,9 +1685,10 @@ ensureSessionsMatch (t :| ts) = do
 -------------------------------------------------------------------------------}
 
 {- |
-The amount of union debt is an upper bound on the amount of computation that
-needs to be performed before an incremental union is completed. This includes
-the cost of completing incremental unions that were part of a union's input.
+Union debt represents the amount of computation that must be performed before an incremental union is completed.
```
jorisdral, line 1688: I'm nitpicking but I think that union debt was defined as being upper bound because it does not exactly equal the amount of computation. Changing the definition of union debt makes this less clear. Ideas?

wenkokke: I'm happy to rename the type `UnionDebtBound`, but that gets a bit messy.

jorisdral: Let's keep it as is for now -- if we could scale credits internally, then maybe the upper bound thing goes away

jorisdral (approved):

PR #701, opened 2025-05-01
doc: document table configuration

wenkokke:
- [X] `TableConfig`
- [ ] `defaultTableConfig`
- [ ] `MergePolicy`
- [ ] `SizeRatio`
- [ ] `WriteBufferAlloc`
- [ ] `BloomFilterAlloc`
- [ ] `FencePointerIndexType`
- [ ] `DiskCachePolicy`
- [ ] `MergeSchedule`


src/Database/LSMTree/Internal/Config.hs
```diff
@@ -49,26 +49,35 @@ newtype LevelNo = LevelNo Int
...
---
--- * Merge policy: Tiering
---
--- * Size ratio: 4
+{- |
+A collection of configuration parameters for tables, which can be used to tune the performance of the table.
+
+Each parameter is associated with its own type.
+Detailed discussion of the use of each parameter can be found in the documentation for its associated type.
+For an overview, see the [Performance](../#performance) section in the package description.
```
jorisdral, line 57: Is this a functional hyperlink?

src/Database/LSMTree/Internal/Config.hs
```diff
@@ -49,26 +49,35 @@ newtype LevelNo = LevelNo Int
...
   , confBloomFilterAlloc  :: !BloomFilterAlloc
+    -- | The fence pointer index type determines the type of indexes,
+    --   which affects the in-memory size of the table and may constrain the table keys.
   , confFencePointerIndex :: !FencePointerIndexType
-    -- | The policy for caching key\/value data from disk in memory.
+    -- | The disk cache policy determines the policy for caching data from disk in memory,
+    --   which may affect the performance of lookup operations.
   , confDiskCachePolicy   :: !DiskCachePolicy
+    -- | The merge schedule determines how the table manages its data.
   , confMergeSchedule     :: !MergeSchedule
```
jorisdral, line 81: What's the difference between "manage data" and "organise data"? It's not so clear to me

src/Database/LSMTree/Internal/Config.hs
wenkokke: Yes, but I'll admit it's brittle.

src/Database/LSMTree/Internal/Config.hs
wenkokke: Fair. It's difficult to describe without resorting to explaining implementation details.

wenkokke: Manage is acting on, organize is laying out.

jorisdral: The size ratio, merge policy and merge schedule all in some sense determine the layout of a table. It's hard (or impossible) to get the same layout with different options, I think?

I think for the configuration options, it's not bad to resort to *some* implementation details, because the configuration options are pretty implementation specific anyway. However, I suppose you could argue that the user only has to know how the settings affect disk I/O to pick meaningful configurations, and then the user doesn't need to know so many implementation details. But what if a user does want to understand configuration options better, how would they find out what the options mean?

dcoutts:
Ahh, the docspec error is because I changed the AllocFixed from Int to Double. So that needs updating.

dcoutts:
Ah, and you did that just as I was writing :smile:

README.md
```diff
@@ -276,81 +278,383 @@ schedule is listed as N/A.
...
-  `TableConfig`. Regardless of write buffer allocation strategy, the
-  size of the write buffer may never exceed 4GiB.
+- The fence-pointer indexes would store approximately 2.29 million keys,
+  which is approximately 74.41MiB.
 
-  `AllocNumEntries maxEntries`  
-  The maximum size of the write buffer is the maximum number of entries
-  multiplied by the average size of a key–operation pair.
+- The Bloom filters would use 9.85 bits per entry, which is
+  approximately 939.37MiB.
```
dcoutts, line 302: I think you've missed the conversion from bits to bytes!
```
> (100e6 * 9.85) / 1024 / 1024
939.3692016601563
```
but for bytes need to divide by 8
```
> (100e6 * 9.85 / 8) / 1024 / 1024
117.42115020751953
```

README.md
```diff
@@ -276,81 +278,383 @@ schedule is listed as N/A.
...
 
-The in-memory size of an LSM-tree is described in terms of the variable
-$`n`$, which refers to the number of *physical* database entries. A
-*physical* database entry is any key–operation pair, e.g., `Insert k v`
-or `Delete k`, whereas a *logical* database entry is determined by all
-physical entries with the same key.
+TODO: Choose default Bloom filter size.
 
-The worst-case in-memory size of an LSM-tree is $`O(n)`$.
+The in-memory size would be approximately 0.99GiB:
```
dcoutts, line 293: So this will need updating too. Bits vs bytes makes a big difference to the size ratio between memory and disk! :smiley:

README.md
```diff
@@ -276,81 +278,383 @@ schedule is listed as N/A.
...
+  which is approximately 74.41MiB.
 
-  `AllocNumEntries maxEntries`  
-  The maximum size of the write buffer is the maximum number of entries
-  multiplied by the average size of a key–operation pair.
+- The Bloom filters would use 9.85 bits per entry, which is
+  approximately 939.37MiB.
 
-- The worst-case in-memory size of the Bloom filters is $`O(n)`$.
+The in-memory size would be approximately 1.54GiB:
```
dcoutts, line 304: I'm not following what we're presenting now. This seems to be a duplicate section with different numbers (16 bits per key?)

README.md
```diff
@@ -276,81 +278,383 @@ schedule is listed as N/A.
...
+The *write buffer capacity* balances the performance of lookups and
+updates against the in-memory size of the table. If the write buffer is
+larger, it takes up more memory, but lookups and updates are more
+efficient. This parameter is referred to as $`B`$ in the disk I/O cost
+of operations. Irrespective of this parameter, the write buffer size
+cannot exceed 4GiB.
+
+`confMergeSchedule`  
+The *merge schedule* balances the performance of lookups and updates
+against the consistency of updates. The merge schedule does not affect
```
dcoutts, line 348: I'd try and find a different word that consistency here. In a DB context, consistency makes people think of C from ACID, which is not what we mean.

Here we mean something like how smooth, even, not spiky the I/O traffic is. In normal English "consistent" would be fine for this meaning, but not in this context.

README.md
```diff
@@ -276,81 +278,383 @@ schedule is listed as N/A.
...
+
+##### Fine-tuning: Bloom Filter Size <span id="fine_tuning_bloom_filter_size" class="anchor"></span>
+
+The configuration parameter `confBloomFilterAlloc` affects the size of
+the Bloom filters, which balances the performance of lookups against the
+in-memory size of the table.
+
+Tables maintain a [Bloom
+filter](https://en.wikipedia.org/wiki/Bloom_filter "https://en.wikipedia.org/wiki/Bloom_filter")
+in memory for each run on disk. These Bloom filter are are
```
dcoutts, line 543: filters

are are

README.md
```diff
@@ -276,81 +278,383 @@ schedule is listed as N/A.
...
+probablilistic datastructure that is used to track which keys are
```
dcoutts, line 544: are used

README.md
```diff
@@ -276,81 +278,383 @@ schedule is listed as N/A.
...
+false-positive rate.
+
+The value must be in the range $`(0, 1)`$. The recommended range is
+$`[1\mathrm{e}{ -5 },1\mathrm{e}{ -2 }]`$.
+
+The total in-memory size of all Bloom filters scales *linearly* with the
+number of physical entries in the table and is determined by the number
+of physical entries multiplied by the number of bits per physical entry,
+i.e, $`n\cdot\operatorname{BPE}`$. For example, a table with 100 million
+physical entries and an FPR of 1 in 1,000 will allocate 188MiB of memory
```
dcoutts, line 596: I concur on the 188Mb calculation.

README.md
```diff
@@ -276,81 +278,383 @@ schedule is listed as N/A.
...
+for Bloom filters. If the table uses lazy levelling, a size ratio of 4,
+and a write buffer capacity of 20,000 entries, then a spurious read can
+be expected once every 46 lookup operations.
```
dcoutts, line 599: Did we cover how this overall FPR is calculated? I missed it.

README.md
```diff
@@ -276,81 +278,383 @@ schedule is listed as N/A.
...
+  the number of levels that are cached. For a description of levels, see
+  [Merge Policy, Size Ratio, and Write Buffer
+  Size](#fine_tuning_data_layout "#fine_tuning_data_layout"). With this
+  setting, the database can be expected to cache up to $`\frac{k}{P}`$
+  pages of memory, where $`k`$ refers to the number of entries that fit
+  in levels $`[1,l]`$ and is defined as $`\sum_{i=1}^{l}BT^{i}`$.
+
+- Use the `DiskCacheNone` policy if the database's access pattern has
+  does not have good spatial or temporal locality. For instance, if the
+  access pattern is uniformly random.
```
dcoutts, line 655: This is good, but on reflection I think I didn't describe it to you quite perfectly.

I think common meaning would say that good temporal locality covers both entries that were recently inserted, or entries that were recently looked up.

`DiskCacheAll` helps with temporal locality for inserts and lookups, however `DiskCacheLevelOneTo` only helps with temporal locality for inserts. `DiskCacheLevelOneTo` does not help with temporal locality for lookups.

So unfortunately, this complicates your presentation somewhat.

lsm-tree.cabal
```diff
@@ -121,65 +124,307 @@ description:
...
+  Let us look at a table that uses the default configuration and has 100 million entries with 34 byte keys and 60 byte values.
+  The total size of 100 million key–value pairs is approximately 8.75GiB.
+  Hence, the on-disk size would be at least 8.75GiB, not counting the overhead for metadata.
 
-  The in-memory size of an LSM-tree is described in terms of the variable \(n\), which refers to the number of /physical/ database entries.
-  A /physical/ database entry is any key–operation pair, e.g., @Insert k v@ or @Delete k@, whereas a /logical/ database entry is determined by all physical entries with the same key.
+  TODO: Choose default Bloom filter size.
 
-  The worst-case in-memory size of an LSM-tree is \(O(n)\).
+  The in-memory size would be approximately 0.99GiB:
```
dcoutts, line 137: Maintaining consistency is awkward!

dcoutts (commented):
I've got part way through this.

README.md
wenkokke: Ah, good catch! That's why it looks so big.

README.md
wenkokke: It sure does!

README.md
wenkokke: I've left in a duplicate section depending on which default we settle on (1:100, 1:1000, 10b/e, 15b/e).

README.md
wenkokke: Good point.

README.md
wenkokke: It's slightly further up, where I state that it's a binomial distribution. The expected value of `B(n,p)` is simply `np` so there's not much to explain.

lsm-tree.cabal
wenkokke: I just left my converter on bytes for too long. Or do you mean "repeated text"? Then, yes. If only users only engaged with my docs in the order that I want them to.

recursion-ninja (approved):
Wow, what extensive changes! I really like the type class law and complexity annotations that were added. This is appears ready to merge to me, and need to delay and risk bit-rot.

wenkokke:
> > > This is coming together nicely 💃
> 
> > > What's your take on having duplicated text? See my comment on the haddocks for config types
> 
> > 
> 
> > My high level take is that people access docs non-linearly and therefore text duplication is necessary. It isn't easily doable to include these bits of text from a single source, so the best we can do is test consistency.
> 
> 
> 
> I'm not sure I follow why duplication would be necessary. Would hyperlinking to a single source be a bad thing?
> 
> 

Especially for short snippets, following hyperlinks disrupts the flow of reading. It also requires you have a link anchor, which I believe in Haddock means having a section header.

wenkokke:
They're such short snippets that are unlikely to be invalidated in the future, given that they only really give a high level intuition. As such, I think linking would be detrimental to the reading experience, which is why they're duplicated.

lsm-tree.cabal
```diff
@@ -121,65 +124,312 @@ description:
...
+
+  [@confWriteBufferAlloc@]
+      The /write buffer capacity/ balances the performance of lookups and updates against the in-memory size of the table.
+      If the write buffer is larger, it takes up more memory, but lookups and updates are more efficient.
+      This parameter is referred to as \(B\) in the disk I\/O cost of operations.
+      Irrespective of this parameter, the write buffer size cannot exceed 4GiB.
+
+  [@confMergeSchedule@]
+      The /merge schedule/ balances the performance of lookups and updates against the smooth performance of updates.
+      The merge schedule does not affect the performance of table unions.
```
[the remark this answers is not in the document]
wenkokke: Do you want this nuanced matter to be described in the docs?

lsm-tree.cabal
```diff
@@ -121,65 +124,312 @@ description:
...
+      If the Bloom filters are larger, they take up more memory, but lookup operations are more efficient.
+
+  [@confFencePointerIndex@]
+      The /fence-pointer index type/ supports two types of indexes.
+      The /ordinary/ indexes are designed to work with any key.
+      The /compact/ indexes are optimised for the case where the keys in the database are uniformly distributed, e.g., when the keys are hashes.
+
+  [@confDiskCachePolicy@]
+      The /disk cache policy/ determines if lookup operations use the OS page cache.
+      Caching may improve the performance of lookups if database access follows certain patterns.
```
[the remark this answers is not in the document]
wenkokke: Do you want this discussed with any nuance, or would the following change suffice?

```diff
- Caching may improve the performance of lookup if database access follows certain patterns.
+ Caching may improve the performance of lookups and updates if database access follows certain patterns.
```

lsm-tree.cabal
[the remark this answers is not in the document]
wenkokke: Does this suffice?
```diff
- The merge schedule does not affect the performance of table unions.
+ The merge schedule does not affect the way that table unions are computed. However, any table union must complete all outstanding incremental updates.
```

PR #702, opened 2025-05-02
fix: remove `Show` instance for `BlobRef`

wenkokke:
Removing the `Show` instance for `BlobRef` is painless, except for a single use in the unit tests for debugging, which I've fixed by masking out the unshowable `BlobRef` with a unit.


wenkokke:
Closing this, as it proved too inconvenient.

PR #703, opened 2025-05-02
fix: rename `MergePolicyLazyLevelling` to `LazyLevelling`

wenkokke:
For consistency with `MergeSchedule` and because we have to mention the constructor over and over in the documentation.


recursion-ninja (approved):
📤 LGTM

PR #705, opened 2025-05-05
doc: account for merge schedule in complexities

wenkokke:


lsm-tree.cabal
```diff
@@ -77,46 +77,47 @@ description:
...
+  |          | Close                  | @LazyLevelling@ | N\/A            | \(O(T \: \log_T \frac{n}{B})\)                 |
+  +----------+------------------------+-----------------+-----------------+------------------------------------------------+
+  |          | Lookup                 | @LazyLevelling@ | N\/A            | \(O(T \: \log_T \frac{n}{B})\)                 |
+  +----------+------------------------+-----------------+-----------------+------------------------------------------------+
+  |          | Range Lookup           | @LazyLevelling@ | N\/A            | \(O(T \: \log_T \frac{n}{B} + \frac{b}{P})\)*  |
+  +----------+------------------------+-----------------+-----------------+------------------------------------------------+
+  |          | Insert\/Delete\/Update | @LazyLevelling@ | @Incremental@   | \(O(\frac{1}{P} \: \log_T \frac{n}{B})\)       |
+  +----------+------------------------+-----------------+-----------------+------------------------------------------------+
+  |          |                        | @LazyLevelling@ | @OneShot@       | \(O(\frac{n}{P})\)                             |
```
jorisdral, line 101: Could you elaborate on how you got to $O(\frac{n}{P})$? It's not clear to me that the $\log_T$ part would disappear when using the `OneShot` merge schedule

src/Database/LSMTree.hs
```diff
@@ -993,10 +995,12 @@ insert table k v b =
...
 
-The worst-case disk I\/O complexity of this operation depends on the merge policy of the table:
+The worst-case disk I\/O complexity of this operation depends on the merge policy and the merge schedule of the table:
 
-['LazyLevelling']:
-    \(O(b \: \log_T \frac{n}{B})\).
+['LazyLevelling'\/'Incremental']:
+    \(O(b \: \frac{1}{P} \: \log_T \frac{n}{B})\).
+['LazyLevelling'\/'OneShot']:
+    \(O(b \log_T \frac{b}{B} + \frac{n}{P})\).
```
jorisdral, line 1003: And how did you get to $O(b \log_T \frac{b}{B} + \frac{n}{P})$ for the bulk operations?

jorisdral:
Maybe for posterity we should document how we got to these complexities. Doesn't have to be something we do now in this PR, but seems useful

src/Database/LSMTree.hs
wenkokke: The added `b*logT b/B` term is the cost of inserting the batch, which is only there because it dominates the other term when `b > n`.

jorisdral:
> > Maybe for posterity we should document how we got to these complexities. Doesn't have to be something we do now in this PR, but seems useful
> 
> Fair, but I won't be able to do that. All I've done is interrogate you, @mheinzel, and @dcoutts, and write down the consensus.

I can take a stab it at some point

> The stated motivations are unlikely to be more informative than the complexities themselves, since they tend to be of the form "worst-case, this needs to do something for every run", which is more or less what the complexity bound expresses.

That's true for some, but not everything is trivial

wenkokke:
@jorisdral Are you happy to merge this?

jorisdral (approved):

PR #709, opened 2025-05-07
In the prototype, merge the union level into regular levels

mheinzel:
# Description

This leads to the union level being merged with runs from the regular levels at some point.

### DIFF
--- a/prototypes/ScheduledMerges.hs
+++ b/prototypes/ScheduledMerges.hs
@@ -65,6 +65,7 @@ module ScheduledMerges (
     MergeDebt(..),
     NominalCredit(..),
     NominalDebt(..),
+    maxBufferSize,
     Run,
     runSize,
     UnionCredits (..),
@@ -85,6 +86,7 @@ import           Prelude hiding (lookup)
 
 import           Data.Bits
 import           Data.Foldable (for_, toList, traverse_)
+import           Data.Functor ((<&>))
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import           Data.Maybe (catMaybes)
@@ -344,19 +346,25 @@ invariant (LSMContent _ levels ul) = do
     expectedRunLengths :: Int -> [Run] -> [Level s] -> ST s ()
     expectedRunLengths ln rs ls =
       case mergePolicyForLevel ln ls ul of
-        -- Levels using levelling have only one (incoming) run, which almost
-        -- always consists of an ongoing merge. The exception is when a
-        -- levelling run becomes too large and is promoted, in that case
-        -- initially there's no merge, but it is still represented as an
-        -- 'IncomingRun', using 'Single'. Thus there are no other resident runs.
-        MergePolicyLevelling -> assertST $ null rs
-        -- Runs in tiering levels usually fit that size, but they can be one
-        -- larger, if a run has been held back (creating a 5-way merge).
-        MergePolicyTiering   -> assertST $ all (\r -> tieringRunSizeToLevel r `elem` [ln, ln+1]) rs
-        -- (This is actually still not really true, but will hold in practice.
-        -- In the pathological case, all runs passed to the next level can be
-        -- factor (5/4) too large, and there the same holding back can lead to
-        -- factor (6/4) etc., until at level 12 a run is two levels too large.
+        MergePolicyLevelling ->
+          -- Levels using levelling have only one (incoming) run, which almost
+          -- always consists of an ongoing merge. The exception is when a
+          -- levelling run becomes too large and is promoted, in that case
+          -- initially there's no merge, but it is still represented as an
+          -- 'IncomingRun', using 'Single'. Thus there are no other resident
+          -- runs.
+          assertST $ null rs
+        MergePolicyTiering -> do
+          -- Runs in tiering levels usually fit that size, but they can be one
+          -- larger, if a run has been held back (creating a 5-way merge).
+          --
+          -- TODO: This is actually still not really true, but will hold in
+          -- practice. In the pathological case, all runs passed to the next
+          -- level can be factor (5/4) too large, and there the same holding
+          -- back can lead to factor (6/4) etc., until at level 12 a run is two
+          -- levels too large.
+          assertST $ all (\r -> runSize r > 0) rs
+          assertST $ all (\r -> tieringRunSizeToLevel r `elem` [ln, ln+1]) rs
 
     -- Incoming runs being merged also need to be of the right size, but the
     -- conditions are more complicated.
@@ -367,11 +375,12 @@ invariant (LSMContent _ levels ul) = do
         MergePolicyLevelling -> do
           case (ir, mrs) of
             -- A single incoming run (which thus didn't need merging) must be
-            -- of the expected size range already
+            -- of the expected size range already, but it could also be smaller
+            -- if it comes from a union level.
             (Single r, m) -> do
               assertST $ case m of CompletedMerge{} -> True
                                    OngoingMerge{}   -> False
-              assertST $ levellingRunSizeToLevel r == ln
+              assertST $ levellingRunSizeToLevel r <= ln
 
             -- A completed merge for levelling can be of almost any size at all!
             -- It can be smaller, due to deletions in the last level. But it
@@ -495,6 +504,11 @@ isCompletedMergingTree (MergingTree ref) = do
       CompletedTreeMerge r -> return r
       OngoingTreeMerge mr  -> isCompletedMergingRun mr
       PendingTreeMerge _   -> failI $ "not completed: PendingTreeMerge"
+
+getCompletedMergingTree :: MergingTree s -> ST s (Maybe Run)
+getCompletedMergingTree t =
+    either (const Nothing) Just
+      <$> evalInvariant (isCompletedMergingTree t)
 
 type Invariant s = E.ExceptT String (ST s)
 
@@ -781,8 +795,11 @@ update tr (LSMHandle scr lsmr) k op = do
     let wb' = Map.insertWith combine k op wb
     if bufferSize wb' >= maxBufferSize
       then do
-        ls' <- increment tr sc (bufferToRun wb') ls unionLevel
-        let content' = LSMContent Map.empty ls' unionLevel
+        -- Before adding the run to the regular levels, we check if we can get
+        -- rid of the union level (by moving it into into the regular ones).
+        (ls', ul') <- migrateUnionLevel tr sc ls unionLevel
+        ls'' <- increment tr sc (bufferToRun wb') ls' ul'
+        let content' = LSMContent Map.empty ls'' ul'
         invariant content'
         writeSTRef lsmr content'
       else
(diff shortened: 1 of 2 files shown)


prototypes/ScheduledMerges.hs
```diff
@@ -367,11 +375,12 @@ invariant (LSMContent _ levels ul) = do
...
           case (ir, mrs) of
             -- A single incoming run (which thus didn't need merging) must be
-            -- of the expected size range already
+            -- of the expected size range already, but it could also be smaller
+            -- if it comes from a union level.
             (Single r, m) -> do
               assertST $ case m of CompletedMerge{} -> True
                                    OngoingMerge{}   -> False
-              assertST $ levellingRunSizeToLevel r == ln
+              assertST $ levellingRunSizeToLevel r <= ln
```
jorisdral, line 383: We can make the check a little bit stricter if we distinguish between a `Single` run and a `MigratedUnion` run. In the former case we can then check equality, and for the latter case check inequality.

It might be nice to make that distinction anyway. It does not come at much of a cost, I think?

prototypes/ScheduledMerges.hs
```diff
@@ -781,8 +795,11 @@ update tr (LSMHandle scr lsmr) k op = do
     let wb' = Map.insertWith combine k op wb
     if bufferSize wb' >= maxBufferSize
       then do
-        ls' <- increment tr sc (bufferToRun wb') ls unionLevel
-        let content' = LSMContent Map.empty ls' unionLevel
+        -- Before adding the run to the regular levels, we check if we can get
+        -- rid of the union level (by moving it into into the regular ones).
```
jorisdral, line 799: Okay, so this is the approach where we try to put the completed union run into the levels whenever we flush a write buffer. What were the other alternatives? Could this not be implemented as part of `supplyUnionCredits`?

prototypes/ScheduledMerges.hs
```diff
@@ -1158,9 +1175,44 @@ depositNominalCredit (NominalDebt nominalDebt)
...
 --
 
+-- | At some point, we want to merge the union level with the regular levels.
+-- We achieve this by moving it into a new last regular level once it is both
+-- completed (merged down to a single run) and fits into such a new level.
+--
+-- Our representation doesn't allow for empty levels, so we can only put the
+-- run directly after the pre-existing regular levels. If it is too large for
+-- that, we don't want to move it yet to avoid violating run size invariants
+-- and doing inefficient merges of runs with very different sizes.
```
jorisdral, line 1185: Maybe we should make a TODO to allow empty levels? Or maybe having the `Single` vs. `MigratedUnion` distinction would help with thiss?

prototypes/ScheduledMerges.hs
```diff
@@ -1158,9 +1175,44 @@ depositNominalCredit (NominalDebt nominalDebt)
...
+migrateUnionLevel :: forall s. Tracer (ST s) Event
+                  -> Counter -> Levels s -> UnionLevel s
+                  -> ST s (Levels s, UnionLevel s)
+migrateUnionLevel _ _ ls NoUnion = do
+    -- nothing to do
+    return (ls, NoUnion)
+migrateUnionLevel _tr _sc ls ul@(Union t _) =
+    -- TODO: tracing
```
jorisdral, line 1193: Is this something you still want to do in this PR?

prototypes/ScheduledMerges.hs
```diff
@@ -1158,9 +1175,44 @@ depositNominalCredit (NominalDebt nominalDebt)
...
+-- run directly after the pre-existing regular levels. If it is too large for
+-- that, we don't want to move it yet to avoid violating run size invariants
+-- and doing inefficient merges of runs with very different sizes.
+migrateUnionLevel :: forall s. Tracer (ST s) Event
+                  -> Counter -> Levels s -> UnionLevel s
+                  -> ST s (Levels s, UnionLevel s)
+migrateUnionLevel _ _ ls NoUnion = do
+    -- nothing to do
+    return (ls, NoUnion)
+migrateUnionLevel _tr _sc ls ul@(Union t _) =
```
jorisdral, line 1192: Seems like the `Counter` argument is unused in both definitions

prototypes/ScheduledMerges.hs
```diff
@@ -1158,9 +1175,44 @@ depositNominalCredit (NominalDebt nominalDebt)
...
+    -- TODO: tracing
+    getCompletedMergingTree t <&> \case
+      Just r
+        | null r ->
+          -- If the union level is empty, we can just drop it.
+          (ls, NoUnion)
```
jorisdral, line 1198: :+1: 

Ideally, we would try to enforce non-empty runs generally, but that's a separate issue

prototypes/ScheduledMerges.hs
```diff
@@ -1158,9 +1175,44 @@ depositNominalCredit (NominalDebt nominalDebt)
...
+        | levellingRunSizeToLevel r <= length ls + 1 ->
+          -- If it fits into a hypothetical new last level, put it there.
+          --
+          -- TODO: In some cases it seems desirable to even add it to the
+          -- existing last regular level (so it becomes part of a merge
+          -- sooner), but that would lead to additional merging work that was
+          -- not accounted for. We'd need to be careful to ensure the merge
+          -- completes in time, without doing a lot of work in a short time.
+          (ls ++ [Level (Single r) []], NoUnion)
```
jorisdral, line 1207: A potential problem here is that we could build up too many levels. If all we do is union tables, and never perform any regular updates, then there is no guarantee that the finished union levels ever get merged together. We'd just add new levels indefinitely.

prototypes/ScheduledMergesTest.hs
```diff
@@ -176,29 +177,72 @@ test_merge_again_with_incoming =
 -- properties
 --
 
+-- TODO: also generate nested unions?
```
jorisdral, line 180: :+1: nesting at least once would potentially show some edge case behaviour

prototypes/ScheduledMergesTest.hs
```diff
@@ -176,29 +177,72 @@ test_merge_again_with_incoming =
...
-        _ <- LSM.supplyUnionCredits t (UnionCredits x)
+        leftovers <- LSM.supplyUnionCredits t (UnionCredits x)
         debt' <- LSM.remainingUnionDebt t
 
         rep <- dumpRepresentation t
         return $ QC.counterexample (show (debt, debt')) $ QC.conjoin
-          [ debt =/= UnionDebt 0
-          , debt' === UnionDebt 0
+          [ QC.counterexample "debt" $ debt =/= UnionDebt 0
+          , QC.counterexample "debt'" $ debt' === UnionDebt 0
```
jorisdral, line 197: Maybe `"debt a priori"` and `"debt a posteriori"`?

prototypes/ScheduledMergesTest.hs
```diff
@@ -176,29 +177,72 @@ test_merge_again_with_incoming =
...
   where
     isCompleted = \case
         MLeaf{} -> True
         MNode{} -> False
 
+-- | The union level will get merged into the last regular level once the union
+-- merge is completed and sufficient new entries have been inserted.
+prop_union_merge_into_levels :: [[(LSM.Key, LSM.Op)]] -> Property
+prop_union_merge_into_levels kopss = length (filter (not . null) kopss) > 1 QC.==>
+    QC.forAll arbitrary $ \firstPay ->
```
jorisdral, line 210: It may be intentional, but `QC.forAll` does not shrink, you'd have to use `QC.forAllShrink`

prototypes/ScheduledMergesTest.hs
```diff
@@ -176,29 +177,72 @@ test_merge_again_with_incoming =
...
+      stToIO $ do
+        ts <- traverse (mkTable tr) kopss
+        t <- LSM.unions ts
+
+        -- pay off the union and insert enough that it fits into
+        -- the last level
+        let payOffDebt = do
+              UnionDebt d <- LSM.remainingUnionDebt t
+              _ <- LSM.supplyUnionCredits t (UnionCredits d)
+              return ()
```
jorisdral, line 221: This is not performing inserts, but the comment says it does

prototypes/ScheduledMergesTest.hs
```diff
@@ -176,29 +177,72 @@ test_merge_again_with_incoming =
...
+
+        -- insert as many new entries as there are in the completed
+        -- union level
+        let fillTable = do
+              unionRunSize <- length <$> LSM.logicalValue t
+              LSM.inserts tr t
+                [(K k, V 0, Nothing) | k <- [1 .. unionRunSize]]
```
jorisdral, line 228: What do we need these inserts for?

prototypes/ScheduledMerges.hs
mheinzel: It requires an extra case in a few pattern matches, but that's not really an issue. Stricter invariants are probably worth it.

prototypes/ScheduledMerges.hs
mheinzel: I just realised I originally wanted to do something even simpler: When the union is completed, always move it to a new level. If it is much larger than that level should be, the existing code will already handle, not creating a merge with it, but just pushing the oversized run down the levels over time, until it fits in and becomes part of a new last level merge. I think combined with the `MigratedUnion` constructor (to avoid watering down the invariant too much), that could be a decent solution. Kind of like allowing empty levels just before a `MigratedUnion`, but not explicitly representing them.

prototypes/ScheduledMerges.hs
mheinzel: I'm not sure what you mean. When creating a `union`, initially everything in there is part of the union level, where it will already be merged together. One the resulting run is moved to the regular levels, that table will never have a union level again.

prototypes/ScheduledMergesTest.hs
mheinzel: To make sure the completed union fits into the last level. Otherwise it currently won't be migrated, although that could change with https://github.com/IntersectMBO/lsm-tree/pull/709#discussion_r2077543468.

prototypes/ScheduledMerges.hs
jorisdral: Ah, yes, now I remember. So one such situation would be where we create an incremental union table, duplicate it, and then only supply credits to one of them. Would it work if the other duplicate (which did not supply credits) always has 1 credit remaining for the "moving back into the levels'?

prototypes/ScheduledMerges.hs
jorisdral: If so, maybe we should have a dedicated test that triggers this particular behaviour so that we can check that it works correctly

prototypes/ScheduledMerges.hs
jorisdral: Hmm yeah maybe I just got confused, makes sense

prototypes/ScheduledMerges.hs
mheinzel: Yes, could be due to a duplicate or also using the table as an input to another union.

> Would it work if the other duplicate (which did not supply credits) always has 1 credit remaining for the "moving back into the levels'?

That would help make it clear that `supplyUnionCredits` still *should* be called, but it still doesn't guarantee it.

prototypes/ScheduledMerges.hs
```diff
@@ -344,19 +357,25 @@ invariant (LSMContent _ levels ul) = do
...
+        MergePolicyTiering -> do
+          -- Runs in tiering levels usually fit that size, but they can be one
+          -- larger, if a run has been held back (creating a 5-way merge).
+          --
+          -- TODO: This is actually still not really true, but will hold in
+          -- practice. In the pathological case, all runs passed to the next
+          -- level can be factor (5/4) too large, and there the same holding
+          -- back can lead to factor (6/4) etc., until at level 12 a run is two
+          -- levels too large.
+          assertST $ all (\r -> runSize r > 0) rs
```
jorisdral, line 377: This is a new assertion. Just there for convenience?

prototypes/ScheduledMerges.hs
```diff
@@ -367,11 +386,14 @@ invariant (LSMContent _ levels ul) = do
         MergePolicyLevelling -> do
           case (ir, mrs) of
             -- A single incoming run (which thus didn't need merging) must be
-            -- of the expected size range already
-            (Single r, m) -> do
+            -- of the expected size range already, but it could also be smaller
+            -- if it comes from a union level.
```
jorisdral, line 390: And could it not also be larger if it came form a union level?

prototypes/ScheduledMerges.hs
```diff
@@ -1158,9 +1212,44 @@ depositNominalCredit (NominalDebt nominalDebt)
...
+          --
+          -- TODO: In some cases it seems desirable to even add it to the
+          -- existing last regular level (so it becomes part of a merge
+          -- sooner), but that would lead to additional merging work that was
+          -- not accounted for. We'd need to be careful to ensure the merge
+          -- completes in time, without doing a lot of work in a short time.
+          (ls ++ [Level (Single MigratedUnion r) []], NoUnion)
+      _ ->
+        -- Otherwise, just leave it for now.
+        (ls, ul)
```
jorisdral, line 1247: Like we discussed elsewhere, it might still be useful to migrate the union level even if it doesn't fit

prototypes/ScheduledMerges.hs
jorisdral: It feels like if we rely on `updates` for the union level to be migrated, that it is also not fully guaranteed that it will migrate, or is it?

PR #710, opened 2025-05-07
rename: improve consistency between full API and test class

wenkokke:


jorisdral (approved):
LGTM!

` git ls-files | xargs grep -i -E "mupdate|mupsert|monoidal update"` tells me there are still some 135 lines with outdated terminology. We should probably fix those at some point too

PR #711, opened 2025-05-08
Rework golden tests

jorisdral:
Previously, we were encoding each snapshot type in the context of a full
metadata type. However, if one field somewhere down in the snapshot metadata
changes, then the encoding of the full metadata stays the same except for that
one field further down. As such, this commit changes the test setup so that we
now encode (and golden-test) each type involved in the snapshot metadata
separately.

We also add a new `EnumGolden` class for enumerating values of snapshot types.
With the more direct tests and `EnumGolden`, it's easier to prevent
combinatorial explosion.


test/Test/Database/LSMTree/Internal/Snapshot/Codec/Golden.hs
```diff
@@ -29,328 +30,453 @@ import           System.FS.API.Types (FsPath, MountPoint (..), fsToFilePath,
...
+spaceToUnderscore :: Char -> Char
+spaceToUnderscore ' ' = '_'
+spaceToUnderscore c   = c
+
+{-------------------------------------------------------------------------------
+  Enumeration class: instances
+-------------------------------------------------------------------------------}
+
+instance EnumGolden SnapshotMetaData where
+  singGolden = SnapshotMetaData singGolden singGolden singGolden singGolden singGolden
```
jorisdral, line 264: This pattern of applying a constructor to a string of `singGoldens` looks like it could be generically derived. Maybe I should add a TODO

test/Test/Database/LSMTree/Internal/Snapshot/Codec/Golden.hs
```diff
@@ -28,329 +30,490 @@ import qualified System.FS.API as FS
...
+
+enumGoldenAnnotated' :: EnumGolden a => Proxy a -> [(Annotation, a)]
+enumGoldenAnnotated' _ = enumGoldenAnnotated
+
+{-------------------------------------------------------------------------------
+  Enumeration class: names and file paths
+-------------------------------------------------------------------------------}
+
+nameGolden :: Typeable a => Proxy a -> Annotation -> String
+nameGolden p ann = map spaceToUnderscore (show $ typeRep p) ++ "." ++ ann
```
jorisdral, line 244: The downside of using `typeRep` is that the name of the golden file/test changes when the type name changes. In some cases, we might want to rename the type without affecting the name, so maybe `nameGolden` should be a member of `EnumGolden`?

jorisdral: The `typeRep` approach can still be the default, but the upside is that you can change the name per instance

wenkokke: Would this be an issue? Seems easier to just recreate the relevant golden files than to maintain a type-name mapping forever.

jorisdral: Maybe not, it's just something that came to mind. I'm fine with leaving it as is too

dcoutts (approved):
Looks great. Much more systematic.

Issue #445, opened 2024-10-29
Test that the *real* merge algorithm matches the prototype


jorisdral:
There is currently no established correspondence between the `ScheduledMerges` prototype and `MergeSchedule` implementation. Preferably, we should trace enough information in both the prototype and implementation to check that the tree progression matches up. 

Some aspects to check would be that the prototype and implementation:
* Have the same tree structure
* Have the same number of total supplied credits per level
* Have the same number of spent credits per level

PR #716, opened 2025-05-08
`ScheduledMerges`: make the write buffer size configurable

jorisdral:
We'd need to configure the write buffer size in the `ScheduledMerges` prototype if we want to compare traces for the prototype and real implementation (see #445).

This PR also changes the run size calculations to be somewhat closer to the calculations in the real implementation.

This PR builds on top of #715 to reduce the number of merge conflicts


jorisdral:
I've put the PR back into draft mode because I found a bug

dcoutts (approved):
Looks reasonable.

PR #717, opened 2025-05-08
`ScheduledMerges`: make the size ratio configurable

jorisdral:
It's probably also a prerequisite to prototype the merge schedule with size ratios other than `4` before we can be confident that we can make it configurable in the real implementation too. 

At that point, we'd need to configure the size ratio in the `ScheduledMerges` prototype if we want to compare traces for the prototype and real implementation (see #445). 

This PR builds on top of #716 to reduce the number of merge conflicts


jorisdral:
Ready for review again!

dcoutts (approved):
Looks good!

PR #721, opened 2025-05-13
Public documentation about ACID properties

jorisdral:
I had some text lying around from a while ago about ACID properties, so I thought it'd be useful to include it in the public API


dcoutts (approved):
LGTM

PR #722, opened 2025-05-14
Make `lint-actionlint.sh` also consider `*.yml` files

jeltsch:
Previously, `scripts/lint-actionlint.sh` only considered files whose paths match the pattern `.github/workflows/*.yaml`. However, the only workflow file that we currently have has the path `.github/workflows/ci.yml`, which doesn’t match the above-mentioned pattern. This pull request extends `scripts/lint-actionlint.sh` to also consider files whose paths match the pattern `.github/workflows/*.yml`.


wenkokke:
You may want to add the sections as well, if that's something actionlint supports.

jeltsch:
> You may want to add the sections as well, if that's something actionlint supports.

What sections do you mean?

By the way, would it make sense to have `actionlint` also cover the files `actions/*/action.yaml`, or are these not supported by `actionlint`? Sorry, I have no idea what `actionlint` precisely does. 😅

wenkokke:
I meant actions, but autocorrect undermined me.

jeltsch:
> I meant actions, but autocorrect undermined me.

Alright: So we talked about the same thing then. I’ll check tomorrow what happens if I add the action files.

jeltsch:
> I’ll check tomorrow what happens if I add the action files.

I’m back at this now but will tackle it in a new pull request.

PR #723, opened 2025-05-15
bloom filter bulk inserts and queries

dcoutts:
Two changes, both from bulk operations for bloom filters.

For Bloom filter inserts in run accumulation: instead of inserting keys into the bloom filter one by one as they are added to the run accumulator, save them up and add them all in one go when the page is being finalised. This then lets us use a bloom filter bulk insert, which lets us use memory prefetching.

For Bloom filter queries in key lookups, update the existing bulk query code to properly take advantage of the new API and use prefetching. We can now also simplify and use a single high performance implementation, rather than needing two (a more compatible one and a faster one that relied on fancier features available in later GHC versions).

Results for the WP8 benchmark (100M elements, 10bits per key, full caching, 10k batches of 256 keys):
* baseline: 88,871 ops/sec
* with bulk inserts: 97,152 ops/sec, ~9% improvement
* with bulk queries: 103,005 ops/sec, ~6% additional, ~16% cumulative improvement

So overall about a 16% improvement in ops/sec on the primary WP8 benchmark, and as a bonus, getting over the magic 100k ops/sec threshold (on my laptop).


src/Database/LSMTree/Internal/RunAcc.hs
```diff
@@ -265,7 +264,6 @@ addLargeSerialisedKeyOp racc@RunAcc{..} k page overflowPages =
   assert (RawPage.rawPageOverflowPages page > 0) $
   assert (RawPage.rawPageOverflowPages page == length overflowPages) $ do
     modifyPrimVar entryCount (+1)
-    Bloom.insert mbloom k
```
jorisdral, line 268: This one should probably remain, because large entries bypass the page accumulator, and therefore `flushPageIfNonEmpty` is bypassed, and `bloomInserts` is bypassed

src/Database/LSMTree/Internal/RunAcc.hs
```diff
@@ -298,6 +298,12 @@ flushPageIfNonEmpty RunAcc{mpageacc, mindex} = do
 
       else pure Nothing
 
+{-# NOINLINE bloomInserts #-}
```
jorisdral, line 301: Why `NOINLINE`?

test/Test/Database/LSMTree/Internal/BloomFilter.hs
```diff
@@ -205,3 +207,18 @@ instance Arbitrary Bloom2.CandidateProbe where
...
+    bloom_insert =
+      BF.create (BF.sizeForFPR fpr n) $ \mb ->
+        mapM_ (BF.insert mb) keys
+
+    bloom_insertMany =
+      BF.create (BF.sizeForFPR fpr n) $ \mb ->
+        Bloom.bloomInserts mb (\k -> pure $! keysv V.! k) n
+
+    !keysv = V.fromList keys
+    !n     = V.length keysv
```
jorisdral, line 224: It might be more fitting to put this test into the bloomfilter test suit instead of the lsm-tree test suite

jorisdral (changes requested):
Some initial comments

src/Database/LSMTree/Internal/RunAcc.hs
```diff
@@ -211,7 +211,6 @@ addLargeKeyOp ::
 addLargeKeyOp racc@RunAcc{..} k e =
   assert (not (PageAcc.entryWouldFitInPage k e)) $ do
     modifyPrimVar entryCount (+1)
-    Bloom.insert mbloom k
```
jorisdral, line 214: Oh, and this one too (see https://github.com/IntersectMBO/lsm-tree/pull/723/files#r2090655471)

src/Database/LSMTree/Internal/RunAcc.hs
dcoutts: Initially it was just so I could see the generated core with a known name (rather than inlined at the use site). It doesn't otherwise hurt or benefit.

test/Test/Database/LSMTree/Internal/BloomFilter.hs
dcoutts: If so, we'd want to move the bulk insert function into the `bloomfilter` lib too.

src/Database/LSMTree/Internal/RunAcc.hs
dcoutts: Right.

bloomfilter/src/Data/BloomFilter/Blocked/BitArray.hs
```diff
@@ -34,7 +34,7 @@ import           Data.Primitive.PrimArray
 import           Data.Word (Word64, Word8)
 
 import           GHC.Exts (Int (I#), prefetchByteArray0#,
-                     prefetchMutableByteArray3#)
+                     prefetchMutableByteArray0#)
```
jorisdral, line 37: What does prefetch 0 vs. prefetch 3 mean again? I read up on `__builtin_prefetch`, which seemed to be the inspiration for the primop in GHC, and there it means temporal locality. Does temporal locality relate directly to cache levels?

src/Database/LSMTree/Internal/BloomFilter.hs
```diff
@@ -125,3 +129,57 @@ hGetByteArrayExactly hfs h len = do
...
+{-# INLINE bloomInserts #-}
+bloomInserts ::
+     forall a s.
+     BF.Hashable a
+  => BF.MBloom s a
+  -> (Int -> ST s a)
+  -> Int
+  -> ST s ()
+bloomInserts bloom key n =
+    P.newPrimArray 0x10 >>= body
```
jorisdral, line 144: So prefetch distance 15 or 16 depending on how you define prefetch distance, right?

src/Database/LSMTree/Internal/BloomFilter.hs
```diff
@@ -125,3 +129,57 @@ hGetByteArrayExactly hfs h len = do
...
+              insertProbe
+                (i+1)
+                ((i_r + 1) .&. 0x0f)
+                ((i_w + 1) .&. 0x0f)
+
+          | ((i_r + 1) .&. 0x0f) /= i_w =
+              insertProbe
+                i
+                ((i_r + 1) .&. 0x0f)
+                i_w
```
jorisdral, line 183: IIUC, this case kicks in when there are no more keys to prefetch, so we just have to do all the remaining probe inserts until we hit the end of the rolling buffer (`i_r == i_w`)

src/Database/LSMTree/Internal/BloomFilter.hs
```diff
@@ -125,3 +129,57 @@ hGetByteArrayExactly hfs h len = do
...
+
+{-# INLINE bloomInserts #-}
+bloomInserts ::
+     forall a s.
+     BF.Hashable a
+  => BF.MBloom s a
+  -> (Int -> ST s a)
+  -> Int
+  -> ST s ()
+bloomInserts bloom key n =
```
jorisdral, line 143: I understand how it achieves faster bulk inserts using prefetching because I have seen it before, but maybe a small (internal) note in a comment might make it clear to future readers. Maybe something along the lines of "prefetch blocks of the bit array before inserting into them using a rolling buffer"

test/Test/Database/LSMTree/Internal/BloomFilter.hs
jorisdral: Oops, yes, true, I was under the impression that it was already in the `bloomfilter` lib. I don't have a preference, so do what you see fit to do

jorisdral (approved):
LGTM! If we fix the test failure that I noted in my previous review, then we're good to go

dcoutts:
CI failure due to flaky test fixed in #724.

bloomfilter/src/Data/BloomFilter/Blocked/BitArray.hs
dcoutts: Values 1--3 mean how many levels of cache to load it into: 1 level (L3 only), 2 levels (L3+L2), 3 levels (L3+L2+L1).

Value 0 means, as you say, means prefetch with low temporal locality. This means read it into some level of cache (unspecified which ones) but mark it with a hint that it will not be needed for long. In the CPU I think this means that the entry is marked so that it is the first to be evicted when that cache entry is next needed, so will not hang around for long or compete with other data that needs to be kept in the cache.

src/Database/LSMTree/Internal/BloomFilter.hs
dcoutts: Exactly.

src/Database/LSMTree/Internal/BloomFilter.hs
dcoutts: Yes exactly. I should probably add a comment to that effect.

src/Database/LSMTree/Internal/BloomFilter.hs
dcoutts: Yeah.

test/Test/Database/LSMTree/Internal/BloomFilter.hs
dcoutts: I was also thinking I might move it.

The only thing is, then we'd want a corresponding multi-lookup. But the lookup we want is for N * M keys and filters, which would not reuse a multi-lookup for many keys in one filter, since we want a sparse result array.

src/Database/LSMTree/Internal/BloomFilter.hs
dcoutts: Documented.

src/Database/LSMTree/Internal/BloomFilter.hs
dcoutts: Documented.

src/Database/LSMTree/Internal/BloomFilter.hs
dcoutts: Added a comment on the general strategy.
```
    -- The general strategy is to use a rolling buffer @buf@ (of size 16). At
    -- the write end of the buffer, we prepare the probe locations and prefetch
    -- the corresponding cache line. At the read end, we do the hash insert.
    -- By having a prefetch distance of 15 between the write and read ends, we
    -- can have up to 15 memory reads in flight at once, thus improving
    -- utilisation of the memory bandwidth.
```

bloomfilter/src/Data/BloomFilter/Blocked/BitArray.hs
dcoutts: Documented.

src/Database/LSMTree/Internal/RunAcc.hs
dcoutts: Done.

src/Database/LSMTree/Internal/BloomFilterQuery1.hs
```diff
@@ -84,47 +89,57 @@ bloomQueries ::
...
+          | otherwise =
+              loop2 res2 resix2 (kix+1)
+
+        loop2_prefetch :: KeyIx -> ST s ()
+        loop2_prefetch !kix
+          | kix == ksN = pure ()
+          | otherwise  = do
+              let !keyhash = P.indexPrimArray keyhashes kix
+              Bloom.prefetchElem filter keyhash
+              loop2_prefetch (kix+1)
```
jorisdral: The commit message for 8eaf7cef9b5ad894a8aa981c6440d323ff8a3d3c says that the prefetch distance is the number of runs, but this looks like the prefetch distance is the number of keys

bench/macro/lsm-tree-bench-bloomfilter.hs
```diff
@@ -108,7 +108,7 @@ benchmarks = do
       benchmark "bloomQueries1"
```
jorisdral, line 108:
```suggestion
      benchmark "bloomQueries"
```

jorisdral (approved):
LGTM with the new bulk query changes

bloomfilter/src/Data/BloomFilter/Blocked.hs
```diff
@@ -190,3 +193,87 @@ deserialise bloomsize fill = do
...
+--
+-- This is somewhat faster than repeated insertion using 'insert'. It uses
+-- memory prefetching to improve the utilisation of memory bandwidth. This has
+-- greatest benefit for large filters (that do not fit in L3 cache) and for
+-- inserting many elements, e.g. > 10.
+--
+-- To get best performance, you probably want to specialise this function to
+-- the 'Hashable' instance and the lookup action. It is marked @INLINE@ to help
+-- with this. But since it is marked inline, avoid calling it in many locations.
+-- Define the specialisations you need and use those.
```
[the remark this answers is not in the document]
dcoutts: GHC's notion of specialise is only for type classes. Here we want to specialise for both a type class (Hashable) and the lookup function. So the user needs to do something themselves if they want to specialise and share the specialised code.

But yes it's fine to make INLINABLE. I'll do that.

src/Database/LSMTree/Internal/BloomFilterQuery1.hs
dcoutts: Yes, oops. I kept changing my mind. I did both versions and benchmarked them.

src/Database/LSMTree/Internal/BloomFilterQuery1.hs
```diff
@@ -84,47 +89,57 @@ bloomQueries ::
...
-      | otherwise
-      = do
-          (res1', resix1') <- loop2 res1 resix1 0 (blooms `V.unsafeIndex` rix)
-          loop1 res1' resix1' (rix+1)
+      -> ST s (P.MutablePrimArray s RunIxKeyIx)
+    loop1 !res !resix !rix | rix == rsN = P.resizeMutablePrimArray res resix
+    loop1 !res !resix !rix = do
+        loop2_prefetch 0
+        (res', resix') <- loop2 res resix 0
+        loop1 res' resix' (rix+1)
```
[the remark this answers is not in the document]
dcoutts: Indeed, it's a lot less sophisticated. I also did a rolling buffer version, but it was no faster (and relied on the fancy data-elevator stuff). So in the end I use just one version, with a little bit of CPP to use the strict array when available.

bench/macro/lsm-tree-bench-bloomfilter.hs
dcoutts: fixed

src/Database/LSMTree/Internal/BloomFilterQuery1.hs
dcoutts: Updated.

So in the benchmarks I did, nesting the loops this way round was slightly faster. But arguably, the other way round should have more stable prefetching behaviour, since it would depend on number of runs and not the lookup batch size.

Switching it round is not that hard. I could still do that. Thoughts?

dcoutts: I've switched it round.

test/Test/Database/LSMTree/Internal/Lookup.hs
```diff
@@ -150,8 +150,8 @@ prop_bloomQueriesModel dats =
 bloomQueriesModel :: [Set SerialisedKey] -> [SerialisedKey] -> [(RunIx, KeyIx)]
 bloomQueriesModel blooms ks = [
       (rix, kix)
-    | (rix, b) <- rs'
-    , (kix, k) <- ks'
+    | (kix, k) <- ks'
+    , (rix, b) <- rs'
     , Set.member k b
     ]
```
jorisdral, line 156: The order has to be changed in `prepLookupsModel` as well

jorisdral (approved):

src/Database/LSMTree/Internal/BloomFilter.hs
```diff
@@ -105,60 +111,97 @@ bloomQueries ::
...
+bloomQueries_loop2 !filters !keyhash !kix = go
+  where
+    go res2 resix2 rix
+#ifdef HAVE_STRICT_ARRAY
+      | rix == P.sizeofStrictArray filters = pure (res2, resix2)
+      | let !filter = P.indexStrictArray filters rix
+#else
+      | rix == V.length filters = pure (res2, resix2)
+      | let !filter = V.unsafeIndex filters rix
+#endif
```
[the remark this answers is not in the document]
dcoutts: Good idea.

test/Test/Database/LSMTree/Internal/Lookup.hs
dcoutts: Done.

dcoutts:
If CI passes now I'll squash the fixes and merge.

dcoutts (commented):
Ok, this looks fine now. All squashed.

PR #725, opened 2025-05-15
doc: fix disk I/O complexity of several operators

wenkokke:


jeltsch:
Since I’m not that familiar with these things, I can’t be very sure that everything is correct, but based on my knowledge I didn’t find any errors. Also, the prose is much clearer now.

jeltsch (approved):
Since I’m not that familiar with these things, I can’t be very sure that everything is correct, but based on my knowledge I didn’t find any errors. Also, the prose is much clearer now.

src/Database/LSMTree.hs
```diff
@@ -531,7 +531,10 @@ closeSession (Session session) =
 {- |
 Run an action with access to an empty table.
 
-The worst-case disk I\/O complexity of this operation is \(O(1)\).
+The worst-case disk I\/O complexity of this operation depends on the merge policy of the table:
+
+['LazyLevelling']:
+    \(O(T \log_T \frac{n}{B})\).
```
mheinzel, line 537: The `n` can be slightly confusing here. Maybe it would help to mention that it refers to the size of the table at the end of the action that is performed on the empty table?

mheinzel (approved):

PR #726, opened 2025-05-19
fix: rename DiskCacheLevelsAtOrBelow to DiskCacheLevelOneTo

wenkokke:


src/Database/LSMTree/Internal/Config.hs
```diff
@@ -271,7 +271,7 @@ data DiskCachePolicy =
        --
        -- Use this policy if the expected access pattern for the table
        -- has good temporal locality for recently inserted keys.
-     | DiskCacheLevelsAtOrBelow !Int
+     | DiskCacheLevelZeroTo !Int
```
jorisdral, line 274: The comment still says "at or below"

jorisdral (approved):

src/Database/LSMTree/Internal/Config.hs
wenkokke: Fair enough. I didn't bother because I have already completely rewritten that comment on my own branch.

PR #728, opened 2025-05-19
Improve the choice of files to be checked by `actionlint`

jeltsch:
This pull request changes the choice of files to be checked by `actionlint`. In particular, it includes the action files and excludes any files `.github/workflows/*.yaml`, as we don’t use the `yaml` file name extension under `.github` anymore.


jeltsch:
Apparently, `actionlint` doesn’t work for action files. Therefore, I’ve excluded them in b3927304929ebfd4ddce1dfb0759c2182eb0c9ff.

wenkokke (approved):

PR #729, opened 2025-05-21
`StateMachine`: fix a test failure where an injected error is swallowed

jorisdral:
Failure found in https://github.com/IntersectMBO/lsm-tree/actions/runs/15140425104/job/42562757365?pr=683

The test failure can be reproduced using the following seed:

```
cabal run lsm-tree-test -- \
  -p '$0=="lsm-tree.Test.Database.LSMTree.StateMachine.propLockstep_RealImpl_MockFS_IO"' \
  --quickcheck-replay="(SMGen 11798857835899383018 1561550946174790845,97)"
```

The test fails because an injected error with an `FsReachedEOF` type is
swallowed by the `openTableFromSnapshot` public API function. We fix the test
failure (for now) by not generating errors of that type.

At some point we should think of a more principled approach to this type of
swallowed error. `FsReachedEOF` is not an error type that any of the `HasFS`
primitives can throw in practice when using the real file system, but `fs-sim`
error injection *can* throw this error type. Maybe `fs-sim` shouldn't inject
this type of error.

### DIFF
--- a/test/Test/Database/LSMTree/Internal/WriteBufferReader/FS.hs
+++ b/test/Test/Database/LSMTree/Internal/WriteBufferReader/FS.hs
@@ -22,7 +22,7 @@ import qualified System.FS.Sim.MockFS as MockFS
 import qualified System.FS.Sim.Stream as Stream
 import           Test.Tasty
 import           Test.Tasty.QuickCheck as QC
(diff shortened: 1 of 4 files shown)


dcoutts (approved):
LGTM

PR #732, opened 2025-05-21
Remove `blockio:fcntl-nocache` in favour of the `unix` package

jorisdral:
File caching setter/getter functions are available since `unix-2.8.7.0`, so we use those in favour of `blockio:fctnl-nocache`.


blockio/src-linux/System/FS/BlockIO/Internal.hs
```diff
@@ -53,7 +52,7 @@ ioHasBlockIO hfs  params =
 
 hSetNoCache :: Handle HandleIO -> Bool -> IO ()
 hSetNoCache h b =
-  FS.withOpenHandle "hSetNoCache" (handleRaw h) (flip Unix.writeFcntlNoCache b)
+  FS.withOpenHandle "hSetNoCache" (handleRaw h) (flip Fcntl.fileSetCaching (not b))
```
jorisdral, line 55: Note that we're setting to `not b`. The old `writeFcntlNoCache` setting the opposite of what the new `fileSetCaching` does

dcoutts (approved):

PR #733, opened 2025-05-22
Update `hackage-index-state`s for GHA lint jobs

jorisdral:
They're over a year old, so it probably doesn't hurt to set them to a more recent date.

Also: update the `hlint` version


dcoutts (approved):
LGTM. If CI passes that's enough.

PR #734, opened 2025-05-22
Downgrade version mismatches in scripts from errors to warnings

jorisdral:
Also: remove double use of `./scripts/lint-actionlint.sh` in `./scripts/pre-commit.sh`.


wenkokke:
@jorisdral For formatters this should remain an error. If we run different versions of the formatters, then we get really, really messy commits formatting the contents of the files back and forth. Furthermore, since we run these formatters on CI, running the wrong version locally leads to the pre-commit script passing without errors but CI failing.

jorisdral:
> @jorisdral For formatters this should remain an error. If we run different versions of the formatters, then we get really, really messy commits formatting the contents of the files back and forth. Furthermore, since we run these formatters on CI, running the wrong version locally leads to the pre-commit script passing without errors but CI failing.

I'm fine with keeping at an error

FWIW however, `cabal-fmt` and `stylish-haskell` have rarely (if never) had breaking changes in the way they format files in the years I've used them. Insisting on a specific version seems a little strict. Alternatively, I could make it error by default and allow running it with lower severity too. And regarding the pre-commit script, we could still make it fail if we read from stderr and exit when there's any output there

jorisdral:
> > > @jorisdral For formatters this should remain an error. If we run different versions of the formatters, then we get really, really messy commits formatting the contents of the files back and forth. Furthermore, since we run these formatters on CI, running the wrong version locally leads to the pre-commit script passing without errors but CI failing.
> > 
> > 
> > I'm fine with keeping it an error
> > FWIW however, `cabal-fmt` and `stylish-haskell` have rarely (if never) had breaking changes in the way they format files in the years I've used them. Insisting on a specific version seems a little strict. Alternatively, I could make it error by default and allow running it with lower severity too. And regarding the pre-commit script, we could still make it fail if we read from stderr and exit when there's any output there
> 
> Is the change meant to relax the experience for developers or relax the constraints on CI? Because I feel that for CI this should very much remain an error, and for developers it's easy enough to ignore the exit code. We could rewrite the scripts to set the exit code but still run the formatter.

Relax the experience for developers. I understand that for CI we'd want the version check to be strict

wenkokke (approved):
LGTM, but I wrote it, so I'll leave this for @jorisdral to merge.

dcoutts (approved):

PR #739, opened 2025-05-26
feat(scripts): add lint-diskio-complexities

wenkokke:
This PR adds a script to dump the disk I/O complexities from the table in the package description.

PR #741, opened 2025-05-28
fix: ensure consistent disk I/O complexities between the package description and the full and simple APIs

wenkokke:
This ensures consistent documentation on the disk I/O complexities between the package description, the full API, and the simple API, correcting those differences found by the scripts on #739.


dcoutts (approved):
Nice. Hard to keep this consistent without tooling, so yeah I agree it's worth it.

PR #744, opened 2025-06-02
Remove internal trace types from the public API

jorisdral:
They are hidden behind a debug flag now.

### DIFF
--- a/src/Database/LSMTree.hs
+++ b/src/Database/LSMTree.hs
@@ -189,19 +189,9 @@ module Database.LSMTree (
   SessionTrace (..),
   TableTrace (..),
   CursorTrace (..),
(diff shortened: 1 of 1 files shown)


jorisdral:
> Honestly, it's probably fine to just not export the constructors from the main API. The internal modules are internal for a reason, so you shouldn't be using those.

Which constructors do you mean? `MergeTrace` isn't exposed from the public API anymore, that's why I removed the exports of types that were mentioned by `MergeTrace`

jorisdral:
That could be a valid alternative, but I'd rather use the current approach and just not have the traces in release builds. Ideally we'd revisit merge traces later when we start comparing the real implementation and prototype

PR #745, opened 2025-06-02
Move `IsList` and `IsString` instances for `RawBytes` to extras

jorisdral:
I think their main use case is as utilities for testing -- using them in production code would probably be bad for performance, and there are probably not a lot of use cases for it.

I also think the `Semigroup` and `Monoid` instances are iffy, but it's too much of hassle to remove those now.


jorisdral:
Hmm, seems there are docspecs where the `IsString` instance is used. I suppose we could keep the instances in the main library :shrug:

wenkokke (approved):

PR #746, opened 2025-06-02
Lift the restriction of minimum 8-byte keys for the compact index

jorisdral:
We make`topBits64` safe for raw bytes of size less than 8. When the size of the raw bytes is less than 8, then a fallback method is used that is likely slower. When the size of the raw bytes is 8 or higher, then an
additional integer comparison is made, but this has very little impact on
performance, which I verified using micro-benchmarks. The upside to making the
function safe for any input raw bytes is that the API and test generators become
simpler, because there is one fewer constraint to satisfy: the minimum size of 8
bytes for serialised keys.

Now that `topBits64` is fully safe, remove the 8-byte key constraint when using
the compact index. Instead, the config option for the index type includes a hint
not to use the compact index if keys are too small, because that will lead
to bad performance.

### DIFF
--- a/doc/format-run.md
+++ b/doc/format-run.md
@@ -198,8 +198,9 @@ big-endian.
 The compact index type is designed to work with keys that are large
 cryptographic hashes, e.g. 32 bytes. In particular it requires:
 
-* keys must be uniformly distributed
-* keys must be at least 8 bytes (64bits), but can otherwise be variable length
+* keys must be uniformly distributed;
+* keys can be of variable length;
+* keys less than 8 bytes (64bits) are padded with zeros (in LSB position).
 
 For this important special case, we can do significantly better than storing a
 whole key per page: we can typically store just 8 bytes (64bits) per page. This
--- a/src/Database/LSMTree/Internal/Config.hs
+++ b/src/Database/LSMTree/Internal/Config.hs
@@ -20,7 +20,6 @@ module Database.LSMTree.Internal.Config (
     -- * Fence pointer index
   , FencePointerIndexType (..)
   , indexTypeForRun
-  , serialiseKeyMinimalSize
     -- * Disk cache policy
   , DiskCachePolicy (..)
   , diskCachePolicyForLevel
@@ -36,11 +35,9 @@ import           Database.LSMTree.Internal.Index (IndexType)
 import qualified Database.LSMTree.Internal.Index as Index
                      (IndexType (Compact, Ordinary))
 import qualified Database.LSMTree.Internal.MergingRun as MR
-import qualified Database.LSMTree.Internal.RawBytes as RB
 import           Database.LSMTree.Internal.Run (RunDataCaching (..))
 import           Database.LSMTree.Internal.RunAcc (RunBloomFilterAlloc (..))
 import           Database.LSMTree.Internal.RunBuilder (RunParams (..))
-import           Database.LSMTree.Internal.Serialise.Class (SerialiseKey (..))
 
 newtype LevelNo = LevelNo Int
   deriving stock (Show, Eq, Ord)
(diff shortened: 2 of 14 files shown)


src/Database/LSMTree/Internal/RawBytes.hs
```diff
@@ -172,14 +175,29 @@ drop = coerce VP.drop
...
+topBits64 rb@(RawBytes v@(VP.Vector (I# off#) _size (ByteArray k#)))
+  | n >= 8
+  = toWord64 (indexWord8ArrayAsWord64# k# off#)
+  | otherwise
+  = VP.foldl' f 0 v `unsafeShiftL` ((8 - n) * 8)
+  where
+    !n = size rb
+
+    f :: Word64 -> Word8 -> Word64
+    f acc w = acc `unsafeShiftL` 8 + fromIntegral w
```
dcoutts, line 200: I think this can be done more straightforwardly and efficiently, without a loop

You'd need to:
1. use `indexWord8ArrayAsWord64#` to do an unaligned read
2. use the `size rb` to generate a bitmask, to zero out all the undefined bits: ((8 - n) * 8) bits of trailing 0s
3. apply the bitmask with `.&.`

I don't think this will be backwards, but if so, use a `byteSwap64` between step 1 and 2

test/Test/Database/LSMTree/Internal/RawBytes.hs
```diff
@@ -92,3 +99,22 @@ prop_ordAntisymmetry = mapSize (const 4)    $
...
+  where
+    expected =
+      let ws = take 8 (VP.toList v ++ repeat 0)
+      in  List.foldl' (\acc w -> acc `shiftL` 8 + fromIntegral w) 0 ws
+
+-- | If @x@ has fewer than 8 bytes, then all missing bits in the result default
+-- to 0s.
+prop_topBits64_default0s :: RawBytes -> Property
+prop_topBits64_default0s x =
+    RB.topBits64 x === RB.topBits64 (x <> mconcat (replicate 8 [0]))
```
dcoutts, line 120: :+1:

doc/format-run.md
```diff
@@ -199,7 +199,7 @@ The compact index type is designed to work with keys that are large
 cryptographic hashes, e.g. 32 bytes. In particular it requires:
 
 * keys must be uniformly distributed
-* keys must be at least 8 bytes (64bits), but can otherwise be variable length
+* keys should ideally be at least 8 bytes (64bits), but can otherwise be variable length
```
dcoutts, line 202: I'm not totally sure it's true. If the keys are smaller then there's also fewer possible keys, since the keys are distinct.

For example, if the keys are 32bit. They're still distinct. So two adjacent keys (last one one page and first on next) will necessarily have distinct compact 64bit key prefixes. Indeed can't we _only_ get collisions for keys larger than 64bits?

src/Database/LSMTree/Internal/Config.hs
```diff
@@ -321,12 +318,7 @@ data FencePointerIndexType =
   | {- |
     Compact indexes are designed  for the case where the keys in the database are uniformly distributed, e.g., when the keys are hashes.
 
-    When using a compact index, the 'Database.LSMTree.Internal.Serialise.Class.serialiseKey' function must satisfy the following additional law:
-
-    [Minimal size]
-      @'Database.LSMTree.Internal.RawBytes.size' ('Database.LSMTree.Internal.Serialise.Class.serialiseKey' x) >= 8@
-
-    Use 'serialiseKeyMinimalSize' to test this law.
+    When using a compact index, serialised keys should ideally contain 8 bytes or more.
```
dcoutts, line 321: If we change our mind above, need to change here too.

src/Database/LSMTree/Internal/RawBytes.hs
jorisdral: I tried not to read outside of the array for fear of segfaults, but I could try your approach

doc/format-run.md
jorisdral: I changed the wording only because it's not a hard requirement from the code anymore -- no errors thrown if you have <8 byte keys

Also, keys don't have to necessarily be all of the same size. I could imagine a key like `data Key = Key1 Word32 | Key2 Word64` where the serialised key includes a single byte for the constructor, and then either a `Word32` or `Word64`. It's still a degenerate use case, I guess

jorisdral: I'm fine with changing the wording back -- "keys must be uniformly distributed" is also not something we check

dcoutts: We can't check they're uniformly distributed.

Suggestion:
```
* keys must be uniformly distributed;
* keys can be of variable length;
* keys less than 8 bytes (64bits) are padded with zeros (in LSB position).
```

PR #756, opened 2025-06-11
GHA: fix caching in `lint-generate-readme` job

jorisdral:
We should use the index state


dcoutts (approved):
I don't understand the github CI but I'll approve anyway since you do :stuck_out_tongue:

PR #757, opened 2025-06-12
Final demo, final report on functional requirements

jorisdral:
And some small changes that popped up while creating the demo.


dcoutts (approved):
LGTM. Drop the patch that adds the final report, since we've got a dedicated PR for that #773.

PR #759, opened 2025-06-16
Add a script that can be used to check release builds

jorisdral:
Basically, if a release tag of the form `blockio-*`, `bloomfilter-blocked-*` or `lsm-tree-*` is created, the workflow will trigger and it will try to build the corresponding library without the local dev environment. Instead, it will build the library without tests and benchmark with the latest Hackage index-state. This should be a sanity check that the library version will work out of the box once it is released to Hackage

### DIFF
--- a/.github/workflows/check-release-builds.yml
+++ b/.github/workflows/check-release-builds.yml
@@ -1,0 +1,123 @@
+name: Check release builds
+
+on:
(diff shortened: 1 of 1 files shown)


dcoutts (approved):
Sounds useful.

Issue #752, opened 2025-06-04
[BUG] Reference never released in `unit_union_blobref_invalidation` test


jorisdral:
**Describe the bug**
The `unit_union_blobref_invalidation` test failed with a "reference never released" exception.

**To Reproduce**
Unfortunately, I can't reproduce the bug locally. The failing test can be run using `cabal run lsm-tree-test -- -p "unit_union_blobref_invalidation"`

**Expected behavior**
Every reference should eventually be released, so such an exception should not occur.

**Output**
The test output can be found here: https://github.com/IntersectMBO/lsm-tree/actions/runs/15437001994/job/43445623120?pr=751

```
     unit_union_blobref_invalidation:                      FAIL
      Exception: Exceptions thrown while committing an action registry.
        A registered action threw an error: 
          The error:
        
            Reference is never released: RefId 488704
            Allocation site: CallStack (from HasCallStack):
              newRefWithTracker, called at src-control/Control/RefCount.hs:316:5 in lsm-tree-0.1.0.0-inplace-control:Control.RefCount
              dupRef, called at src/Database/LSMTree/Internal/MergingRun.hs:279:47 in lsm-tree-0.1.0.0-inplace:Database.LSMTree.Internal.MergingRun
        
          Registration site:
        
            CallStack (from HasCallStack):
              mkAction, called at src-control/Control/ActionRegistry.hs:644:40 in lsm-tree-0.1.0.0-inplace-control:Control.ActionRegistry
              delayedCommit, called at src/Database/LSMTree/Internal/MergeSchedule.hs:300:7 in lsm-tree-0.1.0.0-inplace:Database.LSMTree.Internal.MergeSchedule
```

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - OS version: 24.04.2
 - GHC: ghc-9.12

**Additional context**
Since the bug can't be reproduced locally, I think this failure relies on a very specific scheduling of the garbage collector. My reason for thinking is that references are only marked as forgotten while garbage collecting, after which a "reference never released" exception can be thrown.

jorisdral:
Happened again! https://github.com/IntersectMBO/lsm-tree/actions/runs/16001549609/job/45137589776

PR #774, opened 2025-07-01
Issue #752: try to fix forgotten reference

jorisdral:
Since I can't reproduce the bug but it has popped up twice now, I've tried to fix the test by making sure that we use `bracket`s everywhere. Hopefully that is the fix.


jorisdral:
Superseded by #781

PR #775, opened 2025-07-01
Don't forget to release `HasBlockIO` in the public API

jorisdral:
There are a few places where functions in the public API create a `HasBlockIO` for the user, but since #742 when a session is closed the `HasBlockIO` it contains does not get closed automatically anymore. So in the public API, we have to be extra careful to do so when the public API opens a `HasBlockIO`.


dcoutts (approved):
LGTM

PR #780, opened 2025-07-07
Rename `blockio` to `blockio-api`

jorisdral:
The name `blockio` is already taken on Hackage


jeltsch (changes requested):
I think it’s not a good idea to include the term “API” in the name. For one thing, what the user gets by using this package is support for block I/O; the “API” is just the interface to this package. For another thing, “API” means “application programming interface”, and block I/O is not only used by applications but also by other libraries.

That said, there is already an issue with the name `blockio`: “block I/O” is two words, namely “block” and “I/O”. Therefore, the package name should use a dash as a separator. And this actually shows a way out: Let’s call the package `block-io`, which kills two birds with one stone: the name reflects the two-word structure, and it doesn’t clash with an already used name.

jorisdral:
> I don't see any such package:
> 
>     * https://hackage.haskell.org/package/blockio   -- 404
> 
>     * https://hackage.haskell.org/package/blockio/candidates/  -- 404
> 
>     * https://hackage.haskell.org/packages/search?terms=blockio
> 
> 
> Nope, none, nada, niet.

It gave me an error that the package already existed when I tried to upload to Hackage. The error referred me to this maintainer page: https://hackage.haskell.org/package/blockio/maintainers/

PR #783, opened 2025-07-09
`utxo-bench`: introduce a new `RunMode` option

jorisdral:
Previously, the `run` command of the benchmark executable had two flags: `--pipelined` and `--lookup-only`. The value of the second flag had no effect if `--pipelined` was passed to the executable. Now there is a new `--mode` option that can be set to `Sequential`, `Pipelined`, or `LookupOnly`. `Sequential` is the name for what previously was the default mode of the `run` command.

Also: add an update-only mode


dcoutts (approved):
LGTM. And it turns out the report is not quite so specific, so we don't need to update the report for this change in CLI.

PR #804, opened 2025-11-25
Allow filepath 1.4

jasagredo:
# Description

If some package fixes the dependency of `ghc` then all base libraries are fixed. By allowing filepath 1.4 we open up the possibility of depending on `ghc` 9.6.7 and still using the `lsm-tree` package.

# Checklist


jasagredo:
This should probably just be a revision on Hackage.

PR #805, opened 2025-11-25
Allow unix 2.8.6

jasagredo:
# Description

If some package fixes the dependency of `ghc` then all base libraries are fixed. By allowing `unix 2.8.6` we open up the possibility of depending on `ghc` 9.6.7 and still using the `lsm-tree` package.

# Checklist

### DIFF
--- a/blockio/src-linux/System/FS/BlockIO/Internal/Fcntl.hsc
+++ b/blockio/src-linux/System/FS/BlockIO/Internal/Fcntl.hsc
@@ -1,0 +1,38 @@
+{-# LANGUAGE CPP #-}
+-- | Compatibility layer for the @unix@ package to provide a @fileSetCaching@ function.
+--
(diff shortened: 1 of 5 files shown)


blockio/src-linux/System/FS/BlockIO/Internal/Fcntl.hsc
```diff
@@ -0,0 +1,27 @@
+{-# LANGUAGE CPP #-}
+module System.FS.BlockIO.Internal.Fcntl (fileSetCaching) where
+
+#if MIN_VERSION_unix(2,8,7)
+
+import System.Posix.Fcntl (fileSetCaching)
+
+#else
+
+#define _GNU_SOURCE
```
jorisdral: Through other channels you explained to me why we need this `#define` here. For posterity, maybe we should link to documentation that explains why this has to be included?

PR #814, opened 2026-02-16
ScheduledMerges: also test nested unions and arbitrary configs

mheinzel:
# Description

We had already noted that testing nested unions (unions of tables that are themselves result of a union) would be a good idea. Also, after making some LSM parameters configurable in #716, it seems sensible to test more than just the default values.

As a preparation for further changes to the prototype, this PR addresses these points and makes small improvements to the invariants and properties we assert.


mheinzel:
Small change: I used newtypes for test inputs now, making their `Show` instances much more readable than with the `NonEmpty { getNonEmpty = ... }` sprinkled in. This made it easier to understand the counterexamples from failing properties.

jorisdral (approved):

PR #815, opened 2026-02-23
docs: use 'sequence' instead of 'run' in IndexCompact docs

### LINKED ISSUE #211: Replace the term “run” by something else where it is misleading
The section *Representing clashes and larger-than-page entries* in the `compact` documentation part of `Database.LSMTree.Internal.IndexCompact` uses the word “run” for things other than what is usually called runs in the context of LSMTs. This can be misleading, especially to newcomers. I suggest to replace “run” by something else, like “series” or “sequence”.

kanyuku:
This PR replaces the potentially misleading term "run" with "sequence" within the internal index documentation. This clarification ensures that internal sequences are not confused with high-level LSM-tree runs, making the documentation more precise.

Fixes #211


Emmanuel-Tyty:
@jorisdral can you take a look at this?

PR #822, opened 2026-03-03
ScheduledMerges: migrate union level late (when used in a new merge)

mheinzel:
Builds on top of #814.

When creating the union of multiple tables, we create a special union level containing a merging tree. Once the tree has been completed, i.e. merged into a single run, we want to get rid of the union level by migrating the run to the regular levels. The main motivation is that we want it to become part of a last level merge, so compaction can occur. Last level merges are especially useful for compaction, since they can drop all `delete` entries.

This PR demonstrates one of two approaches for migration. Since the main motivation is to allow for creating last level merges, we delay migration until the exact point where the completed union can become part of a merge. This avoids weakening the invariants of the regular levels, which would often be necessary to migrate the union earlier.

I still plan to add some tracing once we decided which approach for migrating the union we want to take.

### DIFF
--- a/lsm-tree/src-prototypes/ScheduledMerges.hs
+++ b/lsm-tree/src-prototypes/ScheduledMerges.hs
@@ -1227,9 +1227,16 @@ doLookup wb runs ul k =
 
     -- both regular and union level: submit multiple batches, combine in the end
     lookupBoth tree = do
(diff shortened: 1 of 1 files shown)


mheinzel:
Closed in favour of #824.

PR #823, opened 2026-03-03
ScheduledMerges: allow empty levels

mheinzel:
Builds on top of #814.

This can generally be useful, but the main motivation now is to allow migration of the union level into the regular levels as soon as the merging tree is completed.

So far, we always required an `IncomingRun` to be present on each level, but already had a special case for where this could be a plain `Run` that doesn't require any merging. Instead, we can make the presence of the `IncomingRun` itself optional, which allows us to have no runs on a level at all. The resulting changes are minor, but we should perhaps think about more suitable terms for `IncomingRun` and/or the `Maybe IncomingRun` (which might deserve its own data type), at least in the real implementation).

### DIFF
--- a/lsm-tree/src-prototypes/ScheduledMerges.hs
+++ b/lsm-tree/src-prototypes/ScheduledMerges.hs
@@ -1,7 +1,6 @@
 {-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE UnboxedTuples   #-}
 
-{-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
 {-# OPTIONS_GHC -Wno-partial-fields #-}
 
 -- | A prototype of an LSM with explicitly scheduled incremental merges.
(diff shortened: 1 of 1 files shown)


lsm-tree/src-prototypes/ScheduledMerges.hs
```diff
@@ -159,20 +159,20 @@ type Levels s = [Level s]
...
--- somewhat.
+-- | A level is a sequence of resident runs at this level, optionally prefixed
+-- by an incoming run, which consists of multiple runs that are being merged.
+-- Once completed, the resulting run will become a regular run at this level.
+-- A single run that gets added to a level immediately becomes a regular run,
+-- rather than putting it in as a 'CompletedMerge'. This is for two reasons:
+-- to see statically that it's a single run without having to read the 'STRef',
+-- and secondly to make it easier to avoid supplying merge credits. It's not
+-- essential, but simplifies things somewhat.
+data Level s = Level !(Maybe (IncomingRun s)) ![Run]
```
jorisdral: I think we agreed we would try out a representation where the level can be either empty, or non-empty. Something along the lines of

```hs
data Level = EmptyLevel | Level ...
```

lsm-tree/src-prototypes/ScheduledMerges.hs
```diff
@@ -355,6 +359,9 @@ invariant conf@LSMConfig{..} (LSMContent _ levels ul) = do
     levelsInvariant :: Int -> Levels s -> ST s ()
     levelsInvariant !_ [] = pure ()
 
+    levelsInvariant !_ (EmptyLevel : ls) = do
+      assertST $ not (null ls)  -- last level shouldn't be empty
```
jorisdral: :+1: makes sense, we don't want "trailing empty levels" so to speak

PR #824, opened 2026-03-04
ScheduledMerges: migrate union level early (as soon as merging tree is completed)

mheinzel:
When creating the union of multiple tables, we create a special union level containing a merging tree. Once the tree has been completed, i.e. merged into a single run, we want to get rid of the union level by migrating the run to the regular levels. The main motivation is that we want it to become part of a last level merge, so compaction can occur. Last level merges are especially useful for compaction, since they can drop all delete entries.

This PR demonstrates one of two approaches for migration (#822 being the other one). As soon as the merging tree gets completed through `supplyUnionCredits`, we migrate it to a new last regular level. Sometimes it would not fit directly after the existing levels, in which case we can insert a few empty levels.

An interesting case to consider it that a union can get completed by an operation on another table that contains a shared reference to the merging tree. Therefore, at the beginning of any operation, a completed but not yet migrated union might be present. To ensure unions can become part of merges as soon as possible, we also check for possible migrations when flushing the write buffer. It might be beneficial to do the same when doing lookups.

### DIFF
--- a/lsm-tree/src-prototypes/ScheduledMerges.hs
+++ b/lsm-tree/src-prototypes/ScheduledMerges.hs
@@ -175,7 +175,16 @@ data Level s = Level !(IncomingRun s) ![Run]
 data IncomingRun s = Merging !MergePolicyForLevel
                              !NominalDebt !(STRef s NominalCredit)
                              !(MergingRun LevelMergeType s)
-                   | Single  !Run
+                   | Single  !SingleRunOrigin !Run
+
+-- | Additional information about the origin of a 'Single' run. This allows us
+-- to have stronger invariants.
+data SingleRunOrigin = FromWriteBuffer
+                     | FromLevellingLevel
+                       -- | A former union level that was completed (merged down
+                       -- to a single run) and became the last regular level.
+                     | FromMigratedUnion
+  deriving stock (Eq, Show)
 
 -- | The merge policy for a LSM level can be either tiering or levelling.
 -- In this design we use levelling for the last level, and tiering for
@@ -363,7 +372,7 @@ invariant conf@LSMConfig{..} (LSMContent _ levels ul) = do
 
     levelsInvariant !ln (Level ir rs : ls) = do
       mrs <- case ir of
-        Single r ->
+        Single _ r ->
           pure (CompletedMerge r)
         Merging mp _ _ (MergingRun mt _ ref) -> do
           assertST $ ln > 1  -- no merges on level 1
@@ -422,10 +431,16 @@ invariant conf@LSMConfig{..} (LSMContent _ levels ul) = do
             -- 2. when the union level gets migrated.
             --
             -- In the latter case, it can be arbitrarily small.
-            (Single r, m) -> do
+            (Single origin r, m) -> do
               assertST $ case m of CompletedMerge{} -> True
                                    OngoingMerge{}   -> False
-              assertST $ runToLevelNumber LevelLevelling conf r <= ln
+              case origin of
+                FromWriteBuffer ->
+                  assertST False  -- we don't flush to levelling levels
+                FromLevellingLevel ->
+                  assertST $ runToLevelNumber LevelLevelling conf r == ln
+                FromMigratedUnion ->
+                  assertST $ runToLevelNumber LevelLevelling conf r <= ln
 
             -- A completed merge for levelling can be of almost any size at all!
             -- It can be smaller, due to deletions in the last level. But it
@@ -452,9 +467,10 @@ invariant conf@LSMConfig{..} (LSMContent _ levels ul) = do
           case (ir, mrs, mergeTypeForLevel ls ul) of
             -- A single incoming run (which thus didn't need merging) must be
             -- of the expected size already
-            (Single r, m, _) -> do
+            (Single origin r, m, _) -> do
               assertST $ case m of CompletedMerge{} -> True
                                    OngoingMerge{}   -> False
+              assertST $ origin == FromWriteBuffer
               assertST $ runToLevelNumber LevelTiering conf r == ln
 
             -- A completed last level run can be of almost any smaller size due
@@ -1267,7 +1283,8 @@ migrateUnionLevel tr sc conf ls ul@(Union t _) =
                                -- Run must not be too large for the level.
               ]
         let emptyLevels = replicate (levelNo - 1 - length ls) EmptyLevel
-        let levels = ls ++ emptyLevels ++ [Level (Single r) []]
+        let level = Level (Single FromMigratedUnion r) []
+        let levels = ls ++ emptyLevels ++ [level]
         assertST $ length levels == levelNo
         traceWith tr $ EventAt sc levelNo $ UnionLevelMigratedEvent r
         pure (levels, NoUnion)
@@ -1531,7 +1548,7 @@ increment tr sc conf run0 ls0 ul = do
 
     go !ln incoming (Level ir rs : ls) = do
       r <- case ir of
-        Single r -> do
+        Single _ r -> do
           traceWith tr' $ SingleRunCompletedEvent r
           pure r
         Merging mergePolicy _ _ mr -> do
@@ -1602,10 +1619,18 @@ newLevelMerge :: Tracer (ST s) EventDetail
               -> LSMConfig
               -> Int -> MergePolicyForLevel -> LevelMergeType
               -> [Run] -> ST s (IncomingRun s)
-newLevelMerge tr _ _ _ _ [r] =  do
+newLevelMerge tr _ ln mergePolicy _ [r] = do
+    origin <-
+      if ln <= 1
+        then do
+          assertST $ ln == 1
+          pure FromWriteBuffer
+        else do
+          assertST $ mergePolicy == LevelLevelling
+          pure FromLevellingLevel
     traceWith tr $ NewSingleRunEvent r
-    pure (Single r)
-newLevelMerge tr conf@LSMConfig{..} level mergePolicy mergeType rs = do
+    pure (Single origin r)
+newLevelMerge tr conf@LSMConfig{..} ln mergePolicy mergeType rs = do
     mergingRun@(MergingRun _ physicalDebt _) <- newMergingRun mergeType rs
     traceWith tr NewLevelMergeEvent {
                    mergePolicy,
@@ -1621,7 +1646,7 @@ newLevelMerge tr conf@LSMConfig{..} level mergePolicy mergeType rs = do
     -- The nominal debt equals the minimum of credits we will supply before we
     -- expect the merge to complete. This is the same as the number of updates
     -- in a run that gets moved to this level.
-    nominalDebt = NominalDebt (levelNumberToMaxRunSize LevelTiering conf level)
+    nominalDebt = NominalDebt (levelNumberToMaxRunSize LevelTiering conf ln)
 
     -- The physical debt is the number of actual merge steps we will need to
     -- perform before the merge is complete. This is always the sum of the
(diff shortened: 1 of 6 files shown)


lsm-tree/src-prototypes/ScheduledMerges.hs
```diff
@@ -415,12 +415,17 @@ invariant conf@LSMConfig{..} (LSMContent _ levels ul) = do
...
+            -- 1. when a levelling level is full, so the run gets promoted to
+            --    the new last level
+            -- 2. when the union level gets migrated.
+            --
+            -- In the latter case, it can be arbitrarily small.
             (Single r, m) -> do
               assertST $ case m of CompletedMerge{} -> True
                                    OngoingMerge{}   -> False
-              assertST $ runToLevelNumber LevelLevelling conf r == ln
+              assertST $ runToLevelNumber LevelLevelling conf r <= ln
```
jorisdral: Might it make sense to tighten the invariant, and have a separate case for when the `Single` came from a finished union? The invariant now is quite loose: there is not much of a guarantee about the size of the run anymore, so when a new level is created during normal operation with unions, the run could be of a "bad" size

lsm-tree/src-prototypes/ScheduledMerges.hs
```diff
@@ -1151,12 +1159,17 @@ newtype UnionDebt = UnionDebt Debt
...
+-- clear that 'supplyUnionCredits' should still be called to trigger migration,
+-- even if the merging tree itself has been completed. This becomes necessary
+-- when the tree has been completed by operations on another table via sharing.
 remainingUnionDebt :: LSM s -> ST s UnionDebt
 remainingUnionDebt (LSMHandle _ _ _conf lsmr) = do
     LSMContent _ _ ul <- readSTRef lsmr
     UnionDebt <$> case ul of
       NoUnion      -> pure 0
-      Union tree d -> checkedUnionDebt tree d
+      Union tree d -> (+1) <$> checkedUnionDebt tree d
```
jorisdral, line 1188: :+1:

lsm-tree/src-prototypes/ScheduledMerges.hs
```diff
@@ -1225,6 +1224,53 @@ checkedUnionDebt tree debtRef = do
...
+      Just r
+        | runSize r == 0 -> do
+          -- If the union level is empty (due to compaction), just drop it.
+          -- There is no real level number for this event, but it seems like an
+          -- important part of the level events.
+          --
+          -- TODO: instead use the level it *would* have ended up if non-empty?
+          -- Or use a sum type for the level number?
+          traceWith tr $ EventAt sc 0 EmptyUnionLevelDroppedEvent
+          pure (ls, NoUnion)
```
jorisdral: If we drop the union level, could invariants like

```hs
        Merging mp _ _ (MergingRun mt _ ref) -> do
          assertST $ mp == mergePolicyForLevel ln ls ul
```

be violated? I think so, right?

We should double check that dropping the union level has no other adverse affects on the invariants

lsm-tree/src-prototypes/ScheduledMerges.hs
```diff
@@ -1225,6 +1224,53 @@ checkedUnionDebt tree debtRef = do
...
+        pure (ls, ul)
+      Just r
+        | runSize r == 0 -> do
+          -- If the union level is empty (due to compaction), just drop it.
+          -- There is no real level number for this event, but it seems like an
+          -- important part of the level events.
+          --
+          -- TODO: instead use the level it *would* have ended up if non-empty?
+          -- Or use a sum type for the level number?
+          traceWith tr $ EventAt sc 0 EmptyUnionLevelDroppedEvent
```
jorisdral: I'm fine with a sum type or even `Maybe Int`. Let's make that change in this PR still

lsm-tree/src-prototypes/ScheduledMerges.hs
```diff
@@ -1225,6 +1224,53 @@ checkedUnionDebt tree debtRef = do
...
+        | otherwise -> do
+          -- In some cases we could even add it as one of the tiering runs of
+          -- the existing last regular level, but the difference is small and
+          -- we would need to relax the invariant that the last level uses
+          -- levelling.
+          let levelNo = maximum
+                [ 2  -- first level is tiering, don't migrate there
+                , length ls + 1  -- must come after existing levels
+                , runToLevelNumber LevelLevelling conf r  -- must fit
+                ]
```
jorisdral: :+1:

lsm-tree/src-prototypes/ScheduledMerges.hs
```diff
@@ -1225,6 +1224,53 @@ checkedUnionDebt tree debtRef = do
...
+          let emptyLevels = replicate (levelNo - 1 - length ls) EmptyLevel
+          let levels = ls ++ emptyLevels ++ [Level (Single r) []]
+          assertST $ length levels == levelNo
+          traceWith tr $ EventAt sc levelNo $ UnionLevelMigratedEvent r
```
jorisdral: This trace message is good I think!

lsm-tree/src-prototypes/ScheduledMerges.hs
jorisdral: Maybe we should also add a targetted test that leads to an empty union, and then perform some operations after. It's a relatively rare occurrence, so having a specific test would make sense (unless I'm overlooking tests we already have)

jorisdral (commented):
Nice to see this take shape :smile:

lsm-tree/src-prototypes/ScheduledMerges.hs
mheinzel: Yes, done. I even split up the possible origins into more than just two cases since it seemed weird to treat flushing the write buffer and promoting a last level run as the same case. Adds a little bit of code, but probably clearer and allows for some additional invariants.

lsm-tree/src-prototypes/ScheduledMerges.hs
mheinzel: I added a sum type.

edit: no longer necessary since I removed the `EmptyUnionLevelDroppedEvent` based on https://github.com/IntersectMBO/lsm-tree/pull/824#discussion_r3008335529.

lsm-tree/src-prototypes/ScheduledMerges.hs
mheinzel: That's true. I also managed to write a failing test for it using `QuickCheck.DynamicLogic`.

Maybe the easiest thing would then be to simply migrate empty runs. We migrate to a levelling level, which already can have empty runs, so there is no need to weaken invariants. The empty run also doesn't need to impact lookup performance. If we think it matters, we can ignore empty runs when constructing the `LevelsCache`.

mheinzel: I changed it so the run gets migrated even if it's empty.

jorisdral: Quick comment (I haven't had time to review the rest of your changes yet): IIRC the actual implementation relies on runs being non-empty, and it assumes they are non-empty in a bunch of places (like `RunReader`)

lsm-tree/src-prototypes/ScheduledMerges.hs
```diff
@@ -415,12 +424,23 @@ invariant conf@LSMConfig{..} (LSMContent _ levels ul) = do
...
               assertST $ case m of CompletedMerge{} -> True
                                    OngoingMerge{}   -> False
-              assertST $ runToLevelNumber LevelLevelling conf r == ln
+              case origin of
+                FromWriteBuffer ->
+                  assertST False  -- we don't flush to levelling levels
+                FromLevellingLevel ->
+                  assertST $ runToLevelNumber LevelLevelling conf r == ln
+                FromMigratedUnion ->
+                  assertST $ runToLevelNumber LevelLevelling conf r <= ln
```
jorisdral, line 443: There is the third way of flushing a write buffer, we should mention that here too

lsm-tree/src-prototypes/ScheduledMerges.hs
```diff
@@ -1225,6 +1242,50 @@ checkedUnionDebt tree debtRef = do
...
+migrateUnionLevel tr sc conf ls ul@(Union t _) =
+    getCompletedMergingTree t >>= \case
+      Nothing ->
+        -- Still in progress, leave it.
+        pure (ls, ul)
+      Just r -> do
+        -- In some cases we could even add it as one of the tiering runs of
+        -- the existing last regular level, but the difference is small and
+        -- we would need to relax the invariant that the last level uses
+        -- levelling.
```
jorisdral: This one I'm not sure I fully understand. When would it be added as a tiering run?

lsm-tree/src-prototypes/ScheduledMerges.hs
```diff
@@ -1225,6 +1242,50 @@ checkedUnionDebt tree debtRef = do
...
+        --
+        -- Also note that we even migrate empty runs. This doesn't violate any
+        -- invariants, as levelling levels can already contain empty runs.
+        -- Dropping the run would be more complicated since the previous level
+        -- then suddenly becomes the last one while potentially still containing
+        -- midlevel merges.
```
jorisdral: It violates the invariant that runs should be non-empty. Maybe we're not checking that in the prototype, but the actual implementation does rely on that

Would it be simpler only check when a merge is created that it has the correct merge policy, rather than asserting it every time we run the invariant? Then we don't have to migrate empty runs. Or maybe we should allow a single trailing empty level?

lsm-tree/src-prototypes/ScheduledMerges.hs
```diff
@@ -1225,6 +1242,50 @@ checkedUnionDebt tree debtRef = do
...
+        let levelNo = maximum
+              [ 2  -- first level is tiering, don't migrate there
+              , length ls + 1  -- must come after existing levels
+              , runToLevelNumber LevelLevelling conf r  -- must fit
```
jorisdral: "must fit" here only imposes an upper bound on the run size, right? The run can be arbitrarily small otherwise

lsm-tree/test-prototypes/Test/ScheduledMergesDL.hs
```diff
@@ -0,0 +1,47 @@
...
+
+instance DynLogicModel (Lockstep Model)
+
+prop_empty_union :: Property
+prop_empty_union = forAllDL dl_empty_union prop_LSM
+
+-- | Create a table with an empty run in the union level, then run arbitrary
+-- actions to see if all invariants still hold.
+dl_empty_union :: DL (Lockstep Model) ()
+dl_empty_union = do
```
jorisdral, line 27: Nice!

jorisdral (commented):
Looks good -- the only thing we should maybe reconsider a little bit more is how we handle empty union results

lsm-tree/src-prototypes/ScheduledMerges.hs
mheinzel: That can't happen in a levelling level, as the first level always uses tiering. This is also checked with the `FromWriteBuffer -> assertST False`.

lsm-tree/src-prototypes/ScheduledMerges.hs
mheinzel: Before migration, there usually is a last regular level, which uses tiering. If the union level is small enough, we could theoretically add it as a run into that tiering level (instead of creating a new levelling level for it), but that clashes with our current assertions about merge policy of course.

Maybe I should just remove the comment. I was trying to answer the question a reader might have: "Why do we always have to create a new level just for this union run, even if it's small?" But maybe that causes more confusion than it prevents.

lsm-tree/src-prototypes/ScheduledMerges.hs
mheinzel: Yes. I made it a little clearer now.

lsm-tree/src-prototypes/ScheduledMerges.hs
mheinzel: The current implementation already can have empty runs in the last level (they can result from last level merges) and can deal with them. We might want to make our strategy/documentation there clearer, but that's not something specific to the union migration.

So migrating an empty run seems okay (these should be uncommon either way). We can then still revisit how we deal with empty runs.

lsm-tree/src-prototypes/ScheduledMerges.hs
mheinzel: See https://github.com/IntersectMBO/lsm-tree/pull/824#discussion_r3044399207

lsm-tree/src-prototypes/ScheduledMerges.hs
jorisdral: Alright, I think I misremembered about the actual implementation not supporting empty runs. I stand corrected!

lsm-tree/src-prototypes/ScheduledMerges.hs
jorisdral: I think the first paragraph you wrote already cleared up my confusion. Just add some of those detail to the haddocks and then we should be fine. Maybe something along the lines of:

> Before migrating the run that resulted from the union into the regular levels, there is usually a last "regular" level, which uses tiering. We could add the union-result run to the resident tiering runs of the last regular level, but that clashes with some invariants. In any case, the difference is small and we would need to relax the invariant that the last level uses levelling.

jorisdral (approved):
LGTM! Were there any changes since last time I reviewed that you think I should look at? If not, let's resolve the final few review comments, and then we can put this PR on the merge queue

mheinzel:
> Were there any changes since last time I reviewed that you think I should look at?

No, I only slightly tweaked the comments: https://github.com/IntersectMBO/lsm-tree/compare/16e0b0070478b3b97f6d5a37e1aebd59a44d7f81..051001bb5634e1166f386d1c16fe6c406027ee4e

So I will make some last documentation improvements [as discussed](https://github.com/IntersectMBO/lsm-tree/pull/824#discussion_r3085357310), rebase and merge.

PR #833, opened 2026-03-31
Test backwards compatibility of snapshot decoders

mheinzel:
In addition to the existing roundtrip and golden file tests, we should test that parts of the snapshot that changed between versions can in fact still be decoded correctly.

I looked into the tests while working on supporting empty levels (follow up to #823), which will change the snapshot format. I decided to open a separate PR, as these changes are mostly independent.

The commit messages contain some additional reasoning.

### DIFF
--- a/lsm-tree/test/Test/Database/LSMTree/Internal/Snapshot/Codec/Golden.hs
+++ b/lsm-tree/test/Test/Database/LSMTree/Internal/Snapshot/Codec/Golden.hs
@@ -7,9 +7,11 @@ module Test.Database.LSMTree.Internal.Snapshot.Codec.Golden (
   , Annotation
   ) where
 
+import           Codec.CBOR.Read (deserialiseFromBytes)
 import           Codec.CBOR.Write (toLazyByteString)
 import           Control.Monad (when)
-import qualified Data.ByteString.Lazy as BSL (writeFile)
+import           Control.Monad.Class.MonadThrow (Exception (displayException))
+import qualified Data.ByteString.Lazy as BSL
 import qualified Data.Set as Set
 import           Data.Typeable
 import qualified Data.Vector as V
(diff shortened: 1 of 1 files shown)


lsm-tree/test/Test/Database/LSMTree/Internal/Snapshot/Codec/Golden.hs
```diff
@@ -121,6 +131,43 @@ prop_noUnexpectedOrMissingGoldenFiles = once $ ioProperty $ do
...
+  -> a
+  -> Assertion
+assertGoldenFileDecodesTo proxy ann v expected = do
+    let fp = goldenDataFilePath </> filePathGolden proxy ann v
+    lbs <- BSL.readFile fp
+    case deserialiseFromBytes (decodeVersioned v) lbs of
+      Left err ->
+        assertFailure $ "Error decoding " ++ fp ++ ": " ++ displayException err
+      Right (_, decoded) ->
+        assertEqual "" expected decoded
```
jorisdral, line 170: I like it!

Since we don't exhaustively enumerate all combinations of types in the config, it might be the case in the future that we don't have a suitable "old" golden file available for a compat test we might want to write. But I think that in such cases we could just make sure to generate a suitable golden file before upgrading the encoding/decoding

jorisdral (approved):

lsm-tree/test/Test/Database/LSMTree/Internal/Snapshot/Codec/Golden.hs
```diff
@@ -253,31 +254,32 @@ class EnumGolden a where
...
-
-  supportedVersions :: Proxy a -> [SnapshotVersion]
-  supportedVersions _ = allCompatibleSnapshotVersions
+  singGolden :: SnapshotVersion -> a
+  singGolden v = snd $ head (enumGoldenAnnotated v)
 
 type Annotation = String
 
-enumGoldenAnnotated' :: EnumGolden a => Proxy a -> [(Annotation, a)]
-enumGoldenAnnotated' _ = enumGoldenAnnotated
```
jorisdral, line 233: Oh I missed this change. I'm not sure I understand it. Could you show an example why this is necessary?

jorisdral (changes requested):

lsm-tree/test/Test/Database/LSMTree/Internal/Snapshot/Codec/Golden.hs
jorisdral: I think what confuses me is that I wrote `EnumGolden` to only enumerate types in the current snapshot version, not in older snapshot versions

mheinzel: Ah, that makes sense. The use case I need it for is here:
https://github.com/IntersectMBO/lsm-tree/pull/834/changes#diff-0af7fa8c979b1fc2dc7935d2519fcd8a9ac5c4f9f0891c44a3902f0ccfe7b332R408-R412

Originally, `supportedVersions` was added to omit a new type from old snapshot versions, but now that I'm just adding a new constructor, I want to omit that constructor. That granularity was not supported before, as I could only mark the whole type as supported or not.

jorisdral: Makes sense, thank you for explaining!

jorisdral (approved):

PR #834, opened 2026-04-02
Allow table levels to be empty

mheinzel:
Builds on top of #833, therefore currently still a draft PR.

As part of the work to allow migrating completed union levels, we introduce empty levels in the table structure. They don't get created yet, but will be useful for union migration. This is a similar change to what #823 did in the prototype.


lsm-tree/src-core/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -309,10 +311,9 @@ type Levels m h = V.Vector (Level m h)
 -- | A level is a sequence of resident runs at this level, prefixed by an
 -- incoming run, which is usually multiple runs that are being merged. Once
 -- completed, the resulting run will become a resident run at this level.
```
jorisdral, line 313: Add something like: "levels can be empty if ..."

lsm-tree/src-core/Database/LSMTree/Internal/MergeSchedule.hs
```diff
@@ -673,14 +679,21 @@ addRunToLevels tr conf@TableConfig{..} resolve hfs hbio refCtx root salt uc r0 r
...
     go !ln rs (V.uncons -> Nothing) = do
         traceWith tr $ AtLevel ln TraceAddLevel
         -- Make a new level
         let policyForLevel = mergePolicyForLevel confMergePolicy ln V.empty ul
         ir <- newMerge policyForLevel (mergeTypeForLevel V.empty ul) ln rs
         pure $! V.singleton $ Level ir V.empty
+
+    go !ln rs (V.uncons -> Just (EmptyLevel, ls)) = do
+        assert (not (V.null ls)) $ pure ()  -- no trailing empty levels
+        let policyForLevel = mergePolicyForLevel confMergePolicy ln ls ul
```
jorisdral, line 697: See https://github.com/IntersectMBO/lsm-tree/pull/824#discussion_r3038732245

lsm-tree/src-core/Database/LSMTree/Internal/Snapshot/Codec.hs
```diff
@@ -308,7 +311,9 @@ instance Encode TableConfig where
         <> encode mergeBatchSize
 
 instance DecodeVersioned TableConfig where
-  decodeVersioned v@V0 = do
+  decodeVersioned v
+    | v < V1 = do
```
jorisdral, line 315: Explicit pattern matches might be better? The `Ord` instance depends on the order of constructors and that could be mistakenly changed in the future (though unlikely)

Maybe something like:

```hs
  decodeVersioned v = case v of
     V0 -> decode
     V1 -> decode'
     V2 -> decode'
```

lsm-tree/src-core/Database/LSMTree/Internal/Snapshot/Codec.hs
```diff
@@ -534,17 +538,31 @@ instance DecodeVersioned r => DecodeVersioned (SnapLevels r) where
...
+         encodeListLen 3
+      <> encodeWord 1
+      <> encode incomingRun
       <> encode residentRuns
 
 
 instance DecodeVersioned r => DecodeVersioned (SnapLevel r) where
-  decodeVersioned v = do
+  decodeVersioned v
+    | v < V2 = do
```
jorisdral, line 553: Same comment here about explicit pattern matches

lsm-tree/src-core/Database/LSMTree/Internal/MergeSchedule.hs
mheinzel: (that comment was resolved)

lsm-tree/src-core/Database/LSMTree/Internal/Snapshot/Codec.hs
mheinzel: That indeed seems unlikely and we can write a test that guards against it. ([done](https://github.com/IntersectMBO/lsm-tree/pull/834/changes#diff-32f83866bfd0c4b1a31d2a25ef1195af36cad6858de3865ab93d8be57b8c275eR34-R38))

I don't think the pattern match adds value here. It forces us to revisit this instance whenever we add a new version, but there is not really a reason to. The only thing we could forget is that when we remove support for older versions, we can remove the old parsing code. If we remove `V0` from the type (not just `supportedVersions`, which I'm not sure we want to), the compiler could remind us to remove the dead branch. However, we can achieve this without pattern matching too by turning `v < V1` into `v <= V0`.

On the contrary, the additional indirection IMO communicates the intent less clearly (there was a change in V1) and adds new ways of making a mistake (e.g. using V0's decoding in V2 again, especially when the difference is just an apostrophe).

jorisdral (approved):
LGTM! 

I think we should probably mention in the changelog that we added a new snapshot version. I wonder if we should also list it as a breaking change, because the snapshot format has changed... ideas?

mheinzel:
Good point. While the new version can still read old snapshots, we should make it clear that upgrading to the new version of `lsm-tree`, creating a snapshot and then going back to the previous version doesn't work. Not sure if this should be considered breaking, but I'd rather err on the side of caution.

lsm-tree/CHANGELOG.md
```diff
@@ -1,5 +1,27 @@
...
 
+## Unreleased
+
+### Breaking changes
+
+* Move to snapshot version v2 due to changes to the internal table
+  representation. While this change is backwards compatible (i.e. new code
+  is still able to read old snapshots), be aware that v2 snapshots cannot be
+  read by older versions of `lsm-tree`.
+  See [PR #834](https://github.com/IntersectMBO/lsm-tree/pull/834).
```
jorisdral, line 11: :+1:

PR #835, opened 2026-04-06
Support exporting and importing snapshots

### LINKED ISSUE #272: Allow storing snapshots outside of the session directory

jorisdral:
Partially resolves #272 

This enables storing snapshots in directories other than the session directory. For now, exports are only supported for destination directories that are on the same disk volume.

This also does not yet support sharing snapshot directories between different sessions, as suggested in #272.

### DIFF
--- a/lsm-tree/app/Database/LSMTree/Demo.hs
+++ b/lsm-tree/app/Database/LSMTree/Demo.hs
@@ -24,6 +24,7 @@ import           Data.Word (Word64)
 import           Database.LSMTree as LSMT
 import qualified System.Directory as IO (createDirectoryIfMissing,
                      doesDirectoryExist, removeDirectoryRecursive)
+import           System.FilePath ((</>))
 import qualified System.FS.API as FS
 import qualified System.FS.BlockIO.API as FS
 import qualified System.FS.BlockIO.IO as FS
@@ -41,8 +42,8 @@ import           System.IO.Unsafe (unsafePerformIO)
 -- functional requirement.
 demo :: Bool -> IO ()
 demo interactive = do
-  freshDirectory "_demo"
-  withOpenSessionIO tracer "_demo" $ \session -> do
+  freshDirectory ("_demo" </> "session")
+  withOpenMountedSessionIO tracer "_demo" (FS.mkFsPath ["session"]) $ \session -> do
     withTableWith config session  $ \(table :: Table IO K V B) -> do
       pause interactive -- [0]
 
@@ -141,8 +142,8 @@ demo interactive = do
          (LSMT.IOLike m, Typeable h)
       => FS.HasFS m h -> FS.HasBlockIO m h -> m ()
     simpleAction hasFS hasBlockIO = do
-      let sessionDir = FS.mkFsPath ["_demo"]
-      FS.createDirectoryIfMissing hasFS False sessionDir
+      let sessionDir = FS.mkFsPath ["_demo", "session"]
+      FS.createDirectoryIfMissing hasFS True sessionDir
       withOpenSession tracer hasFS hasBlockIO 17 sessionDir $ \session -> do
         withTableWith config session  $ \(table :: Table m K V B) -> do
           inserts table $ V.fromList [ (K i, V i, Just (B i)) | i <- [1 .. 10_000] ]
@@ -215,7 +216,7 @@ freshDirectory :: FilePath -> IO ()
 freshDirectory path = do
     b <- IO.doesDirectoryExist path
     when b $ IO.removeDirectoryRecursive path
-    IO.createDirectoryIfMissing False path
+    IO.createDirectoryIfMissing True path
 
 print' :: (Show a, MonadST m) => a -> m ()
 print' x = stToIO $ unsafeIOToST $ print x
--- a/lsm-tree/src-core/Database/LSMTree/Internal/FS.hs
+++ b/lsm-tree/src-core/Database/LSMTree/Internal/FS.hs
@@ -1,0 +1,72 @@
+module Database.LSMTree.Internal.FS (
+    -- * Hard links
+    hardLink
+    -- * Copy file
+  , copyFile
+  ) where
+
+import           Control.ActionRegistry
+import           Control.Monad (void)
+import           Control.Monad.Class.MonadThrow
+import           Control.Monad.Primitive (PrimMonad)
+
+import qualified System.FS.API as FS
+import           System.FS.API
+import qualified System.FS.API.Lazy as FSL
+import qualified System.FS.BlockIO.API as FS
+import           System.FS.BlockIO.API (HasBlockIO)
+
+{-------------------------------------------------------------------------------
+  Hard links
+-------------------------------------------------------------------------------}
+
+{-# SPECIALISE
+  hardLink ::
+       HasFS IO h
+    -> HasBlockIO IO h
+    -> ActionRegistry IO
+    -> FS.FsPath
+    -> FS.FsPath
+    -> IO ()
+  #-}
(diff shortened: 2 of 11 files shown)


lsm-tree/app/Database/LSMTree/Demo.hs
```diff
@@ -3,6 +3,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 {-# OPTIONS_GHC -Wno-unused-do-bind #-}
+{-# OPTIONS_GHC -Wwarn=deprecations #-}
```
jorisdral:
```suggestion
```

lsm-tree/src-core/Database/LSMTree/Internal/Unsafe.hs
```diff
@@ -243,6 +249,18 @@ data SessionTrace =
     -- | We are listing snapshots.
   | TraceListSnapshots
 
+    -- | We are importing a snapshot. A 'TraceImportedSnapshot' message should
+    -- follow if successful.
+  | TraceImportSnapshot SnapshotName FsPath
+    -- | Importing a snapshot was successful.
+  | TraceImportedSnapshot SnapshotName
+
+    -- | We are exported a snapshot. A 'TraceExportedSnapshot' message should
```
jorisdral:
```suggestion
    -- | We are exporting a snapshot. A 'TraceExportedSnapshot' message should
```

lsm-tree/src-core/Database/LSMTree/Internal/FS.hs
```diff
@@ -0,0 +1,112 @@
...
+  hardLink ::
+       HasFS IO h
+    -> HasBlockIO IO h
+    -> ActionRegistry IO
+    -> FS.FsPath
+    -> FS.FsPath
+    -> IO ()
+  #-}
+-- | @'hardLink' hfs hbio reg sourcePath destinationPath@ creates a hard link from
+-- @sourcePath@ to @targetPath@.
```
mheinzel:
```suggestion
-- | @'hardLink' hfs hbio reg sourcePath destinationPath@ creates a hard link from
-- @sourcePath@ to @destinationPath@.
```

lsm-tree/src-core/Database/LSMTree/Internal/FS.hs
```diff
@@ -0,0 +1,112 @@
...
+
+{-# SPECIALISE
+  copyFile ::
+       HasFS IO h
+    -> ActionRegistry IO
+    -> FS.FsPath
+    -> FS.FsPath
+    -> IO ()
+  #-}
+-- | @'copyFile' hfs reg source target@ copies the @source@ path to the @target@ path.
```
mheinzel:
```suggestion
-- | @'copyFile' hfs reg sourcePath destinationPath@ copies the @sourcePath@ to the @destinationPath@.
```

For consistency with `hardLink`.

lsm-tree/src/Database/LSMTree.hs
```diff
@@ -397,6 +415,26 @@ Run an action with access to a session opened from a session directory.
 If the session directory is empty, a new session is created using the given salt.
 Otherwise, the session directory is restored as an existing session ignoring the given salt.
 
+As a feature, @lsm-tree@ uses abstract 'HasFS' and 'HasBlockIO' interfaces to
+abstract over interaction with the file system.
+These abstract interfaces can be instantiated with a real file system (for use in practice) or
+a simulated file system (for comprehensive testing).
+For more information about the interfaces and its instantiations, see the Haddock
```
mheinzel:
```suggestion
For more information about the interfaces and their instantiations, see the Haddock
```

lsm-tree/src/Database/LSMTree.hs
```diff
@@ -2810,6 +2886,145 @@ listSnapshots ::
 listSnapshots (Session session) =
   Internal.listSnapshots session
 
+
+{- |
+Import a snapshot from an external directory.
+
+Importing does not not check whether the external directoy is a snapshot, and
```
mheinzel:
```suggestion
Importing does not not check whether the external directory is a snapshot, and
```

lsm-tree/src/Database/LSMTree.hs
```diff
@@ -2810,6 +2886,145 @@ listSnapshots ::
...
+relative to a /root/.
+What the root is depends on which function was used to create the session.
+See 'withOpenSession', 'withOpenSessionIO', and 'withOpenMountedSessionIO' for more
+information about the root.
+
+It is generally not advisable to use 'importSnapshot' and 'exportSnapshot' when the session is
+created using 'withOpenSessionIO'.
+Snapshots should only be exported to somewhere /outside/ the session directory, which is not possible
+when the session is created using 'withOpenSessionIO'.
+Use 'withOpenMountedSessionIO' or 'withOpenSession' instead.
```
mheinzel, line 2979: I think we should mention that the destination must be on the same mounted file system. Maybe that is expected from how `fs-api` works, but I'd rather be explicit about it. It might also make sense to leave a short note in the docs for `LSMTree.Internal.FS.{hardLink,hardLinkDirectoryRecursive}` and `BlockIO.API.createHardLink`.

Similarly, we should note that the destination directory must not exist yet (which should make it clear, that no two sessions can export to the same snapshot directory).

lsm-tree/src-core/Database/LSMTree/Internal/Unsafe.hs
```diff
@@ -1895,6 +1913,110 @@ listSnapshots sesh = do
...
+        -- succesfully performed
+        delayedCommit reg $
+          traceWith sesh.sessionTracer $ TraceImportedSnapshot snap
+
+-- | A snapshot was intended to be exported, but the destination directory
+-- already exists.
+data SnapshotExportDirExistsError
+    = SnapshotExportDirExistsError !FsPath
+    deriving stock (Show, Eq)
+    deriving anyclass (Exception)
```
mheinzel, line 1974: On the topic of errors, do you know how a POSIX `EXDEV` error (hardlinking across mounted filesystems) bubbles up through `fs-api`/`blockio` to here? Should we handle it somewhere and report a better error?

mheinzel (commented):
> For now, exports are only supported for destination directories that are on the same disk volume.

> This also does not yet support sharing snapshot directories between different sessions, as suggested in https://github.com/IntersectMBO/lsm-tree/issues/272.

It's okay to add the feature with these limitations, but we should document them in the source code as well.

Apart from that (and some minor suggestions for typos etc.) this looks good.

I think it's reasonable to have this two-step process of first saving then exporting a snapshot (and similarly importing and opening it). It means that `exportSnapshot` is quite simple and once we allow crossing file system boundaries, we only need to consider that there, not in the code that traverses tables to assemble the snapshot. 👍
Out of interest, were there other reasons or requirements that led to that design vs. directly creating an external snapshot?

jorisdral:
> Out of interest, were there other reasons or requirements that led to that design vs. directly creating an external snapshot?

Mainly for simplicity, so that there's not much I have to change about the existing snapshot code. Implementing shared snapshot directories would have required some more thinking about concurrency amongst other thing. But exporting/importing is useful already, and simple to implement!

lsm-tree/src-core/Database/LSMTree/Internal/Unsafe.hs
jorisdral: I'm not sure! I've never tried, and it's possibly a little hard to test.

I do know that any `IOError` in the implementation of `ioHasFS` will be rethrown as an `FsError`. For `ioHasBlockIO` I don't think we are doing that. 

I'll open an issue to investigate.

mheinzel (approved):

PR #838, opened 2026-04-21
Migrate the union level once the merging tree is completed

mheinzel:
~~Building on top of #834 #840, which we should merge first. Apart from that,~~ this PR is ready for review.

Applies the idea outlined in #824 to the real implementation.

### DIFF
--- a/lsm-tree/src/Database/LSMTree.hs
+++ b/lsm-tree/src/Database/LSMTree.hs
@@ -2052,6 +2052,13 @@ _withInternalTables (Table (table :: Internal.Table m h) :| tables) action =
 Get an /upper bound/ for the amount of remaining union debt.
 This includes the union debt of any table that was part of the union's input.
 
+The exact amount of debt returned for a given table is an implementation detail
+and not guaranteed to be stable across versions of @lsm-tree@.
+Within the same version, the union debt of a table can never increase. However,
+it can get decreased by operations on /another/ table if these tables share part
+of their internal structure (e.g. due to one being a union input or duplicate of
+the other).
+
 >>> :{
 runExample $ \session table1 -> do
   LSMT.insert table1 0 "Hello" Nothing
@@ -2061,7 +2068,7 @@ runExample $ \session table1 -> do
     bracket (LSMT.incrementalUnion table1 table2) LSMT.closeTable $ \table3 -> do
       putStrLn . ("UnionDebt: "<>) . show =<< LSMT.remainingUnionDebt table3
 :}
-UnionDebt: 4
+UnionDebt: 5
 
 The worst-case disk I\/O complexity of this operation is \(O(0)\).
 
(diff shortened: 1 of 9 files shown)


lsm-tree/test/Test/Database/LSMTree/UnitTests.hs
```diff
@@ -204,7 +205,7 @@ unit_union_credit_0 =
     withTable @_ @Key1 @Value1 @Blob1 sess $ \table -> do
       inserts table [(Key1 17, Value1 42, Nothing)]
 
-      bracket (table `union` table) closeTable $ \table' -> do
+      bracket (table `incrementalUnion` table) closeTable $ \table' -> do
```
jorisdral, line 208: Hmm, did we forget to change `union` to `incrementalUnion` at some point? Good catch!

lsm-tree/test/Test/Database/LSMTree/UnitTests.hs
```diff
@@ -253,6 +304,9 @@ assertException e assertion = do
    r <- try assertion
    r @?= Left e
 
+(@>=) :: (HasCallStack, Ord a, Show a) => a -> a -> Assertion
+x @>= y = (x >= y) @? ("expected: " ++ show x ++ " > " ++ show y)
```
jorisdral:
```suggestion
x @>= y = (x >= y) @? ("expected: " ++ show x ++ " >= " ++ show y)
```

lsm-tree/src/Database/LSMTree.hs
```diff
@@ -1985,7 +1985,7 @@ runExample $ \session table1 -> do
     bracket (LSMT.incrementalUnion table1 table2) LSMT.closeTable $ \table3 -> do
       putStrLn . ("UnionDebt: "<>) . show =<< LSMT.remainingUnionDebt table3
 :}
-UnionDebt: 4
+UnionDebt: 5
```
jorisdral, line 2071: What's your opinion: should we mark it as a breaking change if union debt changes between releases of `lsm-tree`? My current thinking is: no, that is an implementation detail, and users should not expect it to be stable. But maybe we should be extra clear about this in the Haddocks of the public API? If you agree, maybe we should update the documentation in this PR as well.

jorisdral (approved):
Very nice :)

lsm-tree/src/Database/LSMTree.hs
mheinzel: I agree. I will update the docs.

lsm-tree/test/Test/Database/LSMTree/UnitTests.hs
mheinzel: Yes, looks like it.

lsm-tree/src/Database/LSMTree.hs
mheinzel: Done.

PR #841, opened 2026-04-23
Fix a typo in a bit-mask in `unpackRunIxKeyIx`

jorisdral:
When calling `Database.LSMTree.lookups` with large-sized batches of keys, @jasagredo was seeing unexpected `NotFound` lookup results. The offending code is a typo in a bit-mask that we use while querying bloom filters. This typo effectively limited the number of bloom queries we could perform on a single run to `4096` (hexadecimal: `0xfff`),  while the limit should have been `65536` (hexadecimal: `0xffff`). So under high `lookups` workloads, this limit would be hit and hence the erroneous lookup results.

### DIFF
--- a/lsm-tree/test/Test/Database/LSMTree/Internal/BloomFilter.hs
+++ b/lsm-tree/test/Test/Database/LSMTree/Internal/BloomFilter.hs
@@ -184,7 +184,9 @@ prop_bloomQueries (FPR fpr) filters keys =
 --
 -- There used to be a bug where this went wrong because of a typo in a bit-mask.
 -- This property test should ensure that we catch such mistakes in the future.
--- See PR #
+-- See PR #841 for more information.
+--
+-- <PR https://github.com/IntersectMBO/lsm-tree/pull/841>
 --
 prop_packUnpack_RunIxKeyIx :: Int_0xffff -> Int_0xffff -> Property
 prop_packUnpack_RunIxKeyIx r k =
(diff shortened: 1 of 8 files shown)


jorisdral:
> When calling `Database.LSMTree.lookups` with large-sized batches of keys, @jasagredo was seeing unexpected `NotFound` lookup results. The offending code is a typo in a bit-mask that we use while querying bloom filters. This typo effectively limited the number of bloom queries we could perform on a single run to `4096` (hexadecimal: `0xfff`), while the limit should have been `65536` (hexadecimal: `0xffff`). So under high `lookups` workloads, this limit would be hit and hence the erroneous lookup results.

Also: this limit should probably not be implicit. Instead, `lookups` should return an error if the number of input keys ever exceeds the limit of what we can handle in `lsm-tree`.

lsm-tree/test/Test/Database/LSMTree/Internal/BloomFilter.hs
```diff
@@ -168,3 +170,50 @@ prop_bloomQueries (FPR fpr) filters keys =
...
+--
+-- More specifically, if we apply a 'RunIxKeyIx' pattern synonym to a pair of
+-- 'RunIx' and 'KeyIx', and then pattern match on it agains, then we would
+-- expect to get the same 'RunIx' and 'KeyIx' out as the ones we put in.
+--
+-- There used to be a bug where this went wrong because of a typo in a bit-mask.
+-- This property test should ensure that we catch such mistakes in the future.
+-- See PR #841 for more information.
+--
+-- <PR https://github.com/IntersectMBO/lsm-tree/pull/841>
```
mheinzel, line 189: Does that syntax work in Haddock? At least it's not documented in the [Haddock Markup docs](https://haskell-haddock.readthedocs.io/latest/markup.html#urls).

mheinzel (commented):
Looks good. Good idea to have roundtrip tests for bit manipulation code.

lsm-tree/test/Test/Database/LSMTree/Internal/BloomFilter.hs
jorisdral: Oh having PR in there was typo. I just intended to have the link between `<` `>` and that would be just rendered as a link

PR #845, opened 2026-04-24
Update to fs-sim 0.5.0.0

jasagredo:
# Description

`ouroboros-consensus` needs this new fs-sim version, which has breaking changes:

```
### Breaking

* Make `sim` operations exception safe by using `withMockFS`.
* Require `MonadMask` in `simHasFS`, `simHasFS'`, `runSimFS`.
```

# Checklist

### DIFF
--- a/blockio/CHANGELOG.md
+++ b/blockio/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Revision history for blockio
 
+## Next version -- unreleased
+
+### Breaking changes
+
+* Update to `fs-sim ^>=0.5`. See PR
+  [#845](https://github.com/IntersectMBO/lsm-tree/pull/845).
+
+### New features
+
+None
+
+### Minor changes
+
+None
+
+### Bug fixes
+
+None
+
 ## 0.1.1.2 -- 2026-04-24
 
 ### Breaking changes
--- a/blockio/blockio.cabal
+++ b/blockio/blockio.cabal
@@ -149,7 +149,7 @@ library sim
     , blockio
     , bytestring             ^>=0.11 || ^>=0.12
     , fs-api                 ^>=0.4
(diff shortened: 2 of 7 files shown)


blockio/CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
...
+None
+
+### New features
+
+None
+
+### Minor changes
+
+* Update to `fs-sim ^>=0.5`. See PR
+  [#845](https://github.com/IntersectMBO/lsm-tree/pull/845).
```
jorisdral: I think in this case it should be listed as a breaking change, since the constraints for some of the exported functions in the *public* `blockio:sim` library have changed

blockio/blockio.cabal
```diff
@@ -149,7 +149,7 @@ library sim
     , blockio
     , bytestring             ^>=0.11 || ^>=0.12
     , fs-api                 ^>=0.4
-    , fs-sim                 ^>=0.4
+    , fs-sim                 ^>=0.4 || ^>=0.5
```
jorisdral: I like that we can still use `fs-sim-0.4` :smile: 

Did you try to build with `fs-sim-0.4` after your changes? It probably complains about redundant constraints if so, which is fine, but it would be good to double check that we didn't overlook anything else

jorisdral (approved):
Nice :smile:

blockio/blockio.cabal
jasagredo: No warnings, this succeeds just fine:
```
❯ cabal build all --constraint="fs-sim<0.5"
```

blockio/CHANGELOG.md
jasagredo: Moved to breaking

PR #847, opened 2026-05-10
blockio: Use osx implementation on freebsd

arrowd:
`cabal test` passes.

### DIFF
--- a/blockio/blockio.cabal
+++ b/blockio/blockio.cabal
@@ -103,7 +103,7 @@ library
       other-modules: System.FS.BlockIO.Async
       build-depends: blockio-uring ^>=0.2
 
-  elif os(osx)
+  elif os(osx) || os(freebsd)
     hs-source-dirs: src-macos
     build-depends:  unix ^>=2.8.4
     other-modules:


jorisdral:
Thank you for your contribution!

I'm not positive that the OSX implementation is correct for FreeBSD. Some of the code (like the `Fcntl.hsc` module) assumes that the platform is OSX. I think the way to go would probably be to add a dedicated fallback implementation for Unix that is used when the platform is not specifically Linux, OSX, or Windows. I'm going to think a bit on this, and I'll get back to you @arrowd.

Out of interest: are you using `lsm-tree`/`blockio`, and is that why you want them to support FreeBSD?

arrowd:
osx code at https://github.com/IntersectMBO/lsm-tree/blob/main/blockio/src-macos/System/FS/BlockIO/Internal.hs#L16-L20 says

> -- For now we use the portable serial implementation of HasBlockIO. If you
-- want to provide a proper async I\/O implementation for OSX, then this is where
-- you should put it.
--
-- The recommended choice would be to use the POSIX AIO API.

So this is 100% POSIX implementation, which is why it works on FreeBSD out of the box. Unfortunately, I don't have time to implement aio-based POSIX backend myself.

> Out of interest: are you using lsm-tree/blockio, and is that why you want them to support FreeBSD?

Yes, this package pops out as a cardano-node dependency.

jorisdral:
> So this is 100% POSIX implementation, which is why it works on FreeBSD out of the box. Unfortunately, I don't have time to implement aio-based POSIX backend myself.

Sorry, I didn't mean we had to implement a whole backend. I only meant to say that there are some small changes we have to make to the OSX implementation to make it correct for unix-based systems in general. I'll do that when I can find some time

arrowd:
Any movement on this?

jorisdral:
No, unfortunately I have limited time to work on `lsm-tree`. Are you blocked on this PR?

arrowd:
No, this can wait. I'm using changes from this PR as a temporary solution.

jorisdral:
Alright, it's good to hear you are currently not blocked. I have created issue #872 to track that we should support other platforms in `blockio`, and I'll close this PR for now. Is that okay with you?

arrowd:
Yes, absolutely, thank you.

PR #852, opened 2026-06-11
Export also the salt and expose function to create a new session with it

jasagredo:
# Description

In order to import an exported snapshot, we need the same salt that was used when exporting.

# Checklist


jorisdral:
Will be superseded by #855

PR #870, opened 2026-07-27
Fix `importSnapshot`/`exportSnapshot`

### LINKED ISSUE #853: [FEATURE] Add variant of withOpenMountedSessionIO without FsPath
**Is your feature request related to a problem? Please describe.**
When designing the public API for `lam-tree`, the intended purpose of `withOpenSessionIO` was to allow users to use the `lsm-tree` library without having to add a direct dependency on the `fs-api` package. Unfortunately, the new `withOpenMountedSessionIO` function breaks that purpose, meaning its no longer possible to use `lsm-tree` without managing a direct dependency on `fs-api`.

**Describe the solution you'd like**
I'd like to be able to use `lsm-tree` without a direct dependency on `fs-api`. One possibility would be to revert the deprecation of `withOpenSessionIO` and add variants of `importSnapshot` and `exportSnapshot` that take `FilePath` arguments with a documented side-condition that these *must* be relative to the database directory. Alternatively, a variant of `withOpenMountedSessionIO` could be given two `FilePath` arguments, with the side-condition that the second `FilePath` must either be relative or absolute with the former as a prefix. The variants of `importSnapshot` and `exportSnapshot` could then enforce the more general side-condition that their `FilePath` arguments must be within the mount point.

Is there a technical reason why snapshots *must* be imported from and exported to paths within the same mount point?

wenkokke:
This PR addresses #853. It makes the following changes:

- Revert the deprecation of `withOpenSessionIO`.
- **BREAKING**: Replace the `FsPath` field of `SnapshotImportDirDoesNotExistError`/`SnapshotExportDirExistsError` with an `FsErrorPath`.
- **BREAKING**: Rename the constructors of `SnapshotImportDirDoesNotExistError`/`SnapshotExportDirExistsError` to `ErrSnapshotImportDirDoesNotExist`/`ErrSnapshotExportDirExists` for consistency with the rest of the exceptions defined by the `lsm-tree` package.
- **BREAKING**: Add an optional `HasFS` instance to the `FsPath` argument in `importSnapshot`/`exportSnapshot`. If the `HasFS` instance is provided, the `FsPath` path is interpreted as a path in the corresponding filesystem, and the snapshot is _always copied_. If the `HasFS` instance is not provided, the `FsPath` path is interpreted as a path in the filesystem for the database session, and the snapshot is _hard linked_ with a fallback to copying.
- Add the `importSnapshotIO`/`exportSnapshotIO` functions to `Database.LSMTree`, which accept a `FilePath` argument and always copy the snapshot files.
- Add the `importSnapshot`/`exportSnapshot` functions to `Database.LSMTree.Simple`, which in turn call the `importSnapshotIO`/`exportSnapshotIO` functions and correctly repackage any exceptions.
- Fix a bug where the `SnapshotExportDirExistsError` would contain the internal snapshot path.

### DIFF
--- a/lsm-tree/CHANGELOG.md
+++ b/lsm-tree/CHANGELOG.md
@@ -1,5 +1,105 @@
 # Revision history for `lsm-tree`
 
+## 1.2.0.0 -- 2026-07-27
+
+### Breaking changes
+
+* The constructor for `SnapshotImportDirDoesNotExistError` was renamed to
+  `ErrSnapshotImportDirDoesNotExist` and its field is now of type `FsErrorPath`.
+
+* The constructor for `SnapshotExportDirExistsError` was renamed to
+  `ErrSnapshotExportDirExists`  and its field is now of type `FsErrorPath`.
+
+* The type of `importSnapshot` was changed from...
+
+  ```hs
+  importSnapshot ::
+    forall m h.
+    (IOLike m) =>
+    Session m ->
+    SnapshotName ->
+    FsPath ->
+    m ()
+  ```
+
+  ...to...
+
+  ```hs
+  importSnapshot ::
+    forall m h.
+    (IOLike m) =>
+    Session m ->
+    SnapshotName ->
+    (Maybe (HasFS m h), FsPath) ->
+    m ()
+  ```
+
+  In the previous release, the source directory was passed as an `FsPath`,
+  which was interpreted relative to the session mount point. From this release
+  onwards, it is passed as a pair of an `FsPath` with an optional `HasFS`
+  instance. If the `HasFS` instance is provided, the `FsPath` path is
+  interpreted as a path in the corresponding filesystem, and the snapshot is
+  always copied. If the `HasFS` instance is not provided, the `FsPath` path is
+  interpreted relative to the session mount point, as before, and the snapshot
+  is hard linked with a fallback to copying.
+
+  Likewise, the type of `exportSnapshot` was changed from...
+
+  ```hs
+  exportSnapshot ::
+    forall m h.
+    (IOLike m) =>
+    Session m ->
+    SnapshotName ->
+    FsPath ->
+    m ()
+  ```
+
+  ...to...
+
+  ```hs
+  exportSnapshot ::
+    forall m h.
+    (IOLike m) =>
+    Session m ->
+    SnapshotName ->
+    (Maybe (HasFS m h), FsPath) ->
+    m ()
+  ```
+
+  The change in the type of the destination directory has the same
+  interpretation as for `importSnapshot`.
+
+### New features
+
+#### Full API
+
+* Add a new `importSnapshotIO` function that imports snapshots from disk using
+  a `FilePath` path.
+
+* Add a new `exportSnapshotIO` function that exports snapshots to disk using
+  a `FilePath` path.
+
+#### Simple API
+
+* Add a new `importSnapshot` function that imports snapshots from disk using
+  a `FilePath` path and a variant of `SnapshotImportDirDoesNotExistError` with
+  a `FilePath` field.
+
+* Add a new `exportSnapshot` function that exports snapshots to disk using
+  a `FilePath` path and a variant of `SnapshotExportDirExistsError` with a
+  `FilePath` field.
+
+### Minor changes
+
+* Revert the deprecation of `withOpenSessionIO`, since the other changes in this
+  version have made it safe to use with `importSnapshot` and `exportSnapshot`.
+
+### Bug fixes
+
+* If an `SnapshotExportDirExistsError` is thrown, this now contains the
+  destination directory, rather than the directory of the internap snapshot.
+
 ## 1.1.1.0 -- 2026-07-21
 
 ### Breaking changes
--- a/lsm-tree/CHANGELOG.md
+++ b/lsm-tree/CHANGELOG.md
@@ -1,94 +1,38 @@
 # Revision history for `lsm-tree`
 
-## 1.2.0.0 -- 2026-07-27
+## ?.?.?.? -- ????-??-??
 
 ### Breaking changes
 
 * The constructor for `SnapshotImportDirDoesNotExistError` was renamed to
   `ErrSnapshotImportDirDoesNotExist` and its field is now of type `FsErrorPath`.
-
 * The constructor for `SnapshotExportDirExistsError` was renamed to
   `ErrSnapshotExportDirExists`  and its field is now of type `FsErrorPath`.
-
-* The type of `importSnapshot` was changed from...
-
-  ```hs
-  importSnapshot ::
-    forall m h.
-    (IOLike m) =>
-    Session m ->
-    SnapshotName ->
-    FsPath ->
-    m ()
-  ```
-
-  ...to...
-
-  ```hs
-  importSnapshot ::
-    forall m h.
-    (IOLike m) =>
-    Session m ->
-    SnapshotName ->
-    (Maybe (HasFS m h), FsPath) ->
-    m ()
-  ```
-
-  In the previous release, the source directory was passed as an `FsPath`,
-  which was interpreted relative to the session mount point. From this release
-  onwards, it is passed as a pair of an `FsPath` with an optional `HasFS`
-  instance. If the `HasFS` instance is provided, the `FsPath` path is
-  interpreted as a path in the corresponding filesystem, and the snapshot is
-  always copied. If the `HasFS` instance is not provided, the `FsPath` path is
-  interpreted relative to the session mount point, as before, and the snapshot
-  is hard linked with a fallback to copying.
-
-  Likewise, the type of `exportSnapshot` was changed from...
-
-  ```hs
-  exportSnapshot ::
-    forall m h.
-    (IOLike m) =>
-    Session m ->
-    SnapshotName ->
-    FsPath ->
-    m ()
-  ```
-
-  ...to...
-
-  ```hs
-  exportSnapshot ::
-    forall m h.
-    (IOLike m) =>
-    Session m ->
-    SnapshotName ->
-    (Maybe (HasFS m h), FsPath) ->
-    m ()
-  ```
-
-  The change in the type of the destination directory has the same
-  interpretation as for `importSnapshot`.
+* `importSnapshot` and `exportSnapshot` have a new `SnapshotMode` argument that
+  determines whether the snapshot is imported/exported by hard linking or
+  copying. If the `SnapshotMode` is `HardLink fallback`, the `FsPath` is
+  interpreted relative to the session's *interface root* (see haddocks). The
+  `fallback` flag determines whether the import/export should fall back to
+  copying if hard linking fails. If the `SnapshotMode` is `Copy extFS`, the
+  `FsPath` is interpreted relative to the root of the `extFS` `HasFS` interface.
 
 ### New features
 
 #### Full API
 
-* Add a new `importSnapshotIO` function that imports snapshots from disk using
-  a `FilePath` path.
-
-* Add a new `exportSnapshotIO` function that exports snapshots to disk using
-  a `FilePath` path.
+* Add a new `importSnapshotIO` function that imports snapshots from disk by
+  using a `FilePath` path. This function always copies rather than hard linking.
+* Add a new `exportSnapshotIO` function that exports snapshots to disk using a
+  `FilePath` path. This function always copies rather than hard linking.
 
 #### Simple API
 
-* Add a new `importSnapshot` function that imports snapshots from disk using
-  a `FilePath` path and a variant of `SnapshotImportDirDoesNotExistError` with
-  a `FilePath` field.
-
-* Add a new `exportSnapshot` function that exports snapshots to disk using
-  a `FilePath` path and a variant of `SnapshotExportDirExistsError` with a
-  `FilePath` field.
+* Add a new `importSnapshot` function that imports snapshots from disk using a
+  `FilePath` path and a variant of `SnapshotImportDirDoesNotExistError` with a
+  `FilePath` field. This function always copies rather than hard linking.
+* Add a new `exportSnapshot` function that exports snapshots to disk using a
+  `FilePath` path and a variant of `SnapshotExportDirExistsError` with a
+  `FilePath` field. This function always copies rather than hard linking.
 
 ### Minor changes
 
@@ -97,8 +41,8 @@
 
 ### Bug fixes
 
-* If an `SnapshotExportDirExistsError` is thrown, this now contains the
-  destination directory, rather than the directory of the internap snapshot.
+* If a `SnapshotExportDirExistsError` is thrown, this now contains the
+  destination directory, rather than the directory of the internal snapshot.
 
 ## 1.1.1.0 -- 2026-07-21
 
--- a/lsm-tree/src-core/Database/LSMTree/Internal/FS.hs
+++ b/lsm-tree/src-core/Database/LSMTree/Internal/FS.hs
@@ -4,6 +4,7 @@ module Database.LSMTree.Internal.FS (
   , hardLinkDirectoryRecursive
     -- * Copy file
   , copyFile
+    -- * Hard links with fallback
   , hardLinkOrCopyDirectoryRecursive
   ) where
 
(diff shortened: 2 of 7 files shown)


lsm-tree/src-core/Database/LSMTree/Internal/FS.hs
```diff
@@ -109,9 +111,113 @@ copyFile ::
...
+  -> m ()
+copyFile' sourceFS destinationFS reg sourcePath destinationPath =
+    flip (withRollback_ reg) (FS.removeFile destinationFS destinationPath) $
+      FS.withFile sourceFS sourcePath FS.ReadMode $ \sourceHandle ->
+        FS.withFile destinationFS destinationPath (FS.WriteMode FS.MustBeNew) $ \destinationHandle -> do
+          bs <- FSL.hGetAll sourceFS sourceHandle
+          void $ FSL.hPutAll destinationFS destinationHandle bs
+
+{-------------------------------------------------------------------------------
+  Copy file
```
jorisdral:
```suggestion
 Hard link or copy
```

lsm-tree/src-core/Database/LSMTree/Internal/FS.hs
```diff
@@ -109,9 +111,113 @@ copyFile ::
...
+    -> Either (HasBlockIO IO h) (HasFS IO h')
+    -> ActionRegistry IO
+    -> FS.FsPath
+    -> FS.FsPath
+    -> IO ()
+  #-}
+-- | @'hardLinkOrCopy' sourceFS destinationFS hbio reg sourcePath destinationPath@
+--   attemtps to create a hard link from @sourcePath@ to @destinationPath@, if
+--   both are on the same file system and copies the file otherwise.
+hardLinkOrCopy ::
```
jorisdral:
```suggestion
-- | @'hardLinkOrCopy' sourceFS destinationFS hbio reg sourcePath destinationPath@
--   attemtps to create a hard link from @sourcePath@ to @destinationPath@ if
--   both are on the same file system and copies the file otherwise.
hardLinkOrCopy ::
```

lsm-tree/src/Database/LSMTree/Simple.hs
jorisdral, line 1: When I wrote the first versions of the import/export functions I figured it would just be easier to not include the import/export functions to the `Simple` module, unless someone explicitly asked for them. Or do you think these functions are useful enough that we should put them in the `Simple` module?

lsm-tree/src/Database/LSMTree/Simple.hs
```diff
@@ -1560,6 +1564,62 @@ listSnapshots ::
 listSnapshots (Session session) =
     LSMT.listSnapshots session
 
+{- |
+Import a snapshot from an external directory.
+
+If the source directory is on a different filesystem from the session
+directory, it should be passed as a pair of the `HasFS` instance for that
+filesystem and an `FsPath`. If the source directory is on the same
+filesystem as the session directory, the `HasFS` instance may be omitted.
```
jorisdral: The `Simple` module does not deal with `HasFS`

lsm-tree/src/Database/LSMTree/Simple.hs
```diff
@@ -1560,6 +1564,62 @@ listSnapshots ::
...
+Throws the following exceptions:
+
+['SessionClosedError']:
+    If the session is closed.
+['SnapshotDoesNotExistError']:
+    If no snapshot with the given name exists.
+['SnapshotExportDirExistsError']:
+    If the destination directory for the to-be-exported snapshot already exists.
+-}
+exportSnapshot ::
```
jorisdral, line 1609: Maybe we should mention that this copies the full snapshot

lsm-tree/src/Database/LSMTree/Simple.hs
```diff
@@ -1560,6 +1564,62 @@ listSnapshots ::
...
+Throws the following exceptions:
+
+['SessionClosedError']:
+    If the session is closed.
+['SnapshotExistsError']:
+    If a snapshot with the same name already exists.
+['SnapshotImportDirDoesNotExistError']:
+    If the source directory for the to-be-imported snapshot does not exist.
+-}
+importSnapshot ::
```
jorisdral, line 1583: Maybe we should mention that this copies the full snapshot

lsm-tree/test/Test/Database/LSMTree/Snapshots.hs
```diff
@@ -67,8 +67,8 @@ prop_exportImportSnapshot ins los =
           saveSnapshot "snap1" "KeyValueBlob" table1
 
           -- Export then re-import the snapshot
-          exportSnapshot session "snap1" exportDir
-          importSnapshot session "snap2" exportDir
+          exportSnapshot session "snap1" (Nothing, exportDir)
+          importSnapshot session "snap2" (Nothing, exportDir)
```
jorisdral: let's also test the copying version. Maybe 50/50 pass in `Just hfs`?

lsm-tree/src/Database/LSMTree.hs
```diff
@@ -3022,20 +3014,47 @@ Throws the following exceptions:
...
+  -- | The destination directory and optionally a 'HasFS' instance,
+  --   if the destination directory is on a different filesystem.
+  (Maybe (HasFS m h), FsPath) ->
   m ()
 exportSnapshot (Session session) =
     Internal.exportSnapshot session
 
+{- |
+Variant of 'exportSnapshot' that is specialised to 'IO' using the real filesystem.
+-}
```
jorisdral, line 3041: Maybe we should mention that it copies snapshot contents?

lsm-tree/src/Database/LSMTree.hs
```diff
@@ -3022,20 +3014,47 @@ Throws the following exceptions:
...
+
+  -- NOTE: If exportRelDir is null, then exportAbsDir must be the root
+  --       directory, which, it is fair to assume, exists.
+  when (null exportRelDir) $ do
+    let exportFsErrorPath = FsErrorPath (Just mountPoint) exportDirFsPath
+    throwIO $ ErrSnapshotExportDirExists exportFsErrorPath
+
+  -- Export the snapshot.
+  exportSnapshot session snapshotName (Just (ioHasFS @IO mountPoint), exportDirFsPath)
+
```
jorisdral, line 3057: Right, `splitFileName` can produce an empty `exportRelDir`. But why not use `exportDir`/`exportAbsdir` as the mount point and `exportDirFsPath = mkFsPath []`? That would be fine I think

lsm-tree/src/Database/LSMTree.hs
```diff
@@ -2953,37 +2932,50 @@ Throws the following exceptions:
...
+  let mountPoint = MountPoint mountPointPath
+  let importDirFsPath = mkFsPath [importRelDir]
 
-The 'FsPath' path to the destination directory is a relative path that is interpreted
-relative to a /root/.
-What the root is depends on which function was used to create the session.
-See 'withOpenSession', 'withOpenSessionIO', and 'withOpenMountedSessionIO' for more
-information about the root.
+  -- Import the snapshot.
+  importSnapshot session snapshotName (Just (ioHasFS @IO mountPoint), importDirFsPath)
```
jorisdral: Same here, I think we can use `importDir`/`importAbsDir` as the mount point and `importDirFsPath = mkFsPath []`

lsm-tree/src/Database/LSMTree.hs
```diff
@@ -2953,37 +2932,50 @@ Throws the following exceptions:
...
+  -- | The source directory and optionally a 'HasFS' instance,
+  --   if the source directory is on a different filesystem.
+  (Maybe (HasFS m h), FsPath) ->
   m ()
 importSnapshot (Session session) =
     Internal.importSnapshot session
 
 {- |
-Export a snapshot to an external directory.
+Variant of 'importSnapshot' that is specialised to 'IO' using the real filesystem.
```
jorisdral, line 2954: Same here, maybe we should mention that it copies

lsm-tree/src/Database/LSMTree.hs
```diff
@@ -2953,37 +2932,50 @@ Throws the following exceptions:
...
+-}
+importSnapshotIO ::
+    Session IO ->
+    SnapshotName ->
+    FilePath ->
+    IO ()
+importSnapshotIO session snapshotName importDir = do
+  -- Get the absolute path to the export directory.
+  exportAbsDir <- Dir.makeAbsolute importDir
```
jorisdral:
```suggestion
  importAbsDir <- Dir.makeAbsolute importDir
```

lsm-tree/src/Database/LSMTree.hs
```diff
@@ -2897,24 +2888,12 @@ listSnapshots (Session session) =
 {- |
 Import a snapshot from an external directory.
 
-The source directory should exist.
-Snapshots should only be imported from a directory on the same volume as the session directory.
-
-Importing does not not check whether the external directory is a snapshot, and
-neither does importing verify the snapshot contents if it is a snapshot.
-Open the snapshot to verify that it is a snapshot and that it is not corrupted.
```
jorisdral, line 2905: This comment should be preserved. It is important information for the user

lsm-tree/src/Database/LSMTree.hs
```diff
@@ -2897,24 +2888,12 @@ listSnapshots (Session session) =
 {- |
 Import a snapshot from an external directory.
```
jorisdral, line 2887: I think this comment should specify when it hard links and when it copies. Seeing that it can now copy, we should probably also update the complexity

lsm-tree/src/Database/LSMTree.hs
```diff
@@ -2953,37 +2932,50 @@ Throws the following exceptions:
...
+  -- Import the snapshot.
+  importSnapshot session snapshotName (Just (ioHasFS @IO mountPoint), importDirFsPath)
 
-It is generally not advisable to use 'importSnapshot' and 'exportSnapshot' when the session is
-created using 'withOpenSessionIO'.
-Snapshots should only be exported to somewhere /outside/ the session directory, which is not possible
-when the session is created using 'withOpenSessionIO'.
-Use 'withOpenMountedSessionIO' or 'withOpenSession' instead.
+{- |
+Export a snapshot to an external directory.
```
jorisdral, line 2974: I think this comment should specify when it hard links and when it copies. Seeing that it can now copy, we should probably also update the complexity

lsm-tree/src/Database/LSMTree.hs
```diff
@@ -429,11 +435,9 @@ If the interaces are instantiated with the real file system, then the root is an
...
 If the interfaces are instantiated with a simulation, then the root is some abstract location.
 
-Any 'FsPath' paths used with the session after the session is created are also interpreted
-with respect to the root of these interfaces.
-'importSnapshot' and 'exportSnapshot' are currently the only two functions that take
-'FsPaths' as arguments.
-'FsPath's are subject to a number of constraints, which are mentioned in its Haddock documentation.
+If the 'FsPath' argument to 'importSnapshot' or 'exportSnapshot' is passed without the optional
+'HasFS' instance, these 'FsPath' paths are interpreted with respected to the 'HasFS' instance
+that was passed to /this/ function.
```
jorisdral:
```suggestion
If the 'FsPath' argument to 'importSnapshot' or 'exportSnapshot' is passed without the optional
'HasFS' instance, these 'FsPath' paths are interpreted with respect to the interface roots.
```

lsm-tree/src/Database/LSMTree.hs
```diff
@@ -484,21 +488,8 @@ withOpenSession ::
...
 
-{-# DEPRECATED withOpenSessionIO "withOpenSessionIO is not compatible with importSnapshot and exportSnapshot. Use withOpenMountedSessionIO instead." #-}
 {- |
 Variant of 'withOpenSession' that is specialised to 'IO' using the real filesystem.
-
-Any 'FsPath' paths used with the session after the session is created are interpreted
-with respect to the session directory.
-'importSnapshot' and 'exportSnapshot' are currently the only two functions that take
-'FsPaths' as arguments.
-'FsPath's are subject to a number of constraints, which are mentioned in its Haddock documentation.
```
jorisdral, line 495: I feel like we should at least preserve the comment about what the interface root / mount point of the session is going to be. If I use `importSnapshot`/`exportSnapshot` without the optional `HasFS` argument, then I need to know that the path argument is interpreted relative to the root. Or we should say that users *have to* provide the optional `HasFS` argument (maybe this is also something that we can check)

lsm-tree/CHANGELOG.md
```diff
@@ -1,5 +1,105 @@
...
+  a `FilePath` field.
+
+* Add a new `exportSnapshot` function that exports snapshots to disk using
+  a `FilePath` path and a variant of `SnapshotExportDirExistsError` with a
+  `FilePath` field.
+
+### Minor changes
+
+* Revert the deprecation of `withOpenSessionIO`, since the other changes in this
+  version have made it safe to use with `importSnapshot` and `exportSnapshot`.
```
jorisdral, line 40: I feel like this statement is not yet really true. I can use `importSnapshot`/`exportSnapshot` on a session created with `withOpenSessionIO` and without providing a destionation FS. In that case, we would interpret paths with respect to the session mount point

jorisdral (commented):
There are two general things to think about here:

* If I create a session with `withOpenSessionIO`, then I can still use `importSnapshot`/`exportSnapshot` without providing a destination FS, which gets us back in unsafe territory where paths are interpreted with respect to the session directory. Maybe we can track a simple `Bool` in the `Session` type that tracks whether the session was created in `withOpenSessionIO`, and if it was, throw an error when a desination FS is not provided in `importSnapshot`/`exportSnapshot`

* The hardlink with copy fallback is nice but has its downsides. it's not visible to the user if the fallback was used, or whether the files were hard linked. This is important to know, because changes to hard linked files are visible to all links, but not if the files were copied. Moreover, I think users will want to only hard link or only copy. So maybe we should have separate functions:
  - `importSnapshotCopy` (and `importSnapshotCopyIO`)
  - `exportSnapshotCopy` (and `exportSnapshotCopyIO`)
  - `importSnapshotHardlink`
  - `exportSnapshotHardlink`
  
  I intentionally left out `IO` versions of the hard linking import/export functions, because we want to only allow copies for sessions that are created with `withOpenSessionIO`. And maybe we should then add a new public error that captures the `EXDEV` error, so that users can write something like this *on top of* the `lsm-tree` library: try hard link, and if `EXDEV`, copy

lsm-tree/src/Database/LSMTree/Simple.hs
wenkokke: We should absolutely put them in the simple module. There's no reason not to, and if we don't, the Simple module will start to fall into disrepair.

lsm-tree/CHANGELOG.md
wenkokke: Added a warning to withOpenSessionIO – I think it _is_ true that it's now safe and useful, but it does warrant a warning to let you know that if you're passing `(Nothing, ["hello","world"])` you're writing stuff relative to the session directory.

lsm-tree/src/Database/LSMTree.hs
wenkokke: As the comment says, _if_ that happens, then the path must already exist anyway.

wenkokke:
> * If I create a session with `withOpenSessionIO`, then I can still use `importSnapshot`/`exportSnapshot` without providing a destination FS, which gets us back in unsafe territory where paths are interpreted with respect to the session directory. Maybe we can track a simple `Bool` in the `Session` type that tracks whether the session was created in `withOpenSessionIO`, and if it was, throw an error when a desination FS is not provided in `importSnapshot`/`exportSnapshot`

There's much more unsafe stuff in the library that is only guarded by a library. I think a well written warning is sufficient, which I've added.

wenkokke:
> * The hardlink with copy fallback is nice but has its downsides. it's not visible to the user if the fallback was used, or whether the files were hard linked. This is important to know, because changes to hard linked files are visible to all links, but not if the files were copied.

`ls -l` tells you if a file has hard links, so the user _can_ tell.

jorisdral (approved):

PR #873, opened 2026-08-10
Support FreeBSD

jorisdral:
Related to #873

This PR only adds support for FreeBSD. If there is need in the future for supporting other OS distributions, then we can add them at that point

### DIFF
--- a/.github/workflows/ci-freebsd.yml
+++ b/.github/workflows/ci-freebsd.yml
@@ -1,0 +1,91 @@
+name: CI (FreeBSD)
+
+on:
+  push:
+    branches:
+    - "main"
+  pull_request:
+  merge_group:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+# Set the default shell on all platforms.
+defaults:
+  run:
+    shell: sh
+
+jobs:
+  ################################################################################
+  # Build (FreeBSD)
+  ################################################################################
+  build:
+    name: ${{ 'Build (FreeBSD)' }}
+    runs-on: 'ubuntu-latest'
+    timeout-minutes: 15
+
+    strategy:
+      fail-fast: false
+
+    steps:
+    - name: 📥 Checkout repository
+      uses: actions/checkout@v7
+
+    # We are currently only testing blockio because that is where the
+    # FreeBSD-specific code lives, not in lsm-tree.
+
+    # TODO <https://github.com/IntersectMBO/lsm-tree/issues/874>: test blockio
+    # more extensively
+
+    # Note: The VM image and system dependencies are cached,
+    # but all Haskell dependencies are rebuilt on every run.
+    # It might be nice to cache those too at some point, but it is
+    # tricky because leaving and re-entering the VM also incurs a performance cost.
+    # Also, it is important not to cache too many VM images, since they are close to 1GB in size.
+
+    # Note: We are using the GHC and Cabal versions that come recommended with ghcup.
+
+    - name: 🧪 Build and test in FreeBSD VM
+      id: test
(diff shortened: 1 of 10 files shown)


.github/workflows/ci-freebsd.yml
```diff
@@ -0,0 +1,95 @@
...
+
+    - name: 🧪 Build and test in FreeBSD VM
+      id: test
+      uses: vmactions/freebsd-vm@v1
+      env:
+        TASTY_TIMEOUT: "5m"
+      with:
+        release: "14"
+        envs: 'TASTY_TIMEOUT'
+        prepare: |
```
wenkokke, line 61: Probably a good idea to do the `pkg install` and GHC/Cabal setup in the prepare step?

lsm-tree/lsm-tree.cabal
```diff
@@ -40,6 +40,8 @@ description:
   * This package only supports 64-bit, little-endian systems.
   * On Windows, the package has only been tested with NTFS filesystems.
   * On Linux, executables using this package, including test and benchmark suites, must be compiled with the [@-threaded@](https://downloads.haskell.org/ghc/latest/docs/users_guide/phases.html#ghc-flag-threaded) RTS option enabled.
+  * This package is supported and tested on Linux, MacOS and Windows.
+  * This package is supported but limitedly tested on FreeBSD
```
wenkokke: "with limited testing on FreeBSD"?

lsm-tree/README.md
```diff
@@ -59,6 +59,10 @@ the key is simply deleted.
   [`-threaded`](https://downloads.haskell.org/ghc/latest/docs/users_guide/phases.html#ghc-flag-threaded "https://downloads.haskell.org/ghc/latest/docs/users_guide/phases.html#ghc-flag-threaded")
   RTS option enabled.
 
+- This package is supported and tested on Linux, MacOS and Windows.
+
+- This package is supported but limitedly tested on FreeBSD
```
wenkokke: "with limited testing on FreeBSD"?

lsm-tree/lsm-tree.cabal
wenkokke: Or just be explicit and say CI is using a third-party FreeBSD VM

wenkokke (approved):

.github/workflows/ci-freebsd.yml
jorisdral: indeed, I should try that

lsm-tree/lsm-tree.cabal
jorisdral: It's not only the VM, but also that it's just one combination of GHC/Cabal that we are testing with, and we're not running `lsm-tree` on FreeBSD CI (mainly because it is too expensive)

.github/workflows/ci-freebsd.yml
```diff
@@ -0,0 +1,91 @@
...
+        release: "14"
+        envs: 'TASTY_TIMEOUT'
+        cache-after-prepare: true
+        prepare: |
+          pkg install -y curl gcc gmp gmake ncurses perl5 libffi libiconv
+
+          curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
+          . /root/.ghcup/env
+          ghc --version
+          cabal --version
```
arrowd, line 67: Why not use GHC and Cabal from packages?

jorisdral: `ghcup` is familiar to me, but the FreeBSD ecosystem is not. Do you think it's better to install from FreeBSD's package repository?

arrowd: You can install ghcup from packages as well, but yes - I think simply doing `pkg install ghc` is the most straightforward way.

I'm maintaining all the Haskell stuff for FreeBSD, so if something breaks just ping me.

Issue #864, opened 2026-07-20
Improve error message for unsupported (future) version numbers


jorisdral:
https://github.com/IntersectMBO/lsm-tree/blob/934503f5f8f12014b8ac23a4212bfc280a1365e0/lsm-tree/src-core/Database/LSMTree/Internal/Snapshot/Codec.hs#L222-L230

It could probably say unsupported rather than unknown and have a helpful message about it being a future version

mwihoti:
I addressed in #875. The `fail` in `Decode SnapshotVersion` now uses the message from the new `SnapshotVersionUnknownError`: it says the version is unsupported, lists the versions this library supports, and notes the snapshot was most likely written by a newer version.

PR #875, opened 2026-08-15
Add SnapshotVersionMismatchError for snapshot version mismatches

### LINKED ISSUE #862: Custom error datatype for snapshot version mismatches
When opening a snapshot with an incompatible version number, we currently throw a `SnapshotCorruptedError`. It would be useful if a version mismatch was its own error datatype, so that it is easier for users to catch such exceptions and act on them.

https://github.com/IntersectMBO/lsm-tree/blob/934503f5f8f12014b8ac23a4212bfc280a1365e0/lsm-tree/src-core/Database/LSMTree/Internal/Snapshot/Codec.hs#L193-L208

mwihoti:
# Description
Resolves #862 and #864.

Right now a version mismatch comes out as a `SnapshotCorruptedError` wrapping `ErrFileFormatInvalid`, with the actual version buried in the message string. This adds a dedicated `SnapshotVersionMismatchError` so it can be caught separately from real corruption.

The awkward part is that the check happens inside a CBOR `Decoder`, where `fail :: String -> Decoder s a` is the only exit, so there's no way to throw a typed exception from in there. `readFileSnapshotMetaData` now decodes in two phases instead: read the version header, check compatibility, then decode the payload with the versioned decoder. `deserialiseFromBytes` returns the leftover bytes, so this doesn't cost a second parse over the file.
Note `isCompatible` can't return `Left` today — every constructor is in `allCompatibleSnapshotVersions` — so the only reachable mismatch is a snapshot from a *newer* library, which has no `SnapshotVersion` constructor here. Hence the found version is a raw `Word`. Checksum and parse failures still throw `SnapshotCorruptedError`.

This is breaking: code catching `SnapshotCorruptedError` for version mismatches will quietly stop catching them, and still compile.

Tests: 
compatibility-check and header-decoding properties in `lsm-tree/test/Test/Database/LSMTree/Internal/Snapshot/Codec.hs`; `prop_versionMismatch` in `lsm-tree/test/Test/Database/LSMTree/Internal/Snapshot` writes metadata with a future version and checks what comes back.

[link]( https://github.com/IntersectMBO/lsm-tree/issues/862)

# Checklist

- [x] cabal build all and cabal test all pass (GHC 9.6.7)
- [x] stylish-haskell 0.15.1.0 applied
image

### DIFF
--- a/lsm-tree/src-core/Database/LSMTree/Internal/Snapshot/Codec.hs
+++ b/lsm-tree/src-core/Database/LSMTree/Internal/Snapshot/Codec.hs
@@ -8,6 +8,11 @@ module Database.LSMTree.Internal.Snapshot.Codec (
   , prettySnapshotVersion
   , currentSnapshotVersion
   , allCompatibleSnapshotVersions
+    -- * Errors
+  , SnapshotVersionMismatchError (..)
+  , snapshotVersionFromWord
+  , checkVersionCompatibility
+  , decodeVersionHeader
     -- * Writing and reading files
   , writeFileSnapshotMetaData
   , readFileSnapshotMetaData
@@ -94,6 +99,44 @@ isCompatible otherVersion
   = Right ()
   | otherwise = Left "forward compatibility not supported"
 
+-- | Convert a raw version number to a known snapshot version.
+--
+-- Returns 'Nothing' for version numbers that are not known to this version of
+-- the library.
+snapshotVersionFromWord :: Word -> Maybe SnapshotVersion
+snapshotVersionFromWord 0 = Just V0
+snapshotVersionFromWord 1 = Just V1
+snapshotVersionFromWord 2 = Just V2
+snapshotVersionFromWord _ = Nothing
+
+-- | Check whether a raw version number found in snapshot metadata is
+-- compatible with the current snapshot version of the library.
+--
+-- Returns an explanation of the incompatibility if the version number is
+-- unknown to this version of the library, or known but incompatible.
+checkVersionCompatibility :: Word -> Either String SnapshotVersion
+checkVersionCompatibility versionNum =
+  case snapshotVersionFromWord versionNum of
+    Nothing ->
+      Left ("unknown snapshot format version number: " <> show versionNum)
+    Just version ->
+      version <$ isCompatible version
+
+-- | The snapshot metadata declares a snapshot format version that is
+-- incompatible with the current snapshot version of the library.
+data SnapshotVersionMismatchError
+    = ErrSnapshotVersionMismatch
+        -- | The version number found in the snapshot metadata. This is a raw
+        -- number because it may not correspond to any 'SnapshotVersion' that
+        -- is known to this version of the library.
+        !Word
+        -- | The current snapshot version of the library.
+        !SnapshotVersion
+        -- | An explanation of the incompatibility.
+        !String
+    deriving stock (Show, Eq)
+    deriving anyclass (Exception)
+
 {-------------------------------------------------------------------------------
   Writing and reading files
 -------------------------------------------------------------------------------}
@@ -138,6 +181,10 @@ encodeSnapshotMetaData = toLazyByteString . encode . Versioned
   #-}
 -- | Read from 'SnapshotMetaDataFile' and attempt to decode it to
 -- 'SnapshotMetaData'.
+--
+-- Throws a 'SnapshotVersionMismatchError' if the snapshot metadata declares a
+-- snapshot format version that is incompatible with the current snapshot
+-- version of the library.
 readFileSnapshotMetaData ::
      (MonadThrow m)
   => HasFS m h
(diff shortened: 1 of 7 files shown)


jorisdral:
It seems something has gone wrong with your force push

mwihoti:
@jorisdral Sorry about the noises,something went wrong with my fork's branches. I've restored to the reviewed commit (`040f1222`); the diff is unchanged from before. thanks

mwihoti:
image

lsm-tree/src-core/Database/LSMTree/Internal/Snapshot/Codec.hs
```diff
@@ -94,6 +99,44 @@ isCompatible otherVersion
...
+        -- | The version number found in the snapshot metadata. This is a raw
+        -- number because it may not correspond to any 'SnapshotVersion' that
+        -- is known to this version of the library.
+        !Word
+        -- | The current snapshot version of the library.
+        !SnapshotVersion
+        -- | An explanation of the incompatibility.
+        !String
+    deriving stock (Show, Eq)
+    deriving anyclass (Exception)
```
jorisdral, line 138: There are two possible errors now:

1. The `Word` version number that is read from the snapshot metadata is unknown. This may occur if the snapshot was created with a newer library, but we are reading it with an older library.
2. The `Word` version number that is read  from the snapshot metadata is known as a `SnapshotVersion`, but that `SnapshotVersion` is not compatible with `currentSnapshotVersion`. This can happen in the future as we stop supporting backwards compatibility for some snapshot versions 

I propose we add two separate datatypes for them:

1. `SnapshotVersionUnkownError`
2. `SnapshotVersionIncompatibleError`

lsm-tree/src-core/Database/LSMTree/Internal/Snapshot/Codec.hs
```diff
@@ -156,10 +203,36 @@ readFileSnapshotMetaData hfs contentPath checksumPath = do
...
+-- undecoded input, from which the metadata itself can be decoded using
+-- 'decodeSnapshotMetaData'.
+decodeVersionHeader :: ByteString -> Either String (ByteString, Word)
+decodeVersionHeader lbs =
+    first displayException (deserialiseFromBytes headerDecoder lbs)
+  where
+    headerDecoder = do
+      _ <- decodeListLenOf 2 -- 'Versioned' wrapper
+      _ <- decodeListLenOf 1 -- 'SnapshotVersion' wrapper
+      decodeWord
```
jorisdral, line 228: This now duplicates decoder logic. It's likely that `headerDecoder` gets out of sync with respect to the `Decode (Versioned a)`  and `Decode SnapshotVersion` instances. I'd be happier with these changes if there is no duplication. Ideas?

lsm-tree/src-core/Database/LSMTree/Internal/Snapshot/Codec.hs
```diff
@@ -223,11 +296,9 @@ instance Decode SnapshotVersion where
...
       _ <- decodeListLenOf 1
       ver <- decodeWord
-      case ver of
-        0 -> pure V0
-        1 -> pure V1
-        2 -> pure V2
-        _ -> fail ("Unknown snapshot format version number: " <>  show ver)
+      case snapshotVersionFromWord  ver of
+        Just version -> pure version
+        Nothing -> fail ("Unknown snapshot format version number: " <>  show ver)
```
jorisdral, line 301: This now also duplicates a bit of logic with `checkVersionCompatibility`. We should try to remove that duplication

lsm-tree/src-core/Database/LSMTree/Internal/Snapshot/Codec.hs
```diff
@@ -156,10 +203,36 @@ readFileSnapshotMetaData hfs contentPath checksumPath = do
 
     expectChecksum hfs contentPath expectedChecksum actualChecksum
 
-    expectValidFile hfs contentPath FormatSnapshotMetaData (decodeSnapshotMetaData lbs)
-
-decodeSnapshotMetaData :: ByteString -> Either String SnapshotMetaData
-decodeSnapshotMetaData lbs = bimap displayException (getVersioned . snd) (deserialiseFromBytes decode lbs)
+    -- Decode and check the version header before decoding the payload, so
+    -- that a version mismatch is reported as a 'SnapshotVersionMismatchError'
+    -- instead of a generic parse failure.
```
jorisdral, line 208: For the record, I don't see `SnapshotVersionMismatchError` being thrown anywhere. Are you sure you implemented throwing the error?

jorisdral (changes requested):
I think this is a step in the right direction

BTW: could you sync your fork and rebase your branch on that? It looks like your fork is too old to run the CI jobs correctly

lsm-tree/src-core/Database/LSMTree/Internal/Snapshot/Codec.hs
mwihoti: sure I'll add that

jorisdral:
Superseded by #887

PR #816, opened 2026-02-23
Issue 795

kanyuku:
To fix the missing record fields in the documentation, I updated the library's public modules to use wildcard exports ((..)) for TableConfig and other key types, ensuring that Haddock automatically includes all fields and methods in the generated pages.

PR #885, opened 2026-09-16
 Add TableConfig fields to the haddocks

### LINKED ISSUE #795: [BUG] The haddock documentation on `lsm-tree`'s Hackage page for `TableConfig` is missing record fields
`TableConfig` exposes its record fields from the `Database.LSMTree` module like so:

```haskell
  TableConfig (
    confMergePolicy,
    confSizeRatio,
    confWriteBufferAlloc,
    confBloomFilterAlloc,
    confFencePointerIndex,
    confDiskCachePolicy,
    confMergeSchedule,
    confMergeBatchSize
  ),
```

However, these record fields are not rendered in the Haddocks. See:
https://hackage-content.haskell.org/package/lsm-tree-1.0.0.0/docs/Database-LSMTree.html#t:TableConfig

mwihoti:
# Description
Resolves #795.

The `TableConfig` fields were missing from the docs because Haddock only shows record fields when the constructor is exported, and we keep the constructor hidden on purpose so that adding a field later isn't a breaking change (which is why the `(..)` approach in #816 didn't fly).

So instead of listing the fields inside the `TableConfig` export, this exports the selectors as plain functions from `Database.LSMTree` and`Database.LSMTree.Simple`. Nothing new is exported and the constructor is still hidden, but each field now shows up with its type and docs. I also
moved the per-field docs from the definition list on the type onto the fields  themselves, so they're only written once.

** Before **(Hackage, 1.1.1.0): https://hackage-content.haskell.org/package/lsm-tree-1.1.1.0/docs/Database-LSMTree.html#t:TableConfig — the type doc only, no fields.
** currrent **
image

I also checked `blockio` and `bloomfilter` for the same export pattern, as asked in the issue — `TableConfig` was the only one.

[link](https://github.com/IntersectMBO/lsm-tree/issues/795) to the issue.

# Checklist

- [x] Cabal build and test passed successfully
image


lsm-tree/src-core/Database/LSMTree/Internal/Config.hs
```diff
@@ -52,60 +52,44 @@ A collection of configuration parameters for tables, which can be used to tune t
...
   , confFencePointerIndex :: !FencePointerIndexType
+    -- | The /disk cache policy/ supports caching lookup operations using the OS page cache.
+    -- Caching may improve the performance of lookups and updates if database access follows certain patterns.
   , confDiskCachePolicy   :: !DiskCachePolicy
+    -- | The merge batch size balances the maximum latency of individual update
+    -- operations, versus the latency of a sequence of update operations. Bigger
+    -- batches improves overall performance but some updates will take a lot
+    -- longer than others. The default is to use a large batch size.
   , confMergeBatchSize    :: !MergeBatchSize
   }
```
jorisdral, line 94: I think I'd prefer the previous version, where the documentation was on `TableConfig`. So let's rewind that

lsm-tree/CHANGELOG.md
```diff
@@ -38,6 +38,10 @@
 
 * Revert the deprecation of `withOpenSessionIO`, since the other changes in this
   version have made it safe to use with `importSnapshot` and `exportSnapshot`.
+* The record fields of `TableConfig` are now rendered in the haddocks, with
+  their types and documentation. See [issue
+  #795](https://github.com/IntersectMBO/lsm-tree/issues/795) and [PR
+  #885](https://github.com/IntersectMBO/lsm-tree/pull/885).
```
jorisdral, line 44: I think this is a breaking change, since we're exporting the fields no longer as part of `TableConfig`, but as separate entities in the export list

lsm-tree/src/Database/LSMTree.hs
```diff
@@ -116,17 +116,16 @@ module Database.LSMTree (
...
+  TableConfig,
   defaultTableConfig,
+  confMergePolicy,
+  confMergeSchedule,
+  confSizeRatio,
+  confWriteBufferAlloc,
+  confBloomFilterAlloc,
+  confFencePointerIndex,
+  confDiskCachePolicy,
+  confMergeBatchSize,
```
jorisdral, line 128: Did you deliberately sort them a different way? Let's keep the order as it was 

```suggestion
  confMergePolicy,
  confSizeRatio,
  confWriteBufferAlloc,
  confBloomFilterAlloc,
  confFencePointerIndex,
  confDiskCachePolicy,
  confMergeSchedule,
  confMergeBatchSize,
  defaultTableConfig,
```

lsm-tree/src/Database/LSMTree/Simple.hs
```diff
@@ -96,15 +96,14 @@ module Database.LSMTree.Simple (
...
-        confMergeSchedule
-    ),
+    TableConfig,
+    confMergePolicy,
+    confMergeSchedule,
+    confSizeRatio,
+    confWriteBufferAlloc,
+    confBloomFilterAlloc,
+    confFencePointerIndex,
+    confDiskCachePolicy,
```
jorisdral, line 106: Did you deliberately sort them a different way? Let's keep the order as it was 

```suggestion
    confMergePolicy,
    confSizeRatio,
    confWriteBufferAlloc,
    confBloomFilterAlloc,
    confFencePointerIndex,
    confDiskCachePolicy,
    confMergeSchedule
```

jorisdral (changes requested):
Changes look good to me! I have a few small requests and then we can merge this

lsm-tree/src-core/Database/LSMTree/Internal/Config.hs
mwihoti: I'll revert to the previous version

PR #886, opened 2026-09-17
Added fallback implementation to blockio used when  platform isn't Linux, MacOS and FreeBSD

### LINKED ISSUE #872: Support platforms other than Linux, OSX, and Windows
... 
I'm not positive that the OSX implementation is correct for FreeBSD. Some of the code (like the `Fcntl.hsc` module) assumes that the platform is OSX. I think the way to go would probably be to add a dedicated fallback implementation for Unix that is used when the platform is not specifically Linux, OSX, or Windows. 
...

_Originally posted by @jorisdral in https://github.com/IntersectMBO/lsm-tree/issues/847#issuecomment-4416350922_

mwihoti:
# Description

Resolves #872.

I added a generic portable fallback:  serial `HasBlockIO` in `src-unix`, which gets picked when the platform isn't Linux, MacOS, Windows or FreeBSD. I also introduced a `posixblockio` cabal flag that forces this fallback on any platform, allowing CI to build and test it on Linux without needing extra runners.

Instead of reusing the macOS backend (as #847 did, which assumes `F_NOCACHE`), I chose the dedicated fallback route for Unix. The generic POSIX layer cannot assume that `O_DIRECT`, `F_NOCACHE` or `posix_fallocate` exist, and `unix`'s `fileAllocate` and `fileSetCaching` throw where they're missing. Therefore I guarded those features using the `HAVE_*` macros from `HsUnix.h` (the same include used by the FreeBSD backend) so they gracefully degrade to no-ops, similar to how the macOS backend already handles `hAllocate`. Both modules are `.hsc` rather than `.hs` because GHC's `cpp` can't process `HsUnix.h` (it pulls in glibc headers) while hsc2hs can  the same reason the FreeBSD backend's `Fcntl` is `.hsc`.

Finally, I added a `+posixblockio` entry to the CI matrix alongside `+serialblockio`, and updated the platform notes in `IO.hs`, the README and the changelog.

Tested locally: the default Linux build is unchanged, `cabal build blockio --flag=+posixblockio` compiles the fallback, and `cabal test blockio:test --flag=+posixblockio` passes all 11 tests against it. One limitation: on Linux all the `HAVE_*` macros are true, so CI only compiles the "supported" branches; the `#else` no-ops are only compiled on a platform that actually lacks the syscalls. Running `--flag=+posixblockio` on FreeBSD would be a useful second data point.

# Checklist

image
image
image


blockio/blockio.cabal
```diff
@@ -92,7 +101,16 @@ library
...
 
-  if os(linux)
+  if (flag(posixblockio) || !(((os(linux) || os(osx)) || os(windows)) || os(freebsd)))
+    -- The portable fallback: forced by the flag, or the platform has no
+    -- dedicated implementation.
+    hs-source-dirs: src-unix
+    build-depends:  unix ^>=2.8.4
+    other-modules:
+      System.FS.BlockIO.Internal
+      System.FS.BlockIO.Internal.Fcntl
```
jorisdral, line 110: Are you sure you committed all the necessary files? I don't see `src-unix` anywhere

blockio/blockio.cabal
```diff
@@ -92,7 +101,16 @@ library
     , primitive   ^>=0.9
     , vector      ^>=0.13
 
-  if os(linux)
+  if (flag(posixblockio) || !(((os(linux) || os(osx)) || os(windows)) || os(freebsd)))
+    -- The portable fallback: forced by the flag, or the platform has no
+    -- dedicated implementation.
+    hs-source-dirs: src-unix
+    build-depends:  unix ^>=2.8.4
```
jorisdral: Note that the `HAVE_O_DIRECT` and `HAVE_F_NOCACHE` are only available since `unix-2.8.7`. We are supporting down to `2.8.4` because it is a requirement by the Cardano node, though I don't think the Cardano node requires us to support `unix-2.8.4` even for this fallback backend. So let's change it to this:

```suggestion
    build-depends:  unix ^>=2.8.7
```

blockio/CHANGELOG.md
```diff
@@ -1,5 +1,32 @@
...
+
+### New features
+
+* Add a portable serial `HasBlockIO` implementation for POSIX systems that do
+  not have a dedicated one, so `blockio` builds on more than Linux, MacOS,
+  Windows, and FreeBSD. Operations the platform does not support
+  (`posix_fallocate`, `O_DIRECT`/`F_NOCACHE`) degrade to no-ops. See [issue
+  #872][issue-872] and [PR #NNN][pr-NNN].
+* Add the `posixblockio` cabal flag, which forces the portable implementation
+  on any platform, for testing it. See [PR #NNN][pr-NNN].
```
jorisdral, line 17: Fill in the `[PR #NNN][pr-NNN]` links

jorisdral (changes requested):
I think the plan you outline in the PR description is sensible! But something may have gone wrong with your commits, because I'm missing code changes.

P.S.: no need to post screenshots of your console with test output. CI will verify that the tests succeed.

blockio/blockio.cabal
mwihoti: I've pushed the code

blockio/CHANGELOG.md
mwihoti: updated

blockio/blockio.cabal
mwihoti: changed to 2.8.7

blockio/CHANGELOG.md
jorisdral: I don't see the changes yet

jorisdral, line 17: I'm unsure about using the name "POSIX" to describe the fallback implementation. Posix is a standard, not an operating system. I think we should call it something like "unix-like", or maybe even "fallback". Do you have thoughts on this?

blockio/README.md
```diff
@@ -22,8 +22,10 @@ concurrently. On Windows, MacOS, and FreeBSD systems the *real* implementation
 currently simply performs each I/O operation sequentially, which should achieve
 about the same performance as using non-batched I/O, but the library could be
 extended with asynchronous I/O implementations for Windows, MacOS, and FreeBSD
-as well. The simulated implementation also performs each I/O operation
-sequentially.
+as well. On other POSIX systems a portable serial implementation is used that
+relies only on the parts of the `unix` package available on the platform; it
+can be selected explicitly on any platform with the `posixblockio` cabal flag.
+The simulated implementation also performs each I/O operation sequentially.
```
jorisdral, line 28:
```suggestion
as well. If the operating system is none of the above, then the system is assumed to be Unix-like, in which case the library uses a fallback implementation that performs I/O operations sequentially.
The simulated implementation also performs each I/O operation sequentially.
```

blockio/CHANGELOG.md
```diff
@@ -1,5 +1,32 @@
...
+
+None
+
+### New features
+
+* Add a portable serial `HasBlockIO` implementation for POSIX systems that do
+  not have a dedicated one, so `blockio` builds on more than Linux, MacOS,
+  Windows, and FreeBSD. Operations the platform does not support
+  (`posix_fallocate`, `O_DIRECT`/`F_NOCACHE`) degrade to no-ops. See [issue
+  #872][issue-872] and [PR #NNN][pr-NNN].
```
jorisdral, line 15: This should also emphasize that the OS is now assumed to be Unix-like in case the OS is not one of the explicitly supported OS's (Linux, MacOS, Windows, FreeBSD). This is different from the previous behaviour, where the library would simply be unbuildable on unsupported OS's.

blockio/blockio.cabal
```diff
@@ -71,6 +71,15 @@ flag serialblockio
...
   manual:      True
 
+flag posixblockio
+  description:
+    Use the portable POSIX HasBlockIO implementation regardless of the
+    operating system. Intended for testing that implementation on platforms
+    that have a dedicated one. Not supported on Windows.
+
+  default:     False
+  manual:      True
```
jorisdral, line 81:
```suggestion
flag unixlikeblockio
  description:
    Use the Unix-like HasBlockIO implementation regardless of the
    operating system. Intended for testing that implementation on platforms
    that have a dedicated one. No-op on Windows.

  default:     False
  manual:      True
```

blockio/blockio.cabal
```diff
@@ -92,7 +101,15 @@ library
     , primitive   ^>=0.9
     , vector      ^>=0.13
 
-  if os(linux)
+  if (flag(posixblockio) || !(((os(linux) || os(osx)) || os(windows)) || os(freebsd)))
```
jorisdral, line 104: The Unix-like implementation won't build on Windows, so let's make the flag a no-op on Windows

```suggestion
  if ((flag(posixblockio) && !os(windows)) || !(os(linux) || os(osx) || os(freebsd)))
```

blockio/src/System/FS/BlockIO/IO.hs
```diff
@@ -52,69 +53,83 @@ import           System.FS.IO (HandleIO, ioHasFS)
...
+      otherwise no-op
 
   ['tryLockFile']: This uses different locking methods depending on the OS.
 
     * Linux: Open file descriptor (OFD)
     * MacOS: @flock@
     * Windows: @LockFileEx@
     * FreeBSD: Open file descriptor (OFD)
+    * Other POSIX systems: whatever GHC's @hTryLock@ uses on the platform, i.e.
+      open file descriptor (OFD) locks where available, otherwise @flock@
```
jorisdral, line 108:
```suggestion
    * Other POSIX systems: whatever GHC's @hTryLock@ uses on the platform
```

blockio/src-unix/System/FS/BlockIO/Internal/Fcntl.hsc
jorisdral, line 1: You probably copied this module from another `System.FS.BlockIO.Internal.Fcntl`, but it's no longer necessary. This type of module is only necessary to support `unix` versions `<unix-2.8.7`, but we support `>=unix-2.8.7`.

blockio/src/System/FS/BlockIO/IO.hs
```diff
@@ -39,7 +39,8 @@ import           System.FS.IO (HandleIO, ioHasFS)
 
   Note: if the @serialblockio@ Cabal flag is enabled, then the Linux
   implementation uses a mocked context and serial I\/O for 'close' and
-  'submitIO', just like the MacOS, Windows, and FreeBSD implementations do.
+  'submitIO', just like the MacOS, Windows, FreeBSD, and portable POSIX
+  implementations do.
```
jorisdral, line 43:
```suggestion
  'submitIO', just like the MacOS, Windows, FreeBSD, and Unix-like
  implementations do.
```

blockio/src/System/FS/BlockIO/IO.hs
```diff
@@ -52,69 +53,83 @@ import           System.FS.IO (HandleIO, ioHasFS)
     * MacOS: a mocked context using an @MVar@
     * Windows: a mocked context using an @MVar@
     * FreeBSD: a mocked context using an @MVar@
+    * Other POSIX systems: a mocked context using an @MVar@
```
jorisdral, line 56: We should update this everywhere:
```suggestion
    * Unix-like systems: a mocked context using an @MVar@
```

blockio/src-unix/System/FS/BlockIO/Internal.hsc
```diff
@@ -0,0 +1,76 @@
...
+#if HAVE_POSIX_FALLOCATE
+import qualified System.Posix.Fcntl as Fcntl (fileAllocate)
+#endif
+import qualified System.Posix.Files as Unix
+import qualified System.Posix.Unistd as Unix
+
+-- | A portable, serial implementation of 'HasBlockIO' for POSIX systems that
+-- do not have a dedicated implementation (Linux, MacOS, and FreeBSD do). It
+-- uses only the parts of the @unix@ package that are available on the
+-- platform, and degrades to no-ops where an operation is not supported.
```
jorisdral, line 27:
```suggestion
-- | A portable, serial implementation of 'HasBlockIO' for Unix-like systems that
-- do not have a dedicated implementation (Linux, MacOS, and FreeBSD do). 
```

blockio/src-unix/System/FS/BlockIO/Internal.hsc
```diff
@@ -0,0 +1,76 @@
...
+#if HAVE_POSIX_FALLOCATE
+hAllocate h off len = FS.withOpenHandle "hAllocate" (handleRaw h) $ \fd ->
+    Fcntl.fileAllocate fd off len
+#else
+hAllocate _h _off _len = pure ()
+#endif
+
+hSynchronise :: Handle HandleIO -> IO ()
+hSynchronise h = FS.withOpenHandle "hSynchronise" (handleRaw h) $ \fd ->
+    Unix.fileSynchronise fd
```
jorisdral, line 72: Looks to me like `fileSynchronise` can also throw an error if the system does not `HAVE_FSYNC`. We should also make it a no-op in that case, like we do with `hAllocate`. We should also update the docs accordingly.

jorisdral (changes requested):

